PR #8, opened 2023-02-27
✨ textToImage inference

### LINKED ISSUE #7: Add `text-to-image` API inference
hi @TimMikeladze it would be great to add `text-to-image` on the inference package.
I'm not sure if it's well documented on the [docs](https://huggingface.co/docs/api-inference/detailed_parameters), but it's widely available on all `text-to-image` models widgets on the hub, 
for example https://huggingface.co/prompthero/openjourney?text=mdjrny-v4+style+a+flying+cat

Here a simple example request:
```js
fetch(
  "https://api-inference.huggingface.co/models/stabilityai/stable-diffusion-2",
  {
    method: "POST",
    headers: {
      "X-Wait-For-Model": "true" 
    },
    body: JSON.stringify({
      inputs:
        "award winning high resolution photo of a giant tortoise/((ladybird)) hybrid, [trending on artstation]",
      negative_prompt: "blurry",
    }),
  }
)
  .then((res) => res.blob())
  .then((blob) => {
    const tab = window.open((target = "_blank"));
    tab.location.href = window.URL.createObjectURL(blob);
  });
````

### PR DESCRIPTION
**Pull request by TimMikeladze**
Closes #7 

Adds `text-to-image` support to the inference api. `textToImage` returns a `Buffer`. The consumer can then call `.toString('base64')` to get the generated image for usage in the browser, etc.

```ts
 await hf.textToImage({
			inputs:          "award winning high resolution photo of a giant tortoise/((ladybird)) hybrid, [trending on artstation]",
			negative_prompt: "blurry",
			model:           "stabilityai/stable-diffusion-2",
		});
```

### DIFF
--- a/packages/inference/LICENSE
+++ b/packages/inference/LICENSE
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2022 TimMikeladze
+Copyright (c) 2022 Tim Mikeladze
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
--- a/packages/inference/README.md
+++ b/packages/inference/README.md
@@ -775,5 +775,5 @@ type TextToImageArgs = Args & {
    */
   negative_prompt?: string
 };
-type TextToImageReturn = Buffer
+type TextToImageReturn = ArrayBuffer
 ```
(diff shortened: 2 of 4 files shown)


coyotte508:
Browsers don't support `Buffer`.

I think returning `ArrayBuffer` is fine - we target recent versions of Node (where `fetch` is available) which support it too. 

Regarding `base64` conversion, if we ever want to expose an helper to the user there is [this](https://github.com/huggingface/huggingface.js/blob/9e74503954d6abecaf0032e862f1097a34d374fd/packages/hub/src/utils.ts#LL39) in `@huggingface/hub`, can be called `base64FromBytes(new Uint8Array(arrayBuffer))`

PR #9, opened 2023-02-27
🐛 wait_for_model fix and reanable imageSegmentation test

### LINKED ISSUE #4: `wait_for_model` not being respected / disabled imageSegmentation test
Currently the `wait_for_model` option is not being respected when using the inference package. Due to this the imageSegmentation test is failing  and is currently disabled.

TimMikeladze:
Closes #4 

I noticed that in the code-sample provided in #7 the "wait for model" functionality is activated via the header `X-Wait-For-Model` instead of being sent in the body of the `POST` request. There is a discrepancy with the documentation in this regard https://huggingface.co/docs/api-inference/detailed_parameters

This made me wonder: should other options such as `use_cache` also be sent as headers?

### DIFF
--- a/packages/inference/src/HfInference.ts
+++ b/packages/inference/src/HfInference.ts
@@ -653,10 +653,19 @@ export class HfInference {
 	): Promise<any> {
 		const mergedOptions = { ...this.defaultOptions, ...options };
 		const { model, ...otherArgs } = args;
+
+    const headers = {
+      Authorization: `Bearer ${this.apiKey}`,
+    }
+
+    if (mergedOptions.wait_for_model) {
+      headers["X-Wait-For-Model"] = "true";
+    }
+
 		const response = await fetch(`https://api-inference.huggingface.co/models/${model}`, {
-			headers: { Authorization: `Bearer ${this.apiKey}` },
-			method:  "POST",
-			body:    options?.binary
+			headers,
+			method: "POST",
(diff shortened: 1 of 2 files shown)


coyotte508:
cc @Narsil , is it ok to pass the `wait_for_model` in the payload, or do we need to pass it as `X-Wait-For-Model` ?

Narsil:
Both are fine.

The header is the only viable option for `image` or `audio` payloads (which are pure bytes, hence, there's no other way to pass parameters around).

This isn't documented as it was meant for internal use. If there are issues, we could just document it instead.

packages/inference/src/HfInference.ts
```diff
@@ -653,10 +653,19 @@ export class HfInference {
 	): Promise<any> {
 		const mergedOptions = { ...this.defaultOptions, ...options };
 		const { model, ...otherArgs } = args;
+
+    const headers = {
+      Authorization: `Bearer ${this.apiKey}`,
+    }
+
+    if (mergedOptions.wait_for_model) {
```
coyotte508, line 661:
```suggestion
    if (options?.binary && mergedOptions.wait_for_model) {
```

From Narsil's comment

coyotte508 (commented):
By the way - you should configure your editor to use tabs instead of spaces (or install the recommended extension)

Or we should add back husky ^^

PR #10, opened 2023-02-27
Compute sha in a streaming fashion

coyotte508:
- Compute sha in a streaming fashion
- Remove support for non-Blobs in `commit`
- Add a dependency `hash-wasm` for files > 10 MB on browser

Will let this PR sit a day or so before merging, and then bump a minor version


packages/hub/src/utils/sha256-node.ts
```diff
@@ -0,0 +1,27 @@
+import { Readable } from "stream";
+import { createHash } from "crypto";
```
TimMikeladze, line 2: Will importing `crypto` here work in browser and edge runtimes?

coyotte508: Confirmed it: published under `@huggingface/hub@alpha` and added this to a svelte-kit app and it works:

```ts
	onMount(() => {
		commit({
			title: 'test',
			repo: {
				name: 'coyotte508/gee',
				type: 'model'
			},
			credentials: { accessToken: 'hf_...' },
			operations: [
				{
					operation: 'addOrUpdate',
					path: 'file.lfs.txt',
					content: new Blob(['Hello World boo\n'.repeat(10_000)])
				}
			]
		});
	});
```

`sha256-node` is only imported if not on browser:

```
	if (!cryptoModule) {
		cryptoModule = await import("./sha256-node");
	}
	return cryptoModule.sha256Node(buffer);
```

No idea about CJS browser build though (but if it's failing now, it already was).

---

Although - we should really find a way to test the package in the browser through e2e tests

PR #27, opened 2023-03-02
Fixed #26 by adding 'Content-Type' in the request headers

### LINKED ISSUE #26: Expected request with `Content-Type: application/json`
Update from [this issue](TimMikeladze/huggingface#18):

Hello! I was informed that `huggingface` has been folded into this repo. Therefore, I tried to reinvestigate the issue using this package (`@huggingface/inference`).

However, the response was still `Unexpected token E in JSON at position 0` (which probably means `Expected request with `Content-Type: application/json` again.)

Upon further investigation, this only occurs if I use `google/flan-t5-xxl` as a model, while `gpt2` works perfectly fine.

Will adding `'Content-Type': 'application/json'` to the request headers solve this problem while not breaking other use cases?

Quatton:
Hello!
I have made changes according to #26.

So far, I
- added `"Content-Type": "application/json"` to the request headers
- updated the test for `google/flan-t5-xxl`
- for `zeroShotClassification`, updated `scores` to make it pass the test. (It might regularly change? I am not sure about that)
- made sure every test passed

What do you think about this PR?

### DIFF
--- a/packages/inference/src/HfInference.ts
+++ b/packages/inference/src/HfInference.ts
@@ -685,7 +685,7 @@ export class HfInference {
 			headers["Authorization"] = `Bearer ${this.apiKey}`;
 		}
     
-    if (!options?.binary) {
+		if (!options?.binary) {
 			headers["Content-Type"] = "application/json";
 		}
 
--- a/packages/inference/src/HfInference.ts
+++ b/packages/inference/src/HfInference.ts
@@ -503,7 +503,7 @@ export class HfInference {
 	private readonly apiKey:         string;
 	private readonly defaultOptions: Options;
 
-	constructor(apiKey: string, defaultOptions: Options = {}) {
+	constructor(apiKey = "", defaultOptions: Options = {}) {
 		this.apiKey = apiKey;
 		this.defaultOptions = defaultOptions;
 	}
@@ -680,10 +680,14 @@ export class HfInference {
 		const mergedOptions = { ...this.defaultOptions, ...options };
 		const { model, ...otherArgs } = args;
 
-		const headers = {
-			Authorization:  `Bearer ${this.apiKey}`,
-			"Content-Type": "application/json",
-		};
+		const headers = {};
+		if (this.apiKey) {
+			headers["Authorization"] = `Bearer ${this.apiKey}`;
+		}
+    
+    if (!options?.binary) {
+			headers["Content-Type"] = "application/json";
+		}
 
 		if (options?.binary && mergedOptions.wait_for_model) {
 			headers["X-Wait-For-Model"] = "true";
--- a/packages/inference/src/HfInference.ts
+++ b/packages/inference/src/HfInference.ts
@@ -680,13 +680,14 @@ export class HfInference {
 		const mergedOptions = { ...this.defaultOptions, ...options };
 		const { model, ...otherArgs } = args;
 
-    const headers = {
-      Authorization: `Bearer ${this.apiKey}`,
-    }
+		const headers = {
+			Authorization:  `Bearer ${this.apiKey}`,
+			"Content-Type": "application/json",
+		};
 
-    if (options?.binary && mergedOptions.wait_for_model) {
-      headers["X-Wait-For-Model"] = "true";
-    }
+		if (options?.binary && mergedOptions.wait_for_model) {
+			headers["X-Wait-For-Model"] = "true";
+		}
 
 		const response = await fetch(`https://api-inference.huggingface.co/models/${model}`, {
 			headers,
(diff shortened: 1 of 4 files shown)


packages/inference/test/HfInference.test.ts
```diff
@@ -5,6 +5,9 @@ import { jest } from "@jest/globals"; // 3 minute
 import { HfInference } from "../src";
 import { readFileSync } from "fs";
 
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+require("dotenv").config();
```
TimMikeladze, line 9: Not necessary. Please remove dotenv and rely on passing the token to tests as described in the docs.

packages/inference/test/HfInference.test.ts
```diff
@@ -159,7 +172,7 @@ describe("HfInference", () => {
 					sequence:
 						"Hi, I recently bought a device from your company but it is not working as advertised and I would like to get reimbursed!",
 					labels: ["refund", "faq", "legal"],
-					scores: [0.877787709236145, 0.10522633045911789, 0.01698593981564045],
+					scores: [0.8777875304222107, 0.10522652417421341, 0.01698593609035015],
```
TimMikeladze, line 175: Oh nice fix. It was an oversight on my behalf but these tests needs to simply check for the existence of an array with numbers. Checking for exact results will end up with test failures.

TimMikeladze (changes requested):
Thank for jumping on this!

Quatton:
Hello! I have updated `scores` to `expect.arrayContaining([expect.any(Number)])` instead, and every test passed.

By the way, I have one question. If there are more model-specific issues, is the plan just keeping fixing it case-by-case like this?

packages/inference/src/HfInference.ts
```diff
@@ -680,13 +680,14 @@ export class HfInference {
 		const mergedOptions = { ...this.defaultOptions, ...options };
 		const { model, ...otherArgs } = args;
 
-    const headers = {
-      Authorization: `Bearer ${this.apiKey}`,
-    }
+		const headers = {
+			Authorization:  `Bearer ${this.apiKey}`,
+			"Content-Type": "application/json",
```
coyotte508, line 685: You can avoid setting this only if `options?.binary` - this should also answer your question :)

packages/inference/test/HfInference.test.ts
radames: just for context @Quatton I've fixed this issue on #31 using `expect.closeTo` but I guess `expect.any(Number)` would also also work

Quatton:
@radames 
Hello! I agree that `expect.closeTo` is a more elegant solution because we can both
1. Confirm that this array has a length of 3. (Mine was just any array that contains number)
2. Confirm that this is not just any number.
I have resolved my merge conflict on that by accepting your changes instead.

@coyotte508 
A recent commit has made to avoid setting `Authorization` if no `apiKey` is set; therefore, we can test the rate limit error functionality (?)

I have resolve my merge conflict on that by following the pattern and added this line
```ts
if (!options?.binary) {
	headers["Content-Type"] = "application/json";
}
```
Should we check for `(!options?.blob)` as well?

coyotte508:
> @coyotte508 A recent commit has made to avoid setting `Authorization` if no `apiKey` is set; therefore, we can test the rate limit error functionality (?)

Here we test the client library mainly. We're not testing HF's backend :sweat_smile: 


> I have resolve my merge conflict on that by following the pattern and added this line
> 
> ```ts
> if (!options?.binary) {
> 	headers["Content-Type"] = "application/json";
> }
> ```
> 
> Should we check for `(!options?.blob)` as well?

 `options?.binary` is for the input, `options?.blob` for the output. Maybe `options?.blob` should impact the `Accept` header but it's not related to this PR

packages/inference/src/HfInference.ts
```diff
@@ -684,6 +684,10 @@ export class HfInference {
 		if (this.apiKey) {
 			headers["Authorization"] = `Bearer ${this.apiKey}`;
 		}
+    
+    if (!options?.binary) {
```
coyotte508, line 688: Last nit: the indent should be tabs here

Quatton: Oops! I added the edit right on the GitHub merge conflict editor, so Prettier didn't fix it. I have pushed a formatting fix in the latest commit. Thank you!

Quatton:
Also, the test on the workflow has failed because the access token hasn't been set. Are we supposed to add someone's token onto the env?

coyotte508:
> Also, the test on the workflow has failed because the access token hasn't been set. Are we supposed to add someone's token onto the env?

Will look into it, but I ran the tests locally and it's fine

coyotte508:
Thanks a lot @Quatton!

PR #33, opened 2023-03-03
Check inference CI on PR from fork

coyotte508:


coyotte508:
Seems to work, maybe testing from an unprivileged user would be better, we'll see next time

PR #50, opened 2023-03-05
🔧 Single eslint/prettier config

### LINKED ISSUE #1: Tooling in its own package?
Should we put eslint + prettier config in a package for example?

Also, husky / lint-staged are not configured yet, I didn't research how to properly do it for a mono-repo

coyotte508:
Fix #1 - without using a package, as suggested by @pngwn 

cc @pngwn @o-az - It's suprisingly easy to share eslint/prettier configs without creating an additional package. We can even remove the eslint/prettier dependencies from the sub-packages.

@TimMikeladze - The `"@typescript-eslint/consistent-type-definitions": ["error", "interface"],` now applies to `@huggingface/inference` (for doc purposes, so the doc generated by typedoc puts the interfaces in their own files) and it caused a significant diff. Let me know if you want to revert that.


pngwn:
Looks good. Regarding dev deps I'd  consider adding them all to root and installing them as deps for simplicity. It's generally good practice in monorepos, since you don't need to think about how to install dev deps in that case. 

Might not be possible right now but just thought I'd mention it.

coyotte508:
> Looks good. Regarding dev deps I'd consider adding them all to root and installing them as deps for simplicity. It's generally good practice in monorepos, since you don't need to think about how to install dev deps in that case.
> 
> Might not be possible right now but just thought I'd mention it.

Do you mean renaming `devDependencies` in https://github.com/huggingface/huggingface.js/pull/50/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R3 to `dependencies` or something else? I'm ok with it.

pngwn:
Yep just that.

Issue #51, opened 2023-03-05
Tasks as a pointer in README


merveenoyan:
[Tasks](https://huggingface.co/tasks) is a project that is built for software developers willing to find appropriate ML model for their use case and build apps, and from what I see in this repository I felt like the audience is somewhat same, does it make sense to put it in README or somewhere in documentation? Wanted to open discussion here before opening a PR.

coyotte508:
Maybe in the @huggingface/inference [README](https://github.com/huggingface/huggingface.js/blob/main/packages/inference/README.md)? cc @TimMikeladze

TimMikeladze:
Agreed.

Not everyone is familiar with the ML space and this repo may be their first exposure to it. 

Providing users with additional helpful links and resources in the  respective readme is a great idea.

radames:
very good idea @merveenoyan, I'll also link add a link to the live doc notebook on Observable.
@merveenoyan do you want to open a PR otherwise I can do it thanks

PR #52, opened 2023-03-06
Added short description for tasks

merveenoyan:
I wrote very short description to add tasks in where you have pointed @coyotte508

### DIFF
--- a/docs/inference/README.md
+++ b/docs/inference/README.md
@@ -180,3 +180,7 @@ await hf.textToImage({
 ```console
 HF_ACCESS_TOKEN="your access token" npm run test
 ```
+
+# Finding model appropriate for your task
+
+We have an informative documentation project called [Tasks](https://huggingface.co/tasks) to explain how each task works and how to infer with the models, for software developers.


docs/inference/README.md
```diff
@@ -180,3 +180,7 @@ await hf.textToImage({
 ```console
 HF_ACCESS_TOKEN="your access token" npm run test
 ```
+
+# Finding model appropriate for your task
```
coyotte508, line 184:
```suggestion
## Finding model appropriate for your task
```

docs/inference/README.md
```diff
@@ -180,3 +180,7 @@ await hf.textToImage({
...
+
+We have an informative documentation project called [Tasks](https://huggingface.co/tasks) to explain how each task works and how to infer with the models, for software developers.
```
coyotte508, line 186:
```suggestion
If you want to try out different models for each of the APIs, you can look at our [Tasks](https://huggingface.co/tasks) pages to easily find compatible models.
```

I think what the users of the JS libs will want is just to find a model for the task, they don't care about the inner workings, just that it works. As for the inputs, it's probably handled by the Typescript definitions. 

If there are model-specific inputs, then we should maybe write an additional documentation page (if the lib supports arbitrary inputs).

cc @radames or @TimMikeladze  to make sure I don't say anything ridiculous

radames: Agree with @coyotte508 , that if the model-specific inputs should be explained in the docs here. But I like @merveenoyan docs explaining each task especially for a users landing in ML world via our JS library.

coyotte508: Alright, let me rephrase: 

>  to explain how each task works

The thing is `tasks` is mentioned nowhere before in the README, and here we talk about tasks out of the blue. For example we have a section called `## Supported APIs` above, maybe we should rename it `## Supported Tasks` ?

> and how to infer with the models, for software developers.

The tasks page doesn't do that. The README of the `@huggingface/inference` package explains how to infer, by calling `hf.questionAnswering(...)`.

The advantage of the tasks page is to explain what each tasks does, and to list models for each tasks.

julien-c: maybe orthogonal to this, but i do think we should have a model id as default (for users who just want to do a task but know no models)

Sorry if it's off topic here

julien-c: hmm no i think it's better if people can dive a bit and introspect stuff and realize it's actually a defined model not a black box API

Unlike OpenAI we want to empower people to actually do some ML not just consume it

merveenoyan:
fixes #51

docs/inference/README.md
coyotte508:
```suggestion
We have an informative documentation project called [Tasks](https://huggingface.co/tasks) to list available models for each task and explain how each task works in detail.

It also contain demos, example outputs and other resources should you want to dig more into the ML-side of things.
```

Something like this?

I just don't agree with "to infer with the model for a given task for a software developer" because it's this README's job / `@huggingface/infernece`'s job to do that, at least for JS / TS developers. 

The TS type inference will autocomplete their code to provide the correct parameters / call the correct functions.

coyotte508 (approved):

coyotte508:
Thanks @merveenoyan !

PR #53, opened 2023-03-06
link to live notebook note about api key

radames:
* Link to live notebook
* Remove random string recommendation (it won't work since it will be a random authorization header)
* Note about the private key on the web

### DIFF
--- a/packages/inference/README.md
+++ b/packages/inference/README.md
@@ -2,7 +2,8 @@
 
 A Typescript powered wrapper for the Hugging Face Inference API. Learn more about the Inference API at [Hugging Face](https://huggingface.co/docs/api-inference/index).
 
-Check out the [full documentation](https://huggingface.co/docs/huggingface.js/inference/README).
+Check out the [full documentation](https://huggingface.co/docs/huggingface.js/inference/README) or try out a live [interactive notebook](https://observablehq.com/@huggingface/hello-huggingface-js-inference).
+
 
 
 ## Install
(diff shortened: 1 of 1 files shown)


packages/inference/README.md
```diff
@@ -2,7 +2,8 @@
 
 A Typescript powered wrapper for the Hugging Face Inference API. Learn more about the Inference API at [Hugging Face](https://huggingface.co/docs/api-inference/index).
 
-Check out the [full documentation](https://huggingface.co/docs/huggingface.js/inference/README).
+Check out the [full documentation](https://huggingface.co/docs/huggingface.js/inference/README) or try out a live [interactive notebook](https://observablehq.com/@huggingface/hello-huggingface-js-inference).
```
coyotte508, line 5: JFI, The observableHQ link will be removed when deploying to HF docs unfortunately, due to this in `fix-md-links.ts`:

```ts
			// When on HF, remove links to HF
			.replaceAll(/.*\[full documentation\].*\n\n/g, "")
```

Because it removes the full line including "[full documentation]". (yes it's a hack)

coyotte508 (approved):

packages/inference/README.md
radames: I see, whatever works, is it worth changing it to a new paragraph then?

coyotte508: We can just merge it and open an issue if you want

Issue #37, opened 2023-03-03
Generated docs


coyotte508:
Currently we need to add a commit to the repo each time we want to update the docs.  

IMO, either:

- Add a github action we can run inside PRs before merging them to update the generated doc
- Stop storing the generated docs in the repo, work with github actions / `[doc-builder](https://github.com/huggingface/doc-builder)` to be able to generate on the fly

coyotte508:
Actually, what may be better would be to generate docs at the same time as releasing a package - we'd update the docs for that package in particular

PR #55, opened 2023-03-07
Add `changesets` support

o-az:
How it works, 
1. make changes to both or either packages (`hub`/`interference`),
2. run `pnpm changeset` in root. Once CLI prompts are answered,
3. commit and push code to your branch,
4. create a pull request to `main`,
5. once approved and merged, `release.yml` will run and create a bot pull request with changelog,
6. `npm` & `gpr` publishing are triggered.

Here's a workflow test on my forked repo: https://github.com/o-az/huggingface.js/actions/runs/4353313461/jobs/7607187269

I believe this pr still needs to address the docs generation part. I decided to make the pr first and get some feedback before I went further.

Which one would work best?

1. Run `pnpm --filter 'doc-internal' prepublish-inference && pnpm --filter 'doc-internal' prepublish-hub` right after `changesets` step,
2. right before `changesets` step,
3. run after packages are published,
4. or, add `"prebuild": "pnpm --filter 'doc-internal' prepublish-hub"` scripts to `hub` and inference` (this way it will alwyas run before since `enable-pre-post-scripts` is enabled at `.npmrc`).

I'm leaning towards `4`. Let me know what your thoughts are


coyotte508:
Thanks a lot for this PR!

As @pngwn mentioned in another issue, the publish should happen only after merging the PR generated by `changesets`. Is this the case here ? (not familiar with `changesets`)

package.json
```diff
@@ -1,15 +1,24 @@
...
+    "eslint-config-prettier": "^8.6.0",
+    "eslint-plugin-prettier": "^4.2.1",
+    "prettier": "^2.8.4",
+    "typescript": "^4.9.5"
+  },
+  "scripts": {
+    "changeset:release": "pnpm run build && changeset publish",
+    "changeset:version": "changeset version && pnpm install",
+    "build": "pnpm --filter './packages/*' build",
+    "test": "pnpm --filter './packages/*' test",
```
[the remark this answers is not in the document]
o-az: I think it's useful to keep these in the root package.json so that running a build for all packages is a simple `pnpm build` command. Same for test.

For example the build script is being used in one of the changesets script right now. 

It's best if there's a build, test, lint and typecheck commands in root and in each package individually but more importantly in root. This way `pnpm build|test|lint|typecheck` can each be invoked from root and if someone's working inside a specific package they can invoke from there too. In CI you'd want to have a workflow that runs on pull requests like you already do but it should run the root lint, typecheck, test, and build commands (in sequence) as opposed to running them for specific packages individually in separate workflows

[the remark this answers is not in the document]
coyotte508: Sure - but those can be a separate PR (which also updates the root's README), it's not related to this one

.github/workflows/release.yml
```diff
@@ -0,0 +1,84 @@
...
+        uses: changesets/action@v1.4.1
+        with:
+          title: 'chore: version packages'
+          commit: 'chore: version packages'
+          version: pnpm changeset:version
+          publish: pnpm changeset:release
+        env:
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
```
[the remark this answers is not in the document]
o-az: I need to remove the node one. Npm is needed for publishing and github for pr. 

https://github.com/changesets/action#with-publishing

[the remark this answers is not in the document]
coyotte508: yes, I said only `NPM_TOKEN` is needed :)

o-az:
@coyotte508 yes. What about the docs and their command, should they happen after the packages are published? In this sequence:

`bot PR merged -> npm/gpr publish -> run docs commands`?

o-az:
I tested the latest changes outside of this repo and works but i also want to test in this repo before merge once review is done. So I'm going to create another branch and temporarily update the branch field under `on: push:` to that branch for a test. @coyotte508 thoughts?

coyotte508:
Thanks a lot @o-az for the changes! It's a nice solution for publishing.

As for the docs, they change files inside the repo that need to be commited (at least until #37 ), so they should probably be integrated in the release commands (?), to be integrated when the PR merges.

coyotte508:
And the doc can probably be generated all at once since we publish versions for all packages with changes at the same time (no need to generate specifically for hub & specifically for inference)

o-az:
> And the doc can probably be generated all at once since we're going to publish versions for all packages with changes at the same time (no need to generate specifically for hub & specifically for inference)
> 
> It can be generated with `pnpm --filter @huggingface/doc-internal start`

oh sweet! I just pushed a change that adds `prebuild` scripts to hub and inference and enabled `pre-post-scripts`. But I like the command you shared better so I will update right now.

o-az:
@coyotte508 alrighty I removed the `prebuild` scripts and added `docs:update` script in root that will get called as part of `changesets` release:

https://github.com/o-az/huggingface.js/blob/1aee63d6fc24a70daea1fd52c8beb169b3534beb/package.json#L7-L8

Now I want to update the following two lines temporarily from `main` to `test-changesets` and create a `test-changesets` branch to test this branch:

https://github.com/o-az/huggingface.js/blob/1aee63d6fc24a70daea1fd52c8beb169b3534beb/.github/workflows/release.yml#L15

https://github.com/o-az/huggingface.js/blob/1aee63d6fc24a70daea1fd52c8beb169b3534beb/.changeset/config.json#L13

I've already tested in a separate repo but would like to do a test here too.

Can you create a `test-changesets` branch since you have permission? Then I will make a PR to it.

packages/hub/package.json
```diff
@@ -21,9 +21,8 @@
 	},
 	"source": "index.ts",
 	"scripts": {
-    "prebuild": "pnpm --filter 'docs-internal' prepublish-hub",
-		"format": "prettier --write . && eslint --quiet --fix --ext .cjs,.ts .",
-		"prepublishOnly": "npm run build",
+    "format": "prettier --write . && eslint --quiet --fix --ext .cjs,.ts .",
+		"prepublishOnly": "npm build && pnpm update-docs",
```
coyotte508, line 25:
```suggestion
		"prepublishOnly": "npm build",
```

I don't think this should be here - I may be mistaken but as I understand, changesets:

1. opens a PR
2. keeps it up to date
3. after merging the PR, it publishes

We want to generate the docs in step 2., before the PR is merged (so the doc changes can be included)

coyotte508:
> Can you create a `test-changesets` branch since you have permission? Then I will make a PR to it.

Oops just saw that. Done!

packages/hub/package.json
o-az: Yes! This for sure shouldn't be there. Thanks for catching it

o-az:
> > Can you create a `test-changesets` branch since you have permission? Then I will make a PR to it.
> 
> 
> 
> Oops just saw that. Done!

Thank you. I'll comeback to finish it off later today/tomorrow. Cheers

o-az:
@coyotte508 I can't figure out why this is failing https://github.com/huggingface/huggingface.js/actions/runs/4359734486/jobs/7621845796#step:7:62
Any idea?

coyotte508:
> @coyotte508 I can't figure out why this is failing [huggingface/huggingface.js/actions/runs/4359734486/jobs/7621845796#step:7:62](https://github.com/huggingface/huggingface.js/actions/runs/4359734486/jobs/7621845796#step:7:62) Any idea?

run `pnpm -r format` at the source, basically the json files use spaces instead of tabs

image

.editorconfig
```diff
@@ -1,7 +1,13 @@
...
 indent_size = 2
 trim_trailing_whitespace = true
-insert_final_newline = true
\ No newline at end of file
+insert_final_newline = true
+
+# json uses 2 space indents
+[*.{json,jsonc,json5,yml,yaml}]
+indent_style = space
+indent_size = 2
```
coyotte508, line 13: Hm no, `pnpm format` will scream at you ^^

We use tabs for everything, only the `pnpm-lock.yml` files are ignored

o-az: Oh i thought this meant json should use 2 spaces instead of tabs https://github.com/huggingface/huggingface.js/pull/55#issuecomment-1459093271

make sense now. Though I didn't get any complains from pnpm format. Thanks for looking out

o-az:
can you approve this pr @coyotte508? Base branch is a test branch

.eslintignore
```diff
@@ -0,0 +1,23 @@
...
+.pnpm-workspace.yaml
+*.yaml
+*.yml
+dist
+.dist
+**/dist/**
+dist
+logs
+*.toml
+*.md
```
coyotte508, line 23: I don't think there's a need for this? We don't have toml files, nor server/build or /public/build, package-lock.json etc. And the linter only runs on .ts and .cjs files.

Was there a problem somewhere?

package.json
```diff
@@ -1,18 +1,28 @@
...
+    "eslint-plugin-prettier": "^4.2.1",
+    "prettier": "^2.8.4",
+    "typescript": "^4.9.5",
+    "@types/node": "^18.14.6",
+    "vite": "^4.1.4",
+    "vitest": "^0.29.2"
+  },
+  "license": "MIT",
+  "packageManager": "^pnpm@7.0.0"
+}
```
coyotte508, line 28: I'll add linting for this file (tabs were converted to spaces)

packages/doc-internal/package.json
```diff
@@ -4,10 +4,10 @@
 	"description": "Package to generate doc for other @huggingface packages",
 	"private": true,
 	"scripts": {
-		"start": "npm run doc-hub && npm run doc-inference && cp ../../README.md ../../docs/index.md && npm run update-toc && npm run fix-md-links",
+		"start": "pnpm doc-hub && pnpm doc-inference && cp ../../README.md ../../docs/index.md && pnpm update-toc && pnpm fix-md-links",
 		"format": "prettier --write . && eslint --quiet --fix --ext .cjs,.ts .",
-		"prepublish-hub": "npm run doc-hub && cp ../../README.md ../../docs/index.md && npm run update-toc && npm run fix-md-links",
-		"prepublish-inference": "npm run doc-inference && cp ../../README.md ../../docs/index.md && npm run update-toc && npm run fix-md-links",
+		"prepublish-hub": "pnpm doc-hub && cp ../../README.md ../../docs/index.md && pnpm update-toc && pnpm fix-md-links",
+		"prepublish-inference": "pnpm doc-inference && cp ../../README.md ../../docs/index.md && pnpm update-toc && pnpm fix-md-links",
```
coyotte508, line 10: I don't think we need to change `npm run` to `pnpm`. If an external contributor mistakenly uses `npm install` instead of `pnpm install`, they can at least run the commands.

.vscode/settings.json
```diff
@@ -1,7 +1,9 @@
...
-	"editor.formatOnSaveMode": "file" // required to format on save for vs-code-prettier-eslint
-}
+  "editor.defaultFormatter": "rvest.vs-code-prettier-eslint",
+  "editor.formatOnPaste": false, // required for vs-code-prettier-eslint
+  "editor.formatOnSave": true, // required for vs-code-prettier-eslint
+  "editor.formatOnType": false, // required for vs-code-prettier-eslint
+  "editor.formatOnSaveMode": "file", // required to format on save for vs-code-prettier-eslint
+  "typescript.tsdk": "node_modules/typescript/lib",
+  "npm.packageManager": "pnpm"
+}
```
coyotte508, line 9: I'll add linting for this file (tabs were converted to spaces)

packages/hub/package.json
```diff
@@ -22,7 +22,7 @@
 	"source": "index.ts",
 	"scripts": {
 		"format": "prettier --write . && eslint --quiet --fix --ext .cjs,.ts .",
-		"prepublishOnly": "npm run build",
+		"prepublishOnly": "pnpm build",
```
coyotte508, line 25: Same comment about npm run vs pnpm

packages/hub/package.json
```diff
@@ -2,7 +2,7 @@
 	"name": "@huggingface/hub",
 	"version": "0.1.5",
 	"description": "Utilities to interact with the Hugging Face hub",
-	"repository": "https://github.com/huggingface/huggingface.js.git",
+	"repository": "github:huggingface/huggingface.js",
```
coyotte508, line 5: Let's not change this

packages/inference/package.json
```diff
@@ -4,10 +4,7 @@
 	"license": "MIT",
 	"author": "Tim Mikeladze <tim.mikeladze@gmail.com>",
 	"description": "Typescript wrapper for the Hugging Face Inference API",
-	"repository": {
-		"type": "git",
-		"url": "https://github.com/huggingface/huggingface.js.git"
-	},
+	"repository": "github:huggingface/huggingface.js",
```
coyotte508, line 7: Let's not change this either

packages/inference/package.json
```diff
@@ -41,7 +38,7 @@
 	"scripts": {
 		"build": "tsup src/index.ts --format cjs,esm --clean --dts",
 		"format": "prettier --write . && eslint --quiet --fix --ext .cjs,.ts .",
-		"prepublishOnly": "pnpm run build",
+		"prepublishOnly": "pnpm build",
```
coyotte508, line 41: you can change this to `npm run build` if you want

pnpm-workspace.yaml
```diff
@@ -1,4 +1,3 @@
 packages:
-  - "packages/hub"
-  - "packages/inference"
-  - "packages/doc-internal"
+  - "packages/*"
+  - "!packages/doc-internal"
```
coyotte508, line 3: It's still part of the workspace and we still want linter to run on it in the lint CI. 

You should probably exclude it inside `.changeset/config.json`, in the `ignore` setting instead.

.eslintignore
o-az: I removed this file. I think it should be added in a separate pr with dist, yml/yaml and md. Without .eslintignore the linter lints files in dist which are supposed to be ignored.

coyotte508: ok thanks, I'll look into it

packages/doc-internal/package.json
o-az: it's an anti-pattern to allow more than one package manager in the same repo. `pnpm` is a pretty good choice and easy to install. It also takes much less disk space and is much faster than `npm`.
The `pnpm` team has an official tool to specifically handle the scenario you mention: https://github.com/pnpm/only-allow

.vscode/settings.json
o-az: thanks! hopefully this will prevent [this](https://github.com/huggingface/huggingface.js/pull/55#discussion_r1129223932) from happening again

packages/doc-internal/package.json
coyotte508: Ok, sounds good. 

Can you open a separate PR with the command changes + the only-allow package? 

I'd like to keep each PR changes scoped to the PR, for easier review / minimal diffs / git history once the PR is merged.

o-az:
@coyotte508 I converted the PR to draft. Once I incorporate all the requested changes I'll move it to ready. Hope that's ok with everyone.

.github/workflows/release.yml
```diff
@@ -0,0 +1,98 @@
...
+
+      - name: 'Setup Node.js [1]'
+        uses: actions/setup-node@v3
+        with:
+          cache: 'pnpm'
+          node-version: 18
+          registry-url: 'https://registry.npmjs.org'
+
+      - name: 'Install dependencies'
+        run: pnpm install
```
coyotte508, line 53:
```suggestion
        run: pnpm install --filter .
```

Slight optimization, to only install dependencies of the root package

.github/workflows/release.yml
```diff
@@ -0,0 +1,98 @@
...
+        env:
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+        # for GitHub Packages registry
+      - name: 'Setup Node.js [2]'
+        if: steps.changesets.outputs.published == 'true'
+        uses: actions/setup-node@v3
+        with:
+          cache: 'pnpm'
```
coyotte508, line 72:
```suggestion
```

not needed, we do not run pnpm install afterwards

.vscode/settings.json
```diff
@@ -1,7 +1,13 @@
...
-}
+  "editor.defaultFormatter": "rvest.vs-code-prettier-eslint",
+  "editor.formatOnPaste": false, // required for vs-code-prettier-eslint
+  "editor.formatOnSave": true, // required for vs-code-prettier-eslint
+  "editor.formatOnType": false, // required for vs-code-prettier-eslint
+  "editor.formatOnSaveMode": "file", // required to format on save for vs-code-prettier-eslint
+  "typescript.tsdk": "node_modules/typescript/lib",
+  "npm.packageManager": "pnpm",
+  "editor.tabSize": 2,
+  "editor.insertSpaces": true,
```
coyotte508, line 10:
```suggestion
```

Hm no, I don't think so? It's also overridden by `editor.detectIndentation`

.vscode/settings.json
```diff
@@ -1,7 +1,13 @@
...
+  "editor.detectIndentation": true,
+  "files.insertFinalNewline": true,
```
coyotte508, line 12: Ironic given there's no newline at the end of this file :)

README.md
```diff
@@ -68,5 +68,5 @@ There are more features of course, check each library's README!
 pnpm install
 
 pnpm -r format
-pnpm -r test
+pnpm test
```
coyotte508, line 71: Should revert, it's `pnpm -r test`.

package.json
```diff
@@ -1,18 +1,29 @@
...
-	},
-	"scripts": {
-		"format": "prettier --write .eslintrc.cjs && eslint --quiet --fix --ext .cjs,.ts .eslintrc.cjs",
-		"test": "pnpm -r test"
-	}
+  "author": "Hugging Face",
+  "repository": "github:huggingface/huggingface.js",
+  "scripts": {
+    "build": "pnpm --recursive build",
+    "test": "pnpm --recursive test",
```
coyotte508, line 6:
```suggestion
```

Sorry, no, but there's a good reason why not :)

In the CI, we do `pnpm --filter "...[origin/main]" test`. We do this to only run tests on packages that were updated.

By adding this here, since the root's files are always changed, it will always run ALL tests, when we want to only run _some_ tests.

packages/inference/package.json
```diff
@@ -4,10 +4,7 @@
 	"license": "MIT",
 	"author": "Tim Mikeladze <tim.mikeladze@gmail.com>",
 	"description": "Typescript wrapper for the Hugging Face Inference API",
-	"repository": {
-		"type": "git",
-		"url": "https://github.com/huggingface/huggingface.js.git"
-	},
+	"repository": "https://github.com/huggingface/huggingface.js.git",
```
coyotte508, line 7: Let's not change unnecessarily

.vscode/settings.json
```diff
@@ -1,7 +1,13 @@
 {
-	"editor.defaultFormatter": "rvest.vs-code-prettier-eslint",
-	"editor.formatOnPaste": false, // required for vs-code-prettier-eslint
-	"editor.formatOnSave": true, // required for vs-code-prettier-eslint
-	"editor.formatOnType": false, // required for vs-code-prettier-eslint
-	"editor.formatOnSaveMode": "file" // required to format on save for vs-code-prettier-eslint
-}
```
coyotte508, line 7: you removed the tab indents :(

coyotte508 (commented):
A few things still to address, but it's on a good path.

Most of the comments are on unrelated files to the chore of this PR, which is why opening separate PRs for different things is a good idea :)

packages/inference/package.json
o-az: Ops I must have changed this by accident while updating the hub one. First time I changed all of them was when I was testing publishing to npm so I needed to change repo to the fork under my account so I can publish. Since I'm used to the `github:user/repo-name` format, I automatically use it when filling out this field. Anyways I changed it back.

.vscode/settings.json
o-az: Last try, I will explicitly set my global settings to tabs.

Also, although it's been receiving hate lately, I think pre-commit hooks are very good and useful. If a dev wants to disable, they'll be caught in CI. What do you think?

o-az: @coyotte508 I'm using the vscode builtin `JSON Language Features` which automatically formats JSON files. Is that what you use? If not can you share what you're using and if you have any custom settings? All these tabs to space changes from my commits seem to be automatically happening from my editor.

CleanShot 2023-03-09 at 11 04 09

coyotte508: > Also, although it's been receiving hate lately, I think pre-commit hooks are very good and useful. If a dev wants to disable, they'll be caught in CI. What do you think?

#1 :smiling_face_with_tear: 

I have no strong opinion, @pngwn seems firmly against it though. We can maybe discuss a bit later when there's more contributors.

Still, with the latests changes in main, it should be fixed by `pnpm -r format`

package.json
```diff
@@ -1,17 +1,26 @@
...
+  "dependencies": {
+    "@changesets/changelog-github": "^0.4.8",
+    "@changesets/cli": "^2.26.0",
+    "@typescript-eslint/eslint-plugin": "^5.51.0",
+    "@typescript-eslint/parser": "^5.51.0",
+    "eslint": "^8.35.0",
+    "eslint-config-prettier": "^8.6.0",
+    "eslint-plugin-prettier": "^4.2.1",
+    "prettier": "^2.8.4",
+    "typescript": "^4.9.5",
```
coyotte508, line 19: Tabs are converted to spaces too :grimacing:

o-az: I used the GitHub web editor to resolve conflict in this commit. It should respect the existing indentation style but it doesn't 😕

package.json
```diff
@@ -1,17 +1,26 @@
 {
-	"license": "MIT",
+	"author": "Hugging Face",
+	"repository": "github:huggingface/huggingface.js",
+	"scripts": {
+		"docs:update": "pnpm --filter 'doc-internal' start",
+		"changeset:release": "pnpm --recursive build && pnpm docs:update && changeset publish",
+		"changeset:version": "changeset version",
+		"format": "prettier --write .eslintrc.cjs && eslint --quiet --fix --ext .cjs,.ts .eslintrc.cjs"
```
coyotte508, line 8: The format command has changed since, it's `prettier --write . && eslint --quiet --fix --ext .cjs,.ts .`

o-az: I like it, awesome.

coyotte508: It's still wrong after merging

coyotte508:
Ok, let's try this :grin:

PR #56, opened 2023-03-07
convert all tests to vitest

### LINKED ISSUE #19: Decide on single test library to use
Currently both `mocha` and `jest` are in use.  It would be great to consolidate the testing approach used across all packages.

pngwn:
This converts the mocha and jest tests to vite test.

I had to create a new asymmetric  matcher for `vite` because it doesn't have `expect.closeTo` by default. Added types for this too.

2 package locks have been removed and one added because i think `pnpm` is being run incorrectly, it should always be done from the root and there should only every be one lockfile in the repo, so sure how things are being handled.

Will look at CI but it looks like an install issue.

Fixed the issue with CI, we can't filter the install if we need install devdeps from root.

Closes #19.


coyotte508:
> 2 package locks have been removed and one added because i think `pnpm` is being run incorrectly, it should always be done from the root and there should only every be one lockfile in the repo, so sure how things are being handled.

Can this be handled in a separate PR / issue? It was made separate lockfiles on purpose, and I don't think it's needed for this PR

pngwn:
Sure. I'll add vitest as a dep to each package.

packages/hub/src/lib/commit.spec.ts
```diff
@@ -33,24 +32,24 @@ describe("commit", () => {
...
 						operation: "addOrUpdate",
-						content:   new Blob([lfsContent]),
-						path:      "test.lfs.txt",
+						content: new Blob([lfsContent]),
+						path: "test.lfs.txt",
 					},
 					{
 						operation: "delete",
-						path:      "README.md",
+						path: "README.md",
```
coyotte508, line 52: Can you run `pnpm -r format`?

I'll add a lint action

packages/hub/package.json
```diff
@@ -43,10 +43,8 @@
 	"author": "Hugging Face",
 	"license": "MIT",
 	"devDependencies": {
-		"@types/mocha": "^10.0.1",
 		"@types/node": "^18.13.0",
 		"cross-env": "^7.0.3",
```
coyotte508, line 47: Can be removed if not using it in the test command anymore

coyotte508 (commented):
Thanks! I'll add a lint CI for formatting, still, can we scope this PR to only the tests changes?

coyotte508:
> Sure. I'll add vitest as a dep to each package.

Are you sure it's necessary? `prettier` and `eslint` are only added to the root package and run in the subpackages

pngwn:
Side note, i think the `ListFileEntry` type might be wrong (or the test is wrong). I'm getting a type error in the test file (mocha didn't have type support so it didn't get flagged) but the test is passing.

test is asserting:

```js
{
  lastCommit: {
    date: "2018-11-14T23:35:08.000Z",
    id: "504939aa53e8ce310dba3dd2296dbe266c575de4",
    title: "initial commit",
  },
  oid: "dc08351d4dc0732d9c8af04070ced089b201ce2f",
  path: ".gitattributes",
  security: {
    avScan: {
      virusFound: false,
      virusNames: null,
    },
    blobId: "dc08351d4dc0732d9c8af04070ced089b201ce2f",
    name: ".gitattributes",
    pickleImportScan: null,
    repositoryId: "models/bert-base-uncased",
    safe: true,
  },
  size: 345,
  type: "file",
}
```

But the type is:

```ts
export interface ListFileEntry {
  type: "file" | "directory" | "unknown";
  size: number;
  path: string;
  lfs?: {
    oid:         string;
    size:        number;
   /** Size of the raw pointer file, 100~200 bytes */
    pointerSize: number;
  };
  lastCommit: {
    author: {
      date: string;
    };
    id:      string;
    subject: string;
  } | null;
  security?: unknown;
}
```

Note the `lastCommit` member. The type has an author field but the test case doesn't. The test is passing.

pngwn:
> Are you sure it's necessary? prettier and eslint are only added to the root package and run in the subpackages

They aren't run in CI, so I guess people are running `pnpm i` locally (which will install everything) to run those commands.

packages/hub/src/lib/commit.spec.ts
pngwn: Not sure why my editor isn't picking up the config. Does the repo use eslint for formatting?

coyotte508: Uses a custom extension (recommended)

pngwn:
Sould all be cleaned up now.

coyotte508 (approved):
Thanks a lot!

coyotte508:
(Seems a test is failing)

pngwn:
sad

pngwn:
oh i know

PR #59, opened 2023-03-07
👷 One CI file to rule them all

coyotte508:


packages/inference/test/global.d.ts
```diff
@@ -1,11 +1,13 @@
...
-		interface Assertion extends CustomMatchers {}
-		interface AsymmetricMatchersContaining extends CustomMatchers {}
+		interface Assertion {
+			// todo: fix the any
+			closeTo(expected: number, precision: number): any;
+		}
+		interface AsymmetricMatchersContaining {
+			// todo: fix the any
+			closeTo(expected: number, precision: number): any;
+		}
```
coyotte508, line 10: to pass linting cc @pngwn

package.json
```diff
@@ -7,11 +7,12 @@
...
 		"eslint-plugin-prettier": "^4.2.1",
 		"prettier": "^2.8.4",
-		"typescript": "^4.9.5"
+		"typescript": "^4.9.5",
+		"vite": "^4.1.4",
+		"vitest": "^0.29.2"
 	},
 	"scripts": {
 		"format": "prettier --write .eslintrc.cjs && eslint --quiet --fix --ext .cjs,.ts .eslintrc.cjs",
-		"test:run": "vitest run",
```
coyotte508, line 14: This was giving this error message: 
image

packages/inference/test/global.d.ts
pngwn: exhibit a of why i hate linting :P

pngwn: This isn't the same btw but works well enough

PR #60, opened 2023-03-07
`createRepo` return the repo's URL

julien-c:
we could also do something fancier like `huggingface_hub` i.E. separate the namespace, repo_type etc.: https://github.com/huggingface/huggingface_hub/blob/599647c4d728f8954c6e285ba11aa3d316a6af36/src/huggingface_hub/hf_api.py#L254

but probably overkill

### DIFF
--- a/packages/hub/src/lib/create-repo.ts
+++ b/packages/hub/src/lib/create-repo.ts
@@ -15,7 +15,7 @@ export async function createRepo(params: {
 	/** @required for when {@link repo.type} === "space" */
 	sdk?:        SpaceSdk;
 	hubUrl?:     string;
-}): Promise<{ repoUrl: string; }> {
+}): Promise<{ repoUrl: string }> {
 	const [namespace, repoName] = params.repo.name.split("/");
 
 	const res = await fetch(`${params.hubUrl ?? HUB_URL}/api/repos/create`, {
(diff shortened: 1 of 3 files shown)


packages/hub/src/lib/create-repo.ts
```diff
@@ -54,4 +54,6 @@ export async function createRepo(params: {
 	if (!res.ok) {
 		throw await createApiError(res);
 	}
+	const output = await res.json();
+	return output.url;
```
coyotte508, line 58:
```suggestion
	return { repoUrl: output.url };
```

Maybe? To be consistent with `commit` that returns `{ pullRequestUrl: string, ...}`

packages/hub/src/lib/create-repo.ts
```diff
@@ -15,7 +15,7 @@ export async function createRepo(params: {
 	/** @required for when {@link repo.type} === "space" */
 	sdk?:        SpaceSdk;
 	hubUrl?:     string;
-}): Promise<void> {
+}): Promise<string> {
```
julien-c, line 18:
```suggestion
}): Promise<{ repoUrl: string; }> {
```

packages/hub/src/lib/create-repo.ts
julien-c: Yes, why not.

More generally i would try to be consistent with `huggingface_hub` whenever we can (cc also @Wauplin)

PR #75, opened 2023-03-08
🔥 Remove <source> sanitization in docs

coyotte508:
Since doc-builder is fixed: https://github.com/huggingface/doc-builder/pull/357


coyotte508:
With this, the dark/light images should work on HF's docs too @gary149

PR #77, opened 2023-03-08
Add an estimation of available models

gary149:


coyotte508:
Can those +100k ML models be used with the inference API?

gary149:
From my estimations, yes (at least 2/3 of the 150k public models)

Issue #69, opened 2023-03-08
When or if will there be an NPM package?


hpssjellis:
**When or if will there be an NPM package?**

@coyotte508 

I am a high school teacher who is very good at taking npm packages and making simple machine learning, single page, JavaScript demos, see what I did years ago with tensorflowJS [here](https://hpssjellis.github.io/beginner-tensorflowjs-examples-in-javascript/#tfjs-models).   

Presently I am completely lost at how to get these examples running. I can load huggingface.js onto [gitpod](https://gitpod.io/) (a browser based docker) and install everything but still nothing works.  

Your package.json file does not seem to have many useful scripts. 

HuggingFace.js package.json
```
	"scripts": {
		"format": "prettier --write .eslintrc.cjs && eslint --quiet --fix --ext .cjs,.ts .eslintrc.cjs",
		"test": "pnpm -r test"
	}
```

TensorflowJS  package.json  (Probably not all of these are useful, but the release scripts would be nice.)
```
  "scripts": {
    "lint": "tslint -p tsconfig_tslint.json",
    "test": "bazel test //:tests",
    "test-packages-ci": "yarn generate-cloudbuild-for-packages && ./scripts/run-build.sh",
    "nightly-cloudbuild": "NIGHTLY=true yarn generate-cloudbuild-for-packages && gcloud builds submit . --config=cloudbuild_generated.yml --substitutions=_NIGHTLY=true",
    "generate-cloudbuild-for-packages": "ts-node -s ./scripts/generate_cloudbuild_for_packages.ts",
    "test-generate-cloudbuild": "cd scripts && node --require ts-node/register ../node_modules/jasmine/bin/jasmine.js run generate_cloudbuild_test.ts",
    "test-run-flaky": "jasmine run scripts/run_flaky_test.js",
    "release": "ts-node -s ./scripts/release.ts",
    "release-tfjs": "ts-node -s ./scripts/release-tfjs.ts",
    "publish-npm": "ts-node -s ./scripts/publish-npm.ts",
    "release-notes": "ts-node -s ./scripts/release_notes/release_notes.ts",
    "test-release-notes": "ts-node -s ./scripts/release_notes/run_tests.ts",
    "update-tfjs-lockfiles": "ts-node -s ./scripts/update-tfjs-lockfiles",
    "tag-tfjs-release": "ts-node -s ./scripts/tag-tfjs-release",
[...]

PR #81, opened 2023-03-09
more installing info in README.md

radames:
Give a bit of context on how to instal via NPM and via vanilla JS
responding to #69

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,36 @@ We use modern features to avoid polyfills and dependencies, so the libraries wil
 
 The libraries are still very young, please help us by opening issues!
 
+## Installation
+
+### From NPM
+
+To install via NPM, you can download the libraries as needed:
+
+```bash
+npm install @huggingface/hub
+npm install @huggingface/inference
+```
+
+Then import the libraries in your code:
+
+```ts
+import { createRepo, commit, deleteRepo, listFiles } from "@huggingface/hub";
+import { HfInference } from "@huggingface/inference";
+import type { RepoId, Credentials } from "@huggingface/hub";
+```
+
+### From CDN or Static hosting
+You can run our packages with vanilla JS, without any bundler, by using a CDN or static hosting. Using [ES modules](https://hacks.mozilla.org/2018/03/es-modules-a-cartoon-deep-dive/), i.e. `<script type="module">`, you can import the libraries in your code:
+
+```html
+
+<script type="module">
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -22,6 +22,36 @@ We use modern features to avoid polyfills and dependencies, so the libraries wil
...
+```
+
+### From CDN or Static hosting
+You can run our packages with vanilla JS, without any bundler, by using a CDN or static hosting. Using [ES modules](https://hacks.mozilla.org/2018/03/es-modules-a-cartoon-deep-dive/), i.e. `<script type="module">`, you can import the libraries in your code:
+
+```html
+
+<script type="module">
+    import { HfInference } from 'https://cdn.jsdelivr.net/npm/@huggingface/inference@1/+esm';
+    import { createRepo, commit, deleteRepo, listFiles } from "https://cdn.jsdelivr.net/npm/@huggingface/hub@0/+esm";
```
coyotte508, line 51:
```suggestion
    import { HfInference } from 'https://cdn.jsdelivr.net/npm/@huggingface/inference/+esm';
    import { createRepo, commit, deleteRepo, listFiles } from "https://cdn.jsdelivr.net/npm/@huggingface/hub/+esm";
```

Maybe we need to add an import map for `@huggingface/hub` for `hash-wasm` when uploading files over 10MB in the commit call too

radames: thanks, I can add that example as well. Are you warning/throwing error depends on the size for the file? we could also add that info on the error message, wdyt?

coyotte508: We're not throwing any errors but we're using the `hash-wasm` dependency when dealing with big files in the browser

radames: ok, thanks, I did some tests with sourcemap and unfortunately its not working, the commit still works for small files, I guess we need a fix. Shall we merge this and update later?

coyotte508: Does adding something like this at the beginning work ?

```ts
<script type="importmap">
  {
    "imports": {
      "@huggingface/hub": "https://cdn.skypack.dev/@huggingface/hub",
      "hash-wasm": "https://cdn.skypack.dev/hash-wasm"
    }
  }
</script>
```

but even if it doesn't, I think it's ok to merge for now

radames: not really, :( I can create an issue on try slack , the bundler is not linking it correctly 
https://glitch.com/edit/#!/humble-confirmed-shadow

radames: sorry I think I forget to commit your suggestion above 🙈

PR #83, opened 2023-03-09
🦺 Runtime check on credentials

coyotte508:
cc @o-az

### DIFF
--- a/packages/hub/src/utils/checkCredentials.ts
+++ b/packages/hub/src/utils/checkCredentials.ts
@@ -1,0 +1,11 @@
+import type { Credentials } from "../types";
+
+export function checkCredentials(credentials?: Credentials): void {
+	if (!credentials || credentials.accessToken === undefined || credentials.accessToken === null) {
+		return;
+	}
+
+	if (!credentials.accessToken.startsWith("hf_")) {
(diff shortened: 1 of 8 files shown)


packages/hub/src/utils/checkCredentials.ts
```diff
@@ -0,0 +1,11 @@
+import type { Credentials } from "../types";
+
+export function checkCredentials(credentials?: Credentials): void {
+	if (!credentials || credentials.accessToken === undefined || credentials.accessToken === null) {
+		return;
```
o-az, line 5: it's legal to not provide access token, it will just be rate limited - so it's not an error if token is not provided. Is mu understanding right?

o-az (commented):
awesome. Does the inference package make use of access token as well?

packages/hub/src/utils/checkCredentials.ts
coyotte508: For some `@huggingface/hub` functions, like listing files in a repo, the token is optional. It can be useful if the repo is private, otherwise it's not needed.

coyotte508:
> awesome. Does the inference package make use of access token as well?

Yes. We should also add a check there probably (saying to not provide an access token as an alternative)

PR #86, opened 2023-03-10
🐛 Fix browser detection

### LINKED ISSUE #85: Loading hub package on browser with importmap doesn't work
Trying to run `@huggingface/hub` on the browser adding `hash-wasm` with `importmap` still throws error. Additionally, you can see on the network activity that `hash-wasm` hasn't been loaded. 

Potential solutions, explore `peerDependencies` on package.json ? AFAIK, using rollup you can set https://rollupjs.org/tools/#peer-dependencies and externals, so it doesn't bundle it on the final package. Since jsdelivr is automatically bundling it with rollup, see code transformed [here](https://cdn.jsdelivr.net/npm/@huggingface/hub@0.1.5/+esm)  WDYT?

```
  "peerDependencies": {
    "hash-wasm": "^4.9.0",
  }
```
``` js
<script type="importmap">
      {
        "imports": {
          "@huggingface/hub": "https://cdn.jsdelivr.net/npm/@huggingface/hub/+esm",
          "hash-wasm": "https://cdn.jsdelivr.net/npm/hash-wasm/+esm"
        }
      }
    </script>
    <script type="module">
      import {
        createRepo,
        commit,
        deleteRepo,
        listFiles,
      } from "@huggingface/hub";
</script>

```
To reproduce, you need to upload a file sized > 10MB
https://glitch.com/edit/#!/humble-confirmed-shadow?path=index.htm

image

coyotte508:
Fix #85 

Actually `__filename` is not a valid check because of rollup

### DIFF
--- a/packages/hub/src/utils/sha256.ts
+++ b/packages/hub/src/utils/sha256.ts
@@ -13,7 +13,12 @@ export async function sha256(buffer: Blob): Promise<string> {
 		);
 	}
 
-	if (typeof __filename === undefined) {
+	const isBrowser = typeof window !== "undefined" && typeof window.document !== "undefined";
+
+	const isWebWorker =
+		typeof self === "object" && self.constructor && self.constructor.name === "DedicatedWorkerGlobalScope";
+
+	if (isBrowser || isWebWorker) {
 		if (!wasmModule) {
 			wasmModule = await import("hash-wasm");
 		}


coyotte508:
@radames tested (thanks to your example on glitch) by releasing a version with tag "alpha", and it works. No need to even add `hash-wasm` to the import map!

PR #96, opened 2023-03-13
💥 Return Blob instead of ArrayBuffer for images

coyotte508:
cc @hpssjellis, should fix #93 

It's a breaking change but given the relative newness of the lib + arraybuffer endpoint, let's make it a minor version after merging


TimMikeladze (approved):

coyotte508:
Thanks @TimMikeladze !

radames:
thanks @coyotte508 , this will make it much easier to generate the image.

Issue #68, opened 2023-03-08
Add a helper to generate Blob (lazily read) from a file path


coyotte508:
Or wait on https://github.com/nodejs/node/issues/45188

This would help for the `commit` calls, when uploading very large files, to avoid loading the whole file in memory.

Would help with #62 

Technical PR.

---

Given that it's only needed for Deno & Node (and node should have a native solution soon), it should be an internal helper. We want to be able to remove it at some point.

coyotte508:
> Also about the usage `Blob.length` [here](https://github.com/huggingface/huggingface.js/blob/main/packages/hub/src/lib/commit.ts?rgh-link-date=2023-03-17T19%3A54%3A52Z#L218) and [here](https://github.com/huggingface/huggingface.js/blob/main/packages/hub/src/lib/commit.ts?rgh-link-date=2023-03-17T19%3A54%3A52Z#L325), isn't it an error since it does not exists on the Blob API?

Thanks! I added a bug fix: https://github.com/huggingface/huggingface.js/commit/b7934a51c87ab407dca9b28a2c0d6520334a153d

> I started to look into this and if I understood well the problem, I'm thinking about a solution using [Filehandle.read](https://nodejs.org/api/fs.html#filehandlereadbuffer-offset-length-position) but it will need some change in the code because then `slice` method need to be async in order to read the desired slice from the file and returning it in a `Blob` instance.

It should be fine to only do the `Filehandle` calls when `.arrayBuffer()` / `.text()` / `.stream()` is called. As for `.size`, if we pass the info to the new blob when creating it, it should be fine as well.

Aschen:
I suggested a `LazyBlob` implementation here https://github.com/huggingface/huggingface.js/pull/104

Feel free to give me any suggestion

PR #104, opened 2023-03-17
Add LazyBlob helper class to lazy load file from filesystem

Aschen:
## Description

This PR add a new helper class `LazyBlob` which aim to prevent loading huge file into RAM.

This new class can be used instead of a `Blob` with the `commit` function.

Fix https://github.com/huggingface/huggingface.js/issues/68

```js
const lazyBlob = await LazyBlob.create("./package.json");

await commit({
  repo,
  title: "Some commit",
  credentials: {
    accessToken: "<access token>",
  },
  operations: [
    {
      operation: "addOrUpdate",
      content: lazyBlob,
      path: "package.json",
    },
  ],
});
```


packages/hub/src/lib/LazyBlob.ts
```diff
@@ -0,0 +1,84 @@
+import { createReadStream } from "fs";
+import { open } from "fs/promises";
+import type { FileHandle } from "fs/promises";
```
coyotte508, line 3:
```suggestion
import { createReadStream } from "node:fs";
import { open } from "node:fs/promises";
import type { FileHandle } from "node:fs/promises";
```

packages/hub/src/lib/LazyBlob.ts
```diff
@@ -0,0 +1,84 @@
...
+	}
+
+	async text(): Promise<string> {
+		const buffer = (await this.arrayBuffer()) as Buffer;
+
+		return buffer.toString("utf8");
+	}
+
+	stream(): ReadableStream {
+		const stream = createReadStream(this.path);
```
coyotte508, line 80:
```suggestion
		return Readable.toWeb(createReadStream(this.path));
```

And:

```ts
import { Readable } from "node:stream";
import type { ReadableStream } from "node:stream/web";
```

coyotte508 (commented):
This is going to be really cool when ready!

My main request is that from the outside, it should behave exactly as a `Blob`. No need for `.init()` or `dispose()`.

packages/hub/src/lib/LazyBlob.ts
```diff
@@ -0,0 +1,84 @@
+import { createReadStream } from "fs";
+import { open } from "fs/promises";
+import type { FileHandle } from "fs/promises";
+
+export class LazyBlob {
```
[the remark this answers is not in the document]
Aschen: I'm not sure about this since the `since` methods will have different signatures (promise and not promise)

Aschen:
@coyotte508 I can get ride of `init` with the private constructor and static `create` method as you suggest but for the dispose the only option would be to `open` the file when reading and slice and then `close` it.

Here it's DX vs performances so I will let you choose :+1:

packages/hub/src/lib/LazyBlob.ts
```diff
@@ -0,0 +1,84 @@
...
+
+		await this.file.close();
+	}
+
+	async slice(start: number, end: number): Promise<Blob> {
+		if (this.file === null) {
+			throw new Error("LazyBlob has not been initialized");
+		}
+
+		const size = end - start;
```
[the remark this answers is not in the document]
Aschen: About negative values for `end` and `start`, althought it's possible to do this, it's not trivial because it would require in some case to read the file 2 times. 

For example on a file of size 10:
 - `slice(5, -2)`: read from 5 to 10, then from 0 to 1
 
To be honest, I'm not sure it's necessary to mimic the whole `Blob` class since we could just stick to the usecase we have. If the feature is needed later then we can add it at the cost of more complexity .

packages/hub/src/lib/LazyBlob.ts
```diff
@@ -0,0 +1,84 @@
...
+		const slice = await this.file.read(Buffer.alloc(size), 0, size, start);
```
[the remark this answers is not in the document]
Aschen: I have the same comment about minimal working `LazyBlob` for our usecase. If needed, this feature could be added later.

packages/hub/src/lib/LazyBlob.ts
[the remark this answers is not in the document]
coyotte508: With the other comments, `.size` and `.slice` should be able to be generated synchronously without promises:

- size is given at init
- For `.slice`, a `new LazyBlob(path, start, end)`;

The static function to create the blob can be: `const fileSize = await ...; return new LazyBlob(path, 0, fileSize)`

coyotte508:
> @coyotte508 I can get ride of `init` with the private constructor and static `create` method as you suggest but for the dispose the only option would be to `open` the file when reading and slice and then `close` it.
> 
> Here it's DX vs performances so I will let you choose :+1: 

The most calls will be three I think, on large files:

- a `.slice(0, 512)` to get a sample to send to `/preupload`
- a `.stream()` to compute the `Sha`
- Then whatever `fetch` will call (probably `.stream()`)

On smaller files, it will be two calls: one `.slice(0, 512)` and one `.arrayBuffer()`.

Given that, I'd rather prioritize DX on this one :)

Aschen:
> Given that, I'd rather prioritize DX on this one ! The overhead should be low. Maybe an optimization would be to directly load small files in memory, if we really care about perf.

I agree with that :+1:  Also, better have more syscall to open/close than unused file descriptor :see_no_evil:

packages/hub/src/lib/LazyBlob.ts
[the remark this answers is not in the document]
Aschen: Let me implements that :+1:

packages/hub/src/utils/sha256.ts
coyotte508, line 1: This file should be able to be reverted

packages/hub/src/lib/commit.ts
coyotte508, line 1: This file should now also be able to be reverted (maybe just remove the todo comment)

packages/hub/src/lib/LazyBlob.ts
```diff
@@ -0,0 +1,92 @@
...
+import { Readable } from "node:stream";
+
+export class LazyBlob extends Blob {
+	static async create(path: string, start?: number, end?: number): Promise<LazyBlob> {
+		const lazyBlob = new LazyBlob(path, start, end);
+
+		await lazyBlob.init();
+
+		return lazyBlob;
+	}
```
coyotte508, line 13:
```suggestion
	static async create(path: string): Promise<LazyBlob> {
		const { size } = await stat(path);
		
		const lazyBlob = new LazyBlob(path, 0, size);

		return lazyBlob;
	}
```

That way `LazyBlob.start` and `LazyBlob.end` will always be defined. It'll remove the need for `LazyBlob.totalSize` too.

packages/hub/src/lib/LazyBlob.ts
```diff
@@ -0,0 +1,92 @@
...
+
+		return this.totalSize;
+	}
+
+	get type(): string {
+		return "";
+	}
+
+	slice(start = 0, end = this.size): LazyBlob {
+		const slice = new LazyBlob(this.path, start, end);
```
coyotte508, line 47:
```suggestion
		const slice = new LazyBlob(this.path, this.start + start, Math.min(this.start + end, this.end));
```

We should also throw `new TypeError("Unsupported negative start/end on LazyBlob.slice")`  if `start < 0` or `end < 0`, in case we don't want to support this behavior (we don't really need to support it for now)

packages/hub/src/lib/LazyBlob.ts
```diff
@@ -0,0 +1,92 @@
...
+		return this.slice(0, this.size);
+	}
+
+	async arrayBuffer(): Promise<ArrayBuffer> {
+		const start = this.start || 0;
+		const end = this.end || this.size;
+		const size = Math.abs(end - start);
+		const cappedSize = size > this.size ? this.size : size;
+
+		const slice = await this.execute((file) => file.read(Buffer.alloc(cappedSize), 0, cappedSize, 0));
```
coyotte508, line 62:
```suggestion
		const slice = await this.execute((file) => file.read(Buffer.alloc(this.size), 0, this.size, this.start));
```

With the other requested change on the blob creation, it should be possible to simplify the code to this

packages/hub/test.js
coyotte508, line 1: I think this belongs to another PR :)

coyotte508 (commented):
Getting close!

packages/hub/test.js
Aschen: Oopsi :see_no_evil:

packages/hub/src/lib/LazyBlob.ts
```diff
@@ -0,0 +1,87 @@
...
+
+	stream(): ReadableStream {
+		return Readable.toWeb(createReadStream(this.path));
+	}
+
+	private async init(): Promise<void> {
+		const { size } = await this.execute((file) => file.stat());
+
+		this.totalSize = size;
+	}
```
coyotte508, line 76: This can be removed

packages/hub/src/lib/LazyBlob.ts
```diff
@@ -0,0 +1,87 @@
...
+
+	async text(): Promise<string> {
+		const buffer = (await this.arrayBuffer()) as Buffer;
+
+		return buffer.toString("utf8");
+	}
+
+	stream(): ReadableStream {
+		return Readable.toWeb(createReadStream(this.path));
```
coyotte508:
```suggestion
		return Readable.toWeb(createReadStream(this.path, {start: this.start, end: this.end}));
```

coyotte508:
For the TS Error: maybe `as ReturnType<Blob["stream"]>`

coyotte508 (commented):
I think it's ready after the changes.

I just can't believe this module doesn't already exist on npm (only versions I found loaded everything in memory)

packages/hub/src/lib/LazyBlob.ts
```diff
@@ -0,0 +1,87 @@
...
+
+		const lazyBlob = new LazyBlob(path, 0, size);
+
+		return lazyBlob;
+	}
+
+	private path: string;
+	private start: number;
+	private end: number;
+	private totalSize: number;
```
coyotte508, line 18: This can be removed as well

packages/hub/src/lib/LazyBlob.ts
```diff
@@ -0,0 +1,87 @@
...
+	get size(): number {
+		if (this.start !== null) {
+			if (this.end !== null) {
+				return Math.abs(this.end - this.start);
+			}
+
+			return this.totalSize - this.start;
+		}
+
+		return this.totalSize;
```
coyotte508, line 39:
```suggestion
		return this.end - this.start;
```

packages/hub/src/lib/LazyBlob.ts
Aschen: I thought about this one on my way back from the beach today :grin:

coyotte508 (approved):
Alirght, thanks a lot!

coyotte508:
> @coyotte508 What about the documentation? If I understood well it's in the same repo, I can give it a try

I intend to keep it internal. 

 A file with `require("node:fs")` is directly exposed in the exports, it will cause trouble for the use in browsers, at least for the CJS bundle. See how `sha256-node` is currently dynamically imported and only outside the browser.

So I think maybe a smart way to expose it is to allow passing a `URL` as an alternative to a `Blob` - see https://www.geeksforgeeks.org/node-js-url-pathtofileurl-api/ urls and files paths can be equivalent.

It also opens the door for supporting web urls later on as well. (using Range header, or HEAD calls appropriately to get a slice / the size)

Aschen:
Yes I saw, I'm familiar with this kind of dual behavior when it's needed to provide a SDK for both node and the browser (See [Kuzzle Javascript SDK](https://github.com/kuzzleio/sdk-javascript/))

If I understand well, you might want to allows a string representing a path to a local file for `ContentSource` if it's used in Node.js ?

```
await commit({
  repo,
  title: "Some commit",
  credentials: {
    accessToken: TEST_ACCESS_TOKEN,
  },
  operations: [
    {
      operation: "addOrUpdate",
      // 1) Only a path to a local file, if it's used from Node.js, a LazyBlob will be used
      path: "test.txt",
    },
    {
      operation: "addOrUpdate",
      // 2) URL to a local file, if it's used from Node.js, a LazyBlob will be used
      path: "file://test.txt",
    },
  ],
});
```

coyotte508:
> If I understand well, you might want to allows a string representing a path to a local file for `ContentSource` if it's used in Node.js ?

When I was mentioning `URL`, it was literally :)

Basically, change:

```ts
type ContentSource = Blob;
```

To

```ts
type ContentSource = Blob | URL;
```

Aschen:
Ok understood :+1: 

So if the URL is a local file (starting with `file://`) then the `LazyBlob` will be used, am I right?

Aschen:
Do you want me to do this feature, supporting URL as content, in a separate PR?

coyotte508:
> Do you want me to do this feature, supporting URL as content, in a separate PR?

Sure! Then maybe you can remove the export in this one and keep the `LazyBlob` private ? (I was going to do it later today before merging :grin: )

Issue #91, opened 2023-03-11
Fake inference server to mock tests


coyotte508:
We should run a fake inference server when `env.HF_ACCESS_TOKEN` is undefined.

That way, PRs from works can run tests fine, and we still run real tests on the main repo.

Aschen:
Hello :wave: 

What kind of solution do you have in mind for this?

On my side, I would suggest a [VCR-like](https://github.com/LinkedInAttic/sepia) approach (don't look at the scary `9 years ago`), most likely it would require to:
  - inject the service responsible of the final request (for now it's [fetch](https://github.com/huggingface/huggingface.js/blob/main/packages/inference/src/HfInference.ts#L699)) 
  - decide how to identify an HTTP request (for example `hash(url + verb + body + headers)`)
  - run tests to save answers from HF Inference API in JSON files 
  - commit files and run futur tests with recorded answers in JSON files

coyotte508:
Hey!

In my mind it was to record the responses as you said as a first step. Then run a local server (using express, koa, whichever) that replays those responses depending on the request. The local server can be started / closed in a [globalSetup](https://vitest.dev/config/#globalsetup) file.

Then, in the tests, if HF_ACCESS_TOKEN is defined, call the local server instead. We can add an optional url in the `Options` param for example that can be set to a local server during tests.

Aschen:
I think it would be better to not even use a local server.

For example, we could replace the original `fetch` method to intercept and hash the parameter, do the request with the original fetch and finally save the result in the form of `<hash>.json`

So the next time the same request is executed, the proxy version of `fetch` will first check if a response having the same hash is present.

```js
if (!process.env.HF_ACCESS_TOKEN) {
	const originalFetch = global.fetch;

	global.fetch = (...args) => {
		console.log(args);
		return originalFetch(...args);
	};
}

async function run() {
	await fetch("https://aschen.tech");
}

run();
```

coyotte508:
As long as it works :) 

We should also add a script to regenerate the .json files on command.

Issue #67, opened 2023-03-08
Inference CI from forks


coyotte508:
Forks do not have access to secrets in github actions, so inference tests are rate-limited in the CI.

For now it's fine, but if the rate limit changes or we get more tests it's going to be problematic.

---

Possible solutions:

- Use hard-coded access token from a dummy user for better rate limit, too bad if it's made public
- Use `vitest`'s `--changed` flag to only test what's necessary + split inference source files into many to accomodate it
- ?

Not urgent anyway

Aschen:
@coyotte508 I think you can close this one

coyotte508:
Thanks!

PR #105, opened 2023-03-17
Add a VCR module who intercepts, record and replay API requests

Aschen:
## Description

Add a VCR module inspired by https://github.com/linkedin/sepia.

When imported, this module will be activated by the presence of the `VCR_MODE` environment variable:
 - `record`: external call to API made with `fetch` will be executed and the the response saved to the filesystem
 - `playback`: responses to API calls will be read from the filesystem and an error will be thrown if the answer is missing

This module save the API request responses to the file system and identifies them by using a hash of the input parameters (url + verb + headers + body).

This PR also contains the recorded tapes for current tests so I also edited the CI to run tests using the `playback` mode.

fix https://github.com/huggingface/huggingface.js/issues/91 and https://github.com/huggingface/huggingface.js/issues/67


Aschen:
@coyotte508 I still have problems with the requests made by the `commit.spec.ts` file, I will continue to investigate later

coyotte508:
I'm a bit embarassed to say this after the PR is done, but the issue is on `@huggingface/inference`. There's no issue in `@huggingface/hub`.

> Also, I think it should be considered as a good practice to never run the tests again the real API in the CI.

The [python library](https://github.com/huggingface/huggingface_hub) and this library use a special backend on another environment: hub-ci.huggingface.co. The whole value of the libs is to interact with the Hugging Face hub, so we test against it - the python lib already caught some errors on the API that could've gone undetected.

Aschen:
> I'm a bit embarassed to say this after the PR is done, but the issue is on @huggingface/inference. There's no issue in @huggingface/hub.

Ahah that kind of mistake! It's ok because the logic in the VCR module will works for the other package as well :+1: 

Technically if your hubs have a strong compliance with the no breaking change policy then it's not needed to execute the API requests but I understand if you can to double test it.

I will fix this PR later

Aschen:
@coyotte508 I was working on this but I hit the rate limit before having opportunity to save the tapes. I will try again later, hopefully the limit will be reset

coyotte508:
> @coyotte508 I was working on this but I hit the rate limit before having opportunity to save the tapes. I will try again later, hopefully the limit will be reset

Any token from any free token is good to have a much better rate limit. Also, no worries :)

Aschen:
I generated a token to finish the recording of the tapes.

My main concern is the total size of those tapes because they are quite huge:
```
$ du -h test/tapes 
22M     test/tapes
```

.github/workflows/lint-and-test.yml
```diff
@@ -49,6 +49,6 @@ jobs:
         run: pnpm --filter ...[${{ steps.since.outputs.SINCE }}] type-check
 
       - name: Test
-        run: pnpm --filter ...[${{ steps.since.outputs.SINCE }}] test:ci
+        run: VCR_MODE=playback pnpm --filter ...[${{ steps.since.outputs.SINCE }}] test:ci
```
coyotte508, line 52: I know we have different opinions on this but it should only be playback when `HF_ACCESS_TOKEN` is not defined

coyotte508 (commented):
+1,793,376 lines :grin: 

We should store tapes in LFS

Aschen:
Apparently there is a limitation and I can't push the file to LFS because I'm working on a public fork https://github.com/git-lfs/git-lfs/issues/1906

Lets merge #107 and then I will try one of the workaround (removing and recreating my repo)

Aschen:
@coyotte508 Unfortunately I couldn't push the tapes to LFS :(

What we can do:
  1) merge this PR and then open a new one with the tapes (you just need to run`VCR_MODE=record pnpm test` and commit the tapes)
  2) give me temporarily write access to the repo ("With great power comes great responsibility" would say uncle ben and so I)

Issue #106, opened 2023-03-20
Support `URL` in addition to `Blob` for commited files


coyotte508:
For now, only support URL for local files in Node / Deno / ... context, using the `LazyBlob` class provided by @Aschen .

PR #107, opened 2023-03-20
Allows usage of `URL` pointing to a local file in file operations

Aschen:
## Description

Allows to provide an `URL` object pointing to a local file with `CommitFile` operation.

Providing a non-local file will throw an error.

This is not available in the browser.

Fix https://github.com/huggingface/huggingface.js/issues/106

```
await commit({
  repo,
  title: "Some commit",
  credentials: {
    accessToken: TEST_ACCESS_TOKEN,
  },
  operations: [
    {
      operation: "addOrUpdate",
      content: new URL("file://tsconfig.json"),
      path: "tsconfig.json",
    },
  ],
});
```

### DIFF
--- a/packages/hub/src/lib/commit.ts
+++ b/packages/hub/src/lib/commit.ts
@@ -101,26 +101,35 @@ async function* commitIter(params: CommitParams): AsyncGenerator<unknown, Commit
 
 	const { LazyBlob } = isNode ? await import("../utils/LazyBlob") : { LazyBlob: FakeLazyBlob };
 
-	for (const operation of params.operations) {
-		if (isFileOperation(operation) && operation.content instanceof URL) {
-			if (operation.content.protocol !== "file:") {
-				throw TypeError('Only "file://" protocol is supported for now');
+	const operations = await Promise.all(
+		params.operations.map(async (operation) => {
+			if (isFileOperation(operation) && operation.content instanceof URL) {
+				if (operation.content.protocol !== "file:") {
+					throw TypeError('Only "file://" protocol is supported for now');
+				}
+
+				// Ignore the "file://" at the beginning and the trailing slash
+				const lazyBlob = await LazyBlob.create(operation.content.href.slice(7, operation.content.href.length - 1));
+
+				return {
+					...operation,
+					content: lazyBlob,
+				};
 			}
 
-			// Ignore the "file://" at the beginning and the trailing slash
-			const lazyBlob = await LazyBlob.create(operation.content.href.slice(7, operation.content.href.length - 1));
-		}
-	}
+			return operation;
+		})
+	);
 
 	const gitAttributes = (
-		params.operations.find((op) => isFileOperation(op) && op.path === ".gitattributes") as CommitFile | undefined
+		operations.find((op) => isFileOperation(op) && op.path === ".gitattributes") as CommitFile | undefined
 	)?.content;
 
-	for (const operations of chunk(params.operations.filter(isFileOperation), 100)) {
+	for (const operationsChunk of chunk(operations.filter(isFileOperation), 100)) {
 		const payload: ApiPreuploadRequest = {
 			gitAttributes: gitAttributes && (await (gitAttributes as Blob).text()),
 			files: await Promise.all(
-				operations.map(async (operation) => ({
+				operationsChunk.map(async (operation) => ({
 					path: operation.path,
 					size: (operation.content as Blob).size,
 					sample: base64FromBytes(new Uint8Array(await (operation.content as Blob).slice(0, 512).arrayBuffer())),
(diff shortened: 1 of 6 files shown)


packages/hub/src/lib/commit.ts
```diff
@@ -78,6 +78,49 @@ function isFileOperation(op: CommitOperation): op is CommitFile {
 	return op.operation === "addOrUpdate";
 }
 
+const isBrowser = typeof window !== "undefined" && typeof window.document !== "undefined";
```
coyotte508, line 81: Let's reuse the full check from `sha256.ts`:

```ts
	const isBrowser = typeof window !== "undefined" && typeof window.document !== "undefined";

	const isWebWorker =
		typeof self === "object" && self.constructor && self.constructor.name === "DedicatedWorkerGlobalScope";
```

packages/hub/src/lib/commit.ts
```diff
@@ -78,6 +78,49 @@ function isFileOperation(op: CommitOperation): op is CommitFile {
...
+		const { LazyBlob } = isBrowser ? { LazyBlob: FakeLazyBlob } : await import("../utils/LazyBlob");
+
+		const blobMap = new FileBlobs();
+
+		for (const operation of operations) {
+			if (operation.content instanceof Blob) {
+				blobMap.set(operation.content, operation.content);
+			} else {
+				if (operation.content.protocol !== "file:") {
+					throw TypeError('Only "file://" protocol is supported for now');
```
coyotte508, line 107:
```suggestion
					throw TypeError('Only "file://" protocol is supported for URLs');
```

packages/hub/src/lib/commit.ts
```diff
@@ -89,18 +132,20 @@ async function* commitIter(params: CommitParams): AsyncGenerator<unknown, Commit
 
 	const lfsShas = new Map<string, string | null>();
 
+	const blobs = await FileBlobs.createBlobMap(params.operations.filter(isFileOperation));
```
coyotte508, line 135:
```suggestion
	// Convert `operation.content` to `Blob`
	const operations = await Promise.all(params.operations.map(op => op.operation === "addOrUpdate" ? {...op, content: await ...} : op));
	
	...
```

No need to create/override a separate `Map` class

Aschen: I did it to avoid casting systematically the `content` to `Blob` but I realize that's a very poor design choice :roll_eyes:

coyotte508: (and changing `isFileOperation` too)

Aschen: I tried this but unfortunately since `operation.content` is a `ContentSource`, ts complain that it could be `URL` and not `Blob`.

```js
const operations = await Promise.all(
  params.operations.map(async (operation) => {
    if (isFileOperation(operation) && operation.content instanceof URL) {
      if (operation.content.protocol !== "file:") {
        throw TypeError('Only "file://" protocol is supported for now');
      }

      // Ignore the "file://" at the beginning and the trailing slash
      const lazyBlob = await LazyBlob.create(operation.content.href.slice(7, operation.content.href.length - 1));

      return {
        ...operation,
        content: lazyBlob,
      };
    }

    return operation;
  })
);
```

Aschen:
@coyotte508 About handling file from the browser side, shouldn't we support [File](https://w3c.github.io/FileAPI/) object that can be get when using the file picker API? 

See also https://developer.chrome.com/articles/file-system-access/

packages/hub/src/lib/commit.ts
```diff
@@ -196,9 +235,9 @@ async function* commitIter(params: CommitParams): AsyncGenerator<unknown, Commit
 				}
 
 				if (obj.error) {
-					const errorMessage = `Error while doing LFS batch call for ${operations[shas.indexOf(obj.oid)].path}: ${
-						obj.error.message
-					}${batchRequestId ? ` - Request ID: ${batchRequestId}` : ""}`;
+					const errorMessage = `Error AsyncGeneratorError while doing LFS batch call for ${
+						operations[shas.indexOf(obj.oid)].path
+					}: ${obj.error.message}${batchRequestId ? ` - Request ID: ${batchRequestId}` : ""}`;
```
coyotte508, line 240: Why the change?

packages/hub/src/lib/commit.ts
```diff
@@ -89,11 +99,40 @@ async function* commitIter(params: CommitParams): AsyncGenerator<unknown, Commit
 
 	const lfsShas = new Map<string, string | null>();
 
-	const gitAttributes = (
-		params.operations.find((op) => isFileOperation(op) && op.path === ".gitattributes") as CommitFile | undefined
-	)?.content;
+	const allOperations = await Promise.all(
+		params.operations.map(async (operation) => {
+			if (operation.operation === "addOrUpdate") {
```
Aschen, line 104: Actually here (and l.105 on the second if), you could use a guard to save 2 levels of indentation

coyotte508 (approved):
Nice work!

PR #110, opened 2023-03-22
Add a VCR module who intercepts, record and replay API requests

Aschen:
Add a VCR module inspired by https://github.com/linkedin/sepia.

When imported, this module will be activated by the presence of the `VCR_MODE` environment variable:
 - `record`: external call to API made with `fetch` will be executed and the the response saved to the filesystem
 - `playback`: responses to API calls will be read from the filesystem and an error will be thrown if the answer is missing
 - `cache`: try to load the response from the filesystem and otherwise call the API and save the response

This module save the API request responses into a JSON file and identifies them by using a hash of the input parameters (url + verb + body).

I also updated the test so they work in the headless browser

(Previous obsolete PR https://github.com/huggingface/huggingface.js/pull/105)

fix https://github.com/huggingface/huggingface.js/issues/91 and https://github.com/huggingface/huggingface.js/issues/67


coyotte508:
Contributors would be unable to git push w/ LFS if they add new inference endpoints.

I don't think we test the exact content of the binary responses, so maybe when a blob/buffer is returned we can just store an empty array?

Then we don't need to use LFS for the tapes.

.github/workflows/lint-and-test.yml
```diff
@@ -49,11 +49,15 @@ jobs:
...
+          [ -z "$HF_ACCESS_TOKEN" ] && export VCR_MODE="playback" || export VCR_MODE="disabled"
+          pnpm --filter ...[${{ steps.since.outputs.SINCE }}] test:ci
         env:
           HF_ACCESS_TOKEN: ${{ secrets.HF_ACCESS_TOKEN }}
 
       - name: Test in browser
-        run: pnpm --filter ...[${{ steps.since.outputs.SINCE }}] test:browser
+        run: |
+          [ -z "$HF_ACCESS_TOKEN" ] && export VCR_MODE="playback" || export VCR_MODE="disabled"
+          pnpm --filter ...[${{ steps.since.outputs.SINCE }}] test:browser
```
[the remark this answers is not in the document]
Aschen: Sure

Aschen:
> I don't think we test the exact content of the binary responses, so maybe when a blob/buffer is returned we can just store an empty array in the tape?

I will update the tape accordingly :+1:

coyotte508:
Hm it seems that tapes are now stored with Git LFS (I don't understand!)

Also the lint file was reverted somehow? (using `test:ci` which was removed since).

Thinking about it, if we want to run the tests on browser where we can't use `node:fs` to read the tapes, maybe we can store all the tapes in a single json file? Then `import` it. It should work both in a browser & node context. (might need to change `tsconfig.json` to allow importing json directly).

packages/inference/test/vcr.ts
```diff
@@ -88,45 +87,185 @@ async function vcr(
...
 	}
 
 	return response;
 }
+
+/**
+ * MD5 implementation in pure JS from https://stackoverflow.com/a/60467595
+ * So we can don't have to care about cross platform issues (Node.js vs browser)
+ */
+function md5(inputString) {
```
coyotte508, line 145: You can use ```await crypto.subtle.digest("SHA-256", arrayBuffer)```. 

`crypto.subtle` is available both on node 18+ & browser.

Aschen: Yes but then I still need to check if we are in the backend to import `crypto` or just use it directly if we are in the frontend.

Still, I will use the latest API :+1:

Aschen:
I think I'm almost good with this one, I just have a problem with one of the test because I get an Error `overloaded` with `pnpm test -- -t "questionAnswer"`

I will try again later (I'm already using an auth token)

packages/inference/test/vcr.ts
```diff
@@ -0,0 +1,163 @@
+import { isBackend, isFrontend } from "../src/utils/env-predicates";
+
+const TAPES_FILE = "./tapes.json";
+
+let VCR_MODE;
```
coyotte508, line 5: It should be typed (I guess `tsconfig.json` or `eslint` should be configured to pick this up)

packages/inference/test/vcr.ts
```diff
@@ -0,0 +1,163 @@
...
+	window.fetch = (...args) => {
+		return vcr(originalFetch, args[0], args[1]);
+	};
+} else {
+	const originalFetch = global.fetch;
+
+	global.fetch = (...args) => {
+		return vcr(originalFetch, args[0], args[1]);
+	};
+}
```
coyotte508, line 34: I think instead of `window` and `global` you can use `globalThis` in both cases?

packages/inference/test/vcr.ts
```diff
@@ -0,0 +1,163 @@
...
+ *   - record: requests will be made to the external API and responses will be saved in files
+ *   - playback: answers will be read from the filesystem, if they don't have been recorded before then an error will be thrown
+ *   - cache: same as playback but if the response is not found in the filesystem then it will be recorded
+ */
+async function vcr(
+	originalFetch: typeof global.fetch,
+	input: RequestInfo | URL,
+	init: RequestInit = {}
+): Promise<Response> {
+	let url;
```
coyotte508, line 104: Should be typed

packages/inference/test/vcr.ts
```diff
@@ -0,0 +1,163 @@
...
+
+		const tape: Tape = {
+			url,
+			init: {
+				headers: init.headers,
+				method: init.method,
+			},
+			response: {
+				// Truncating the body to 30KB to avoid having huge files
+				body: Buffer.from(arrayBuffer.slice(0, 30 * 1024)).toString("base64"),
```
coyotte508, line 148:
```suggestion
				body: arrayBuffer.byteLength > 30_000 ? "" : Buffer.from(arrayBuffer).toString("base64"),
```

Since we don't use the large payload may as well not save them at all

coyotte508 (commented):
I think it's close to being merged!

coyotte508 (approved):
Great :)

Issue #109, opened 2023-03-22
Support more URLs in commit call


coyotte508:
Maybe we should then rename `LazyBlob` to `FileBlob` or `LazyFileBlob`? And create `WebBlob`?

That way we'll truly be able to support `URL` in addition to `Blob`. We already do support it but only on node for local files, given the typings are the same on node & browser we should support it on the frontend as well.

There's a real use case here in the demo by @julien-c : https://huggingface.co/spaces/huggingfacejs/push-model-from-web/blob/e0d4ec039a9540dc9aa9fbdffc68d76c5cb2e38c/app.ts#L65

We can use `HEAD` and the `Range` header to be smart about this.

Aschen:
I can give it a try :+1: 

In case the `Range` header is not supported, should I just download the whole file and use a standard `Blob` ?

coyotte508:
> I can give it a try +1
> 
> In case the `Range` header is not supported, should I just download the whole file and use a standard `Blob` ?

Good question... I guess a v1 we can do it rough and maybe optimize somehow in a v2 (to prevent downloading the same file multiple times).

> 3\. The `WebBlob.stream` method may be tricky to implements because we need to return a promise (`fetch` expose a stream in `response.body` but the promise need to resolve first

Maybe you can create a `TransformStream`, and pipe the response to the `writable` part while returning immediately the `readable` part.

Aschen:
> Good question... I guess a v1 we can do it rough and maybe optimize somehow in a v2 (to prevent downloading the same file multiple times).

I don't see why it would be downloaded multiple time unless the action (like click the upload button) is repeated :thinking: (and in this case I think we should re-download) 

> Maybe you can create a TransformStream, and pipe the response to the writable part while returning immediately the readable part.

Unfortunately to access any kind of stream, first an HTTP request should be done and the response awaited. I don't see how it whould be possible to create a stream directly from an URL. (See [here](https://github.com/huggingface/huggingface.js/pull/111/files#diff-d0e720e070055b4f1a7b7ab7dcf160050ddb229ba4bc3ef56965c85f9b18afedR59))

Aschen:
> Something like this? With a way to emit error as well if the fetch fails

Thanks for the hint, I will try something like that (good workshop on stream those features :D )

PR #111, opened 2023-03-22
Support loading of web resources with HTTP protocol

Aschen:
## Description

Add a `WebBlob` utility class that allows to lazy load a resource from an URL using `Range` HTTP requests.

It's now possible to load web resources from an URL using the `http` protocol.

Fix https://github.com/huggingface/huggingface.js/issues/109


Aschen:
@coyotte508 I still need to fix some tests for the `createBlob` function but maybe it's better if you tell me what do you think about it before I invest more time.

Any suggestions are welcome of course :-)

coyotte508:
I think it's fine to not change the commit API. We already allow passing local files through an `URL`, now we can allow passing remote files without changing the API (we just support more protocols).

We don't need to expose a `createBlob` helper, I think? The API would be cleaner IMO without needing to import an helper.

Aschen:
> We don't need to expose a `createBlob` helper, I think? The API would be cleaner IMO without needing to import an helper.

Yes I agree, it's much simpler with URL only :+1:

Aschen:
I have a strange behavior in the headless browser during the test of WebBlob, it's like most of the headers are not present in the `response.headers` object :thinking: 

Headers when I'm running the request in the headless browser
image


Headers when I'm running the request in Chrome
image

packages/hub/src/lib/commit.spec.ts
```diff
@@ -72,19 +80,33 @@ describe("commit", () => {
...
+				);
+			}
+
+			const webResourceContent = await downloadFile({ repo, path: "lamaral.json" });
+			assert.strictEqual(webResourceContent?.status, 200);
+			assert.strictEqual(
+				await webResourceContent?.text(),
+				await (
+					await fetch("https://huggingface.co/spaces/aschen/push-model-from-web/raw/main/mobilenet/model.json")
+				).text()
```
coyotte508, line 109: You can put the URL in a constant to avoid repeating it

packages/hub/src/lib/commit.ts
```diff
@@ -91,66 +91,35 @@ function isFileOperation(op: CommitOperation): op is CommitBlob {
...
-export function createBlob(url: string): Promise<Blob> {
+async function createBlob(url: URL): Promise<Blob> {
 	if (isFrontend) {
-		if (url.startsWith("file://")) {
+		if (url.protocol === "file:") {
 			throw new TypeError("File URLs are not supported in browsers");
 		}
 
-		return WebBlob.create(new URL(url));
+		return WebBlob.create(url);
```
coyotte508: Should we throw if the protocol is not `http:` or `https:`?

packages/hub/package.json
```diff
@@ -28,7 +28,7 @@
 		"prepublishOnly": "npm run build",
 		"build": "tsup src/index.ts --format cjs,esm --clean",
 		"test": "vitest run",
-		"test:browser": "vitest run --browser.name=chrome --browser.headless --config ./vitest-browser.config.ts",
+		"test:browser": "vitest run --browser.name=chrome --browser.headless --config ./vitest-browser.config.ts ",
```
coyotte508, line 31:
```suggestion
		"test:browser": "vitest run --browser.name=chrome --browser.headless --config ./vitest-browser.config.ts",
```

packages/hub/pnpm-lock.yaml
coyotte508, line 1: Revert this file by doing `pnpm install`?

It's already installed in the root package

coyotte508:
> I have a strange behavior in the headless browser during the test of WebBlob, it's like most of the headers are not present in the `response.headers` object thinking
> 
> If you have no idea about this issue I think I will post on vitest repo
> 
> Headers when I'm running the request in the headless browser image
> 
> Headers when I'm running the request in Chrome image

It's probably related to CORS & exposed headers. Will look into fixing it from our end

packages/hub/README.md
```diff
@@ -52,23 +64,7 @@ await deleteRepo({ repo, credentials });
...
-  operations: [
-    {
-      operation: "addOrUpdate",
-      path: 'large-file.bin',
-      content: pathToFileURL("./large-file.bin"),
-    },
-  ],
-});
-```
+Web resources and local files should be passed as `URL` whenever it's possible so they will be lazy loaded in chunks to reduce RAM usage.
```
coyotte508, line 67:
```suggestion
Remote resources and local files should be passed as `URL` whenever it's possible so they can be lazy loaded in chunks to reduce RAM usage. Passing a `File` inside the browser's context is fine, because it natively behaves as a `Blob`.
```

packages/hub/src/lib/commit.ts
```diff
@@ -88,6 +90,38 @@ function isFileOperation(op: CommitOperation): op is CommitBlob {
...
+		}
+
+		throw new TypeError(`Unsupported URL protocol "${url.protocol}"`);
+	}
+
+	if (url.protocol.startsWith("http")) {
+		return WebBlob.create(url);
+	}
+
+	if (url.protocol.startsWith("file:")) {
```
coyotte508, line 116:
```suggestion
	if (url.protocol === "file:") {
```

packages/hub/src/lib/commit.ts
```diff
@@ -88,6 +90,38 @@ function isFileOperation(op: CommitOperation): op is CommitBlob {
...
+async function createBlob(url: URL): Promise<Blob> {
+	if (isFrontend) {
+		if (url.protocol.startsWith("http")) {
+			return WebBlob.create(url);
+		}
+
+		throw new TypeError(`Unsupported URL protocol "${url.protocol}"`);
+	}
+
+	if (url.protocol.startsWith("http")) {
```
coyotte508, line 112:
```suggestion
	if (url.protocol === "http:" || url.protocol === "https:") {
```

coyotte508 (commented):
Looks good, will test tomorrow, also with changes to the accept-ranges on HF's side

coyotte508:
> As far as I saw in my tests, Moon already support `Accept-Ranges: bytes`, are you talking about the CORS issue?

Yes, now `Content-Range` and `Accept-Range` should be exposed

packages/hub/src/utils/WebBlob.spec.ts
```diff
@@ -0,0 +1,51 @@
+import { describe, expect, it } from "vitest";
+import { WebBlob } from "./WebBlob";
+
+describe("WebBlob", async () => {
+	const resourceUrl = new URL("https://huggingface.co/spaces/aschen/push-model-from-web/raw/main/mobilenet/model.json");
```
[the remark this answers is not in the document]
Aschen: I'm not sure about the purpose of such test since we already tested the `slice` with a file hosted on HF. Since it's only using standard HTTP protocol then I think it should works everywhere and if it doesn't then it's not our reponsability but rather a mistake in the HTTP support from the server.

If you want I can add a test but I will add it in the `commit.spec.ts` file instead since we are uploading a LFS file there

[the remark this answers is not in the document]
coyotte508: > I'm not sure about the purpose of such test since we already tested the slice with a file hosted on HF. 

It's two different paths, for large files we do not proxy them but _redirect_ to S3 / Cloudfront
 
> I think it should works everywhere and if it doesn't then it's not our reponsability but rather a mistake in the HTTP support from the server.

I agree in an ideal world it's not the lib's responsibility but the AWS configuration / devops team, still loading large files from HF is a significant use case, and an early warning system if it stops being true is appreciated. 

From HF's perspective, what matter is that things work, and even if it's an AWS CORS misconfiguration that stops the lib from working as it should, we want to detect / bubble up the error fast.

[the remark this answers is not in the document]
Aschen: Understood :+1: I will add a test for a LFS file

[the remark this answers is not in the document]
coyotte508: yea my initial comment wasn't super clear either, I meant to add a test to see if it's loaded as a lazy blob or not.

The "slice" comment was just in case we wanted to test data downloaded for the LFS file, we don't need to download the whole LFS file (since they can be several GBs) just test on a small portion

packages/hub/src/lib/commit.ts
```diff
@@ -88,6 +89,38 @@ function isFileOperation(op: CommitOperation): op is CommitBlob {
...
+		if (url.protocol === "http:" || url.protocol === "https:") {
+			return WebBlob.create(url);
+		}
+
+		throw new TypeError(`Unsupported URL protocol "${url.protocol}"`);
+	}
+
+	if (url.protocol === "http:" || url.protocol === "https:") {
+		return WebBlob.create(url);
+	}
```
coyotte508, line 113:
```suggestion
	if (url.protocol === "http:" || url.protocol === "https:") {
		return WebBlob.create(url);
	}
	
	if (isFrontend) {
		throw new TypeError(`Unsupported URL protocol "${url.protocol}"`);
	}
```

coyotte508 (approved):
Yay!

PR #118, opened 2023-03-27
Inference request headers: `X-Use-Cache` & `X-Load-Model`

### LINKED ISSUE #115: Add support for X-Load-Model header
Setting the `X-Load-Model` header to "0" means that we do not want to load the model if it's not already available.

Could be exposed through a parameter `doNotLoadModel` / `onlyIfAlreadyLoaded` / ... .

cc @mishig25 @radames

### LINKED ISSUE #116: Add support for skipping cache
Currently cache can be skipped by setting the `X-Use-Cache` header to `"false"`

~~But maybe we can also make it skip cache with `Cache-control: no-cache` which is more standard? Cc @Narsil~~

Also cc @mishig25

mishig25:
Closes #115
Closes #116

### DIFF
--- a/packages/inference/src/HfInference.ts
+++ b/packages/inference/src/HfInference.ts
@@ -701,7 +701,7 @@ export class HfInference {
 		}
 
 		if (options?.binary) {
-			if (mergedOptions.wait_for_model){
+			if (mergedOptions.wait_for_model) {
 				headers["X-Wait-For-Model"] = "true";
 			}
 			if (mergedOptions.use_cache === false) {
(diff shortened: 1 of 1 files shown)


packages/inference/src/HfInference.ts
```diff
@@ -697,6 +701,12 @@ export class HfInference {
 		if (options?.binary && mergedOptions.wait_for_model) {
 			headers["X-Wait-For-Model"] = "true";
 		}
+		if (options?.binary && mergedOptions.use_cache === false) {
+			headers["X-Use-Cache"] = "false";
+		}
+		if (options?.binary && mergedOptions.dont_load_model) {
+			headers["X-Load-Model"] = "0";
+		}
```
coyotte508, line 709: Maybe those could be grouped inside a `if (options?.binary) {`

mishig25: handled in be4ee36e3a97c0744efd00887b2390caf2de3140

Issue #114, opened 2023-03-26
Use URL and FileBlob/WebBlob to load binary data in Inference API instead of Blob


Aschen:
The Inference API use Blob or ArrayBuffer as argument for the request who result in loading huge file into RAM before sending them to the API.

We should rather use the new FileBlob or LazyBlob as for the `commit` method of the Hub API.

I already started to externalize the first version of the `FileBlob` https://github.com/aschen/lazyblob

I would also understand if you want to:
  1) keep the package in the `huggingface` namespace 
  2) create a shared package between `hub` and `inference` package in this monorepo.

coyotte508:
I think publishing the package is neat and you did most of the work, so definitely go ahead! It'll be useful to the community / nice to show off :).

Still, for our usage we'll want to keep it in an internal local internal package.

Aschen:
This work was teamwork and I had good advices on a topic I rarely encounter ;-)

For our matter, I suggest to create an internal package then (if later I have more time then I will publish)

Shall I create a `packages/shared` for the code shared between both libraries?

coyotte508:
I guess `packages/utils`? Taking inspiration from https://github.com/typescript-eslint/typescript-eslint/tree/main/packages/utils

Let's do this after merging the two PRs (web blobs & VCR), just in case it interferes

Aschen:
Sounds good :+1: 

Do we really need to publish it as a separate package on NPM?

The alternative would be to import the classes directly from a relative path. Additional chunks will be added for the code coming from the `packages/utils` package

![image](https://user-images.githubusercontent.com/4447392/228095043-79339397-ed28-4b97-afc6-ea0b218cb6b5.png)

coyotte508:
Yes I don't want to add it as a separate package, either importing the files directly works, or changing the `tsup` config (or associated `rollup` config)

Aschen:
:100: agree with KISS :+1: 

I wil try something

coyotte508:
Note that you can add internal dependencies in pnpm with `workspace:*`, just they will appear in the release's package.json by default

coyotte508:
And if it's too much headache, a shared folder like you suggested is :+1:

Aschen:
> Note that you can add internal dependencies in pnpm with `workspace:*`, just they will appear in the release's package.json by default

I'm not sure to understand what you mean. 

[...]

PR #121, opened 2023-03-28
Support lazy blob in Inference

Aschen:
## Description

Methods from inference package now support `URL` as body data in addition to `Blob` and `ArrayBuffer`.

Fix https://github.com/huggingface/huggingface.js/issues/114

### Other changes

Add an `packages/shared` package to share utilities between published packages.

This package is not published but only used internaly. Files used from this package into published packages will be built as well and included as chunks.

The shared package include:
  - `createBlob` utility function
  - `base64FromBytes`
  - `isFrontend` and `isBackend`
  
WebBlob and FileBlob are only exposed as types intentionnaly because they should only be created using `createBlob`

### Boyscout

 - split commit tests in separate steps


Aschen:
@coyotte508 I started to move shared utilities into a `shared` internal/local package.

(Sorry for the mess, I started from https://github.com/huggingface/huggingface.js/pull/111 which is not merged into main yet, I will clean the history when it's merged)

packages/hub/src/lib/commit.spec.ts
```diff
@@ -55,129 +127,77 @@ describe("commit", () => {
...
-				title: "upload model",
-				operations,
-			});
-
-			const LFSSize = (await fileDownloadInfo({ repo, path: "mobilenet/group1-shard1of2" }))?.size;
+		const LFSSize = (await fileDownloadInfo({ repo, path: "mobilenet/group1-shard1of2" }))?.size;
 
-			assert.strictEqual(LFSSize, 4_194_304);
-		} finally {
-			if (!Math.random()) {
```
Aschen, line 171: Why was it there ?

packages/shared/package.json
```diff
@@ -2,6 +2,16 @@
...
+	"main": "./dist/index.js",
+	"module": "./dist/index.mjs",
+	"types": "./dist/index.d.ts",
+	"exports": {
+		".": {
+			"require": "./dist/index.js",
+			"import": "./dist/index.mjs",
+			"types": "./dist/index.d.ts"
+		}
+	},
```
coyotte508, line 14: We don't need all this (not even the package.json) if we import the files directly (without using `@huggingface/shared`)

packages/shared/src/is-supported-by-fetch.ts
```diff
@@ -0,0 +1,6 @@
+import { isFrontend } from "./env-predicates";
+import { WebBlob } from "./WebBlob";
+
+export function isSupportedByFetch(blob: Blob) {
+	return blob instanceof WebBlob && isFrontend;
```
coyotte508: I think it's the opposite condition

pnpm-workspace.yaml
```diff
@@ -2,3 +2,4 @@ packages:
   - "packages/hub"
   - "packages/inference"
   - "packages/doc-internal"
+  - "packages/shared"
```
coyotte508, line 5: If there's no package.json, I think this line shouldn't be here

coyotte508 (commented):
I like the refacto of `commit.spec.ts`!

Regarding everything moved to `shared`, I'd like to only move things there once they are used by multiple packages, not preemptively.

Maybe you can start by functions used by the VCR module in this PR (and convert the VRC module to use them), and do another PR later with the blob classes + using them in inference?

coyotte508:
> It's actually tricky to move the VCR module since it rely on a relative `import` and `writeFileSync` and anyway it's used only by the inference API.

Oh no, I meant moving the functions used by the VCR module, not the VCR module itself. For example `base64fromBytes` could replace `Buffer.from(...).toString("base64")` that is used in the VCR module, and it also uses `isFrontend` and `isBackend`

packages/hub/src/lib/commit.spec.ts
coyotte508: I use it to write code that returns/throws without the linter being unhappy about unreachable code

packages/inference/test/tapes.json
```diff
@@ -388,5 +388,71 @@
...
+        "content-security-policy": "default-src none; sandbox",
+        "content-type": "image/png",
+        "etag": "\"6a6262641f05cb63ecb2851d5d5b73eabb99d328\"",
+        "server": "nginx",
+        "strict-transport-security": "max-age=31536000; includeSubDomains",
+        "vary": "Origin",
+        "via": "1.1 1de7ecec44e546a1e71d662ee3f2ba42.cloudfront.net (CloudFront)"
+      }
+    }
   }
```
coyotte508: This one shouldn't be taped, only calls to `api-inference`

packages/inference/test/tapes.json
coyotte508, line 1: the file was 100% diff'd (indent changes)

packages/shared/package.json
```diff
@@ -21,6 +29,7 @@
 	"devDependencies": {
 		"@types/node": "^18.13.0",
 		"ts-node": "^10.9.1",
+		"tsup": "^6.6.3",
```
coyotte508, line 32: This can probably be moved to the root (like vitest), maybe `ts-node` too?

packages/shared/README.md
```diff
@@ -3,3 +3,5 @@
 This local package contains utilities shared by the other packages.
 
 It is not meant to be published to npm.
+
+This package need to be built and it's export correctly set up in order to be used by the other packages in browser tests.
```
coyotte508, line 7: Hm, are you sure? Maybe adding `../../shared` to `tsconfig.json` inside hub/inference woudl work?

Aschen: It didn't really worked. Actually I'm not confident about releasing any of this because I built the package (`pnpm pack`) and try to use it with the Create React App and I have errors about having node related imports

```
Failed to compile.

Module build failed: UnhandledSchemeError: Reading from "node:fs/promises" is not handled by plugins (Unhandled scheme).
Webpack supports "data:" and "file:" URIs by default.
You may need an additional plugin to handle "node:" URIs.
ERROR in node:fs/promises
Module build failed: UnhandledSchemeError: Reading from "node:fs/promises" is not handled by plugins (Unhandled scheme).
Webpack supports "data:" and "file:" URIs by default.
You may need an additional plugin to handle "node:" URIs.
    at /tmp/my-app/node_modules/webpack/lib/NormalModule.js:835:25
    at Hook.eval [as callAsync] (eval at create (/tmp/my-app/node_modules/tapable/lib/HookCodeFactory.js:33:10), <anonymous>:6:1)
    at Object.processResource (/tmp/my-app/node_modules/webpack/lib/NormalModule.js:832:8)
    at processResource (/tmp/my-app/node_modules/loader-runner/lib/LoaderRunner.js:220:11)
    at iteratePitchingLoaders (/tmp/my-app/node_modules/loader-runner/lib/LoaderRunner.js:171:10)
    at runLoaders (/tmp/my-app/node_modules/loader-runner/lib/LoaderRunner.js:398:2)
    at NormalModule._doBuild (/tmp/my-app/node_modules/webpack/lib/NormalModule.js:822:3)
    at NormalModule.build (/tmp/my-app/node_modules/webpack/lib/NormalModule.js:966:15)
    at /tmp/my-app/node_modules/webpack/lib/Compilation.js:1371:12
    at NormalModule.needBuild (/tmp/my-app/node_modules/webpack/lib/NormalModule.js:1257:26)
  ```
  
  I really don't understand because then I manually deleted all references to any node import in `node_modules/@huggingface/inference` but the bug is still there :upside_down_face:

coyotte508: do those problems happen with the current version? We have problems with some CDNs like skypack

coyotte508: opened #137 

It only impacts `@huggingface/hub` for now, it should definitely be fixed before adding this to `@huggingface/inference` which is the more impactful lib

Aschen:
@coyotte508 I feel sorry to leave this on the way but I'm taking a long holiday and I won't be working anymore.
btw it was cool to discover some new things here ;-)

See you :wave:

coyotte508:
Ok @Aschen it was nice working with you, thanks for everything :)

Closing the PR for now

PR #124, opened 2023-04-03
Missing `await` ?

mishig25:
`await` is probably missing on that line


coyotte508 (approved):
In practice it's not needed, when returning promises, `await` doesn't change the behavior (except maybe a little more overhead / debug callstack)

Feel free to merge!

Aschen:
Just my 2 cents here, adding await will create 2 additional promises on the event loop (that will resolve on next ticks)

It only made sens to have `return await` when you want to catch early:
```
try {
  return await ...;
} catch (error) {
  
}
```

coyotte508:
Thanks for the info!

In any case, https://github.com/huggingface/huggingface.js/pull/125 overwrites this part of the code so this was a temporary change x)

PR #125, opened 2023-04-03
Check if api inference output is correct

mishig25:
This PR adds checks to api inference calls:
1. Checks if api inference outputs are as expected (for example, TextClaissification api output should be `array<{score: number, label: string}>`)
2. If not, throw `new TypeError`


coyotte508:
It seems there's a failing test on the shape of `zeroshot_classification`'s response

coyotte508 (approved):
Thanks!

coyotte508:
Released in `@huggginface/inference` `v1.6.3`

PR #130, opened 2023-04-05
Add token streaming for text generation

### LINKED ISSUE #78: Enable token streaming for text-generation
We have a few models that already support token streaming. 
We could add an extra options
```
export interface Options { 
	streaming?: boolean;
....
}
```

Here's an example by @julien-c @olivierdehaene using an improved API for [`EventSource`](https://developer.mozilla.org/en-US/docs/Web/API/EventSource) with [`fetch-event-source`](https://github.com/Azure/fetch-event-source). 
https://gist.github.com/julien-c/9565fbc85d0baecee9f910017584d157

Available models

```
[
    "Rallio67/joi2_20B_instruct_alpha",
    "google/flan-t5-xxl",
    "google/flan-ul2",
    "bigscience/bloom",
    "bigscience/bloomz",
    "EleutherAI/gpt-neox-20b"
]
```

vvmnnnkv:
An attempt to fix #78 :-)

Added a separate method `textGenerationStream` (similar to python client).

Not using `fetch-event-source` because it doesn't play well with nodejs (https://github.com/Azure/fetch-event-source/issues/20).
But should be easy to replace in the future.


coyotte508:
You can run `pnpm -r format` at the root to fix lint issues!

julien-c:
> Not using `fetch-event-source` because it doesn't play well with nodejs ([Azure/fetch-event-source#20](https://github.com/Azure/fetch-event-source/issues/20)).

Oh i missed that! We don't use it from Node in moon-landing, @SBrandeis?

coyotte508 (commented):
Thanks a lot for this PR @vvmnnnkv !

There's a big refacto request (streams => async generators), it's maybe more than you bargained for when opening the PR ^^

packages/inference/src/HfInference.ts
```diff
@@ -615,6 +699,16 @@ export class HfInference {
...
 	}
 
+	/**
+	 * Use to continue text from a prompt. Same as `textGeneration` but returns stream that can be read one token at a time
+	 */
+	public async textGenerationStream(
+		args: TextGenerationArgs,
+		options?: Options
+	): Promise<ReadableStream<TextGenerationStreamReturn>> {
+		return await this.streamingRequest<TextGenerationStreamReturn>(args, options);
```
Aschen, line 709: It's really nitpicking but this `await` is useless and create unnecesary promises on the event loop

packages/inference/src/HfInference.ts
```diff
@@ -899,4 +993,88 @@ export class HfInference {
...
+
+		if (mergedOptions.retry_on_error !== false && response.status === 503 && !mergedOptions.wait_for_model) {
+			return this.request(args, {
+				...mergedOptions,
+				wait_for_model: true,
+			});
+		}
+		if (!response.ok) {
+			throw new Error(`Server response contains error: ${response.status}`);
+		}
```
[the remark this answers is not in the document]
vvmnnnkv: Extracted most of it into options preparation helper. For error/re-try logic, it will be quite different if we switch to `fetch-event-source` someday, so left it in individual methods for now.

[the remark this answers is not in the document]
coyotte508: Thanks!

packages/inference/src/utils/Uint8ToSseParser.ts
coyotte508, line 1: I think there's several pitfalls when dealing with low-level code like this, code review can be complicated.

For example the line of a SSE can be of the format `"<field>:<value>"` (handled correctly at the moment) or `"<field>: <value>"` (handled incorrectly)

We're lucky to have a TS implementation in [`parse.ts`](https://github.com/Azure/fetch-event-source/blob/main/src/parse.ts) from `Azure/fetch-event-source` which is battle-tested and readable, I think we should copy the file (with adding their license at the top of the file + linking to original file) and reuse it

vvmnnnkv: OK :)

coyotte508 (approved):
Thanks a lot!  :hugs: 

cc @radames @TimMikeladze  :)

coyotte508:
A demo is live with the feature: https://huggingface.co/spaces/huggingfacejs/streaming-text-generation

It's released in version 1.7.1.

julien-c:
nice demo!

PR #131, opened 2023-04-09
Image to text

### LINKED ISSUE #13: Image-to-text

vvmnnnkv:
This should close #13

While #114 is not ready yet, passing local image file to browser using `define`.


packages/inference/vitest.config.ts
```diff
@@ -0,0 +1,11 @@
...
+import { readFileSync } from "node:fs";
+import { join } from "node:path";
+
+export default defineConfig({
+	define: {
+		__TEST_FILES__: JSON.stringify({
+			"cheetah.png": readFileSync(join(__dirname, "test", "cheetah.png")).toString("base64"),
+		}),
+	},
+});
```
coyotte508, line 11: This is a neat way to test on both browser & front-end. 

Do you think you can change the other HFInference tests to work on the frontend too? :) (audio-classification, ...)

coyotte508 (commented):
Thanks, nice!

It'll be a great addition to the lib

packages/inference/test/tapes.json
```diff
@@ -207,7 +207,7 @@
       "statusText": "OK"
     }
   },
-  "ca042a4ac6a4f262adf83890752b75c22bab477f8ed65e469a87dbe8dcb4941a": {
+  "741f4bc66664a45c59af4ca4251b0604c297cd8927ea66d7fa138d5f6799c61e": {
     "url": "https://api-inference.huggingface.co/models/superb/hubert-large-superb-er",
     "init": {
       "headers": {
```
Aschen, line 213: I didn't thought about this issue, the `Authorization` header is saved in the tapes, we should fix the VCR, update the tapes and possibly revoke the exposed tokens (cc @coyotte508 )

vvmnnnkv:
That was some weirdness when the `process.env.VCR_MODE` is replaced with `undefined`.
It was entering in `if (process.env.VCR_MODE) { ...`  branch. 🤔

packages/inference/test/tapes.json
vvmnnnkv: Oops I thought it's a special access token for testing :D
I can fix in separate PR

packages/inference/test/tapes.json
```diff
@@ -179,7 +179,7 @@
       "statusText": "OK"
     }
   },
-  "ea3752fcae68367386fad4918964a5de0c4eb2a46aa4854ef2ab41de13429c6b": {
+  "45d0b01267b465f464cb9e208b877ee1c7d9986fa4594e03b8b918dc04568d2c": {
```
coyotte508, line 182: any ideas why all the hashes changed?

packages/inference/test/tapes.json
Aschen: Actually it's my token, I revoked it

packages/inference/test/tapes.json
vvmnnnkv: I think it's because fetch inputs has changed to blob in these tests.

packages/inference/vitest.config.ts
```diff
@@ -0,0 +1,19 @@
...
+for (const filename of testFilesToPreload) {
+	testFilesContents[filename] = readFileSync(join(__dirname, "test", filename)).toString("base64");
+}
+
+export default defineConfig({
+	define: {
+		__TEST_FILES__: JSON.stringify(testFilesContents),
+		// make sure browser receives env vars
+		"process.env.HF_ACCESS_TOKEN": JSON.stringify(process.env.HF_ACCESS_TOKEN || ""),
+		"process.env.VCR_MODE": JSON.stringify(process.env.VCR_MODE || ""),
```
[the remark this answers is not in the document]
vvmnnnkv: Yes, `process.env` seems to work better (keeping `process.env.VCR_MODE` undefined had weird effect)

packages/inference/test/tapes.json
vvmnnnkv: Fixed in #133

packages/inference/src/HfInference.ts
```diff
@@ -946,6 +960,18 @@ export class HfInference {
 		return res;
 	}
 
+	/**
+	 * This task reads some image input and outputs the text caption.
+	 */
+	public async imageToText(args: ImageToTextArgs, options?: Options): Promise<ImageToTextReturn> {
+		return (
+			await this.request<ImageToTextReturn[]>(args, {
```
coyotte508, line 968:
```suggestion
			await this.request<[ImageToTextReturn]>(args, {
```

coyotte508:
Thanks a lot!

coyotte508:
v1.8.0 released, demo [here](https://huggingface.co/spaces/huggingface-projects/image-to-text) (and tweeted)

PR #132, opened 2023-04-11
fix text-image parameters

radames:
Since the community api now [supports kwargs,](https://github.com/huggingface/api-inference-community/commit/414c5a08c4f28d49ac6594650877b87b44b8afa4), we can pass more diffusers parameters via the api, I choose a couple of them that [make sense](https://github.com/huggingface/diffusers/blob/fbc9a736dd5d8c20144ea44ef5266c87303ecacd/src/diffusers/pipelines/stable_diffusion/pipeline_stable_diffusion.py#L525-L542)

For the test, we could check the output image dimensions, but it would require an import such as `canvas`, to convert the blob image to an image and get its dimensions. wdyt @coyotte508 

The test with negative_prompt will fail until this [PR is merged](https://github.com/huggingface/api-inference-community/pull/228), as it is now it requires `negative_prompt: ["...negative prompt"]` to be a list not a single value `negative_prompt: "...negative prompt"`


radames:
@coyotte508 updated!

coyotte508 (approved):
Thanks!

PR #133, opened 2023-04-11
💚 Fix VCR module

coyotte508:
cc @Aschen @vvmnnnkv @radames

- Store text data when possible instead of base 64
- Do not store authorization header
- Store response headers, especially `content-type` needed for textGenerationStream

@Aschen now `base64fromBytes` is in both `inference` and `hub` ;)


Aschen:
Not anymore https://github.com/huggingface/huggingface.js/commit/9c606b2738f95328732f08c60ffea203645cf8fc :stuck_out_tongue_winking_eye:

PR #135, opened 2023-04-11
Add codeql github action

Aschen:
## Description

Add a Github Action for CodeQL code analysing

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,39 @@
+name: "CodeQL"
+
+on:
+  pull_request:
+    branches: ["main"]
+  schedule:
+    - cron: "14 6 * * 5"
+
+jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql.yml
```diff
@@ -0,0 +1,39 @@
+name: "CodeQL"
+
+on:
+  pull_request:
+    branches: ["main"]
```
coyotte508, line 5: Seems it's not happy and wants this too:

```
  push:
    branches: ["main"]
```

Aschen: Give a finger and they take the arm :roll_eyes:

coyotte508:
It's possible to enable it directly in the repo settings too, without a GH action:

image

image

So this PR was useful to bring this to attention, but it should be directly handled by github

PR #140, opened 2023-04-13
fix streaming test

radames:
Not sure if this is due to the non-deterministic of text generation, however the space before the `" one"` keep making the test fail.  So I've added the index to add the space WDYT @vvmnnnkv 

```
Array(5) [
  0: Object {
  token: Object {id: 80, text: "one", logprob: -0.75390625, special: false}
  generated_text: null
  details: null
}
  1: Object {
  token: Object {id: 192, text: " two", logprob: -0.01940918, special: false}
  generated_text: null
  details: null
}
  2: Object {
  token: Object {id: 386, text: " three", logprob: -0.015197754, special: false}
  generated_text: null
  details: null
}
  3: Object {
  token: Object {id: 662, text: " four", logprob: -0.01940918, special: false}
  generated_text: null
  details: null
}
  4: Object {
  token: Object {id: 1, text: "</s>", logprob: -0.030883789, special: true}
  generated_text: "one two three four"
  details: null
}
]
```

### DIFF
--- a/packages/inference/src/HfInference.ts
+++ b/packages/inference/src/HfInference.ts
@@ -1048,6 +1048,12 @@ export class HfInference {
 
 		if (options?.blob) {
 			if (!response.ok) {
+				if (response.headers.get("Content-Type")?.startsWith("application/json")) {
+					const output = await response.json();
+					if (output.error) {
+						throw new Error(output.error);
+					}
+				}
 				throw new Error("An error occurred while fetching the blob");
 			}
 			return (await response.blob()) as T;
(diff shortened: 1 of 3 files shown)


coyotte508 (commented):
It seems the generated response changed @radames 

we can ignore spaces imo in the token, eg `text: expect.stringContaining(tokenText)` to be more robust

vvmnnnkv:
It might be due to update of text-generation-inference to v0.5.0, noticed it has changes related to token decoding in https://github.com/huggingface/text-generation-inference/pull/144.

vvmnnnkv (approved):

Issue #22, opened 2023-02-28
Add Document Question Answer


radames:
[Document Question Answering](https://huggingface.co/models?pipeline_tag=document-question-answering)

```js


const base64Image = await fetch("https://s3.amazonaws.com/moonup/production/uploads/1677559577886-6064e095abd8d3692e3e2ed6.png")
    .then(res => res.blob())
    .then(blob => {
      return new Promise((resolve, reject) => {
        const reader = new FileReader();
        reader.onloadend = ()=> resolve(reader.result);
        reader.onerror = reject;
        reader.readAsDataURL(blob);
      });
    });


 const response = await fetch("https://api-inference.huggingface.co/models/impira/layoutlm-document-qa", {
    "headers": {
      "content-type": "application/json",
    },
    body: JSON.stringify({
      inputs: {
        question: "What is the invoice number?",
        image: base64Image.replace(/^data:.+;base64,/, '')
      }
    }),
    method: "POST"
  }).then(res => res.json());

```

coyotte508:
Interesting, I didn't know this way to do base64 conversion!

radames:
yes @coyotte508, I like it because it's simpler than the `fromCharCode` method. However, I wish it was promise-based by default

vvmnnnkv:
Done in #143

PR #143, opened 2023-04-14
Visual and Document Q&A

### LINKED ISSUE #21: Add Visual Question Answering task
Visual question answering tasks, models
https://huggingface.co/models?pipeline_tag=visual-question-answering

```js
fetch(
  "https://api-inference.huggingface.co/models/dandelin/vilt-b32-finetuned-vqa",
  {
    headers: {
    },
    body: JSON.stringfy({
      inputs: { question: "What is on top of the building?", image: base64image },
    }),
    method: "POST",
  }
);
```

vvmnnnkv:
This should close #21 and #22 

Couple questions:
1. Visual Q&A returns multiple results, currently just the first one is returned by hf.visualQuestionAnswer - is that fine or better return all of them?
2. Is there a way for HF API to accept multi-part body instead of submitting base64 of the file in JSON?


coyotte508:
> Visual Q&A returns multiple results, currently just the first one is returned by hf.visualQuestionAnswer - is that fine or better return all of them?

It's better to return all of them I think (cc @radames)

> Is there a way for HF API to accept multi-part body instead of submitting base64 of the file in JSON?

I don't think so, @Narsil do you know ? :)

coyotte508:
#145 changed the name of the functions @vvmnnnkv to be more consistent with HF, from `xxxQuestionAnswer` to `xxxQuestionAnswering`

coyotte508:
> It's better to return all of them I think (cc @radames)

Hmm maybe not :thinking: 

It may be better for the default API to be simple and only return one answer indeed ^^ 

Feel free to chime in @Pierrci @Wauplin too

packages/inference/src/HfInference.ts
```diff
@@ -720,6 +730,20 @@ export class HfInference {
 		return res;
 	}
 
+	/**
+	 * Answers a question on a document image. Recommended model: impira/layoutlm-document-qa.
+	 */
+	public async documentQuestionAnswer(args: ImageQuestionAnswerArgs, options?: Options): Promise<QuestionAnswerReturn> {
+		return (await this.request<[QuestionAnswerReturn]>(args, options))?.[0];
```
radames, line 737: hi @vvmnnnkv could we also add a validation here for the single response value here and for visualQuestionAnswer? thanks

```js
const res = await this.request<[QuestionAnswerReturn]>(args, options))?.[0];
const isValidOutput = Array.isArray(res) && res.every((x) => typeof x.answer === "string" && typeof x.score === "number")
if (!isValidOutput) {
	throw new TypeError("Invalid inference output: output must be of type Array<answer: string, score: number>");
}
```

Narsil:
>   Is there a way for HF API to accept multi-part body instead of submitting base64 of the file in JSON?

Not currently. What's the use-case ?

vvmnnnkv:
> Not currently. What's the use-case ?

Just to avoid base64 encoding and reduce cpu/mem/bandwidth consumption

Narsil:
> Just to avoid base64 encoding and reduce cpu/mem/bandwidth consumption

I sympathize with the goal but the issue is that it breaks the correspondance of 1-1 with the python implementation when sending requests (the API basically calls `pipe = pipeline(...); pipe(inputs, **parameters)` whereas with multipart it becomes harder to write the request correctly and obviously.

Happy to hear suggestions on that front where it's "obvious" to write a request knowing the API expects something like `{"inputs": image, "parameters": {...}}` .

This is the reason we went with base64. I must not that the API does accept binary file when you send a single file with no parameters/other input.

packages/inference/test/HfInference.spec.ts
```diff
@@ -99,6 +99,39 @@ describe.concurrent(
 			});
 		});
 
+		it("documentQuestionAnswer", async () => {
```
coyotte508, line 102:
```suggestion
		it("documentQuestionAnswering", async () => {
```

packages/inference/test/HfInference.spec.ts
```diff
@@ -99,6 +99,39 @@ describe.concurrent(
...
+			).toMatchObject({
+				answer: "us-001",
+				score: expect.any(Number),
+				// not sure what start/end refers to in this case
+				start: expect.any(Number),
+				end: expect.any(Number),
+			});
+		});
+
+		it("visualQuestionAnswer", async () => {
```
coyotte508, line 120:
```suggestion
		it("visualQuestionAnswering", async () => {
```

packages/inference/src/tasks/multimodal/documentQuestionAnswering.ts
```diff
@@ -0,0 +1,49 @@
...
+	options?: Options
+): Promise<DocumentQuestionAnsweringOutput> {
+	const res = (await request<[DocumentQuestionAnsweringOutput]>(args, options))?.[0];
+	const isValidOutput =
+		typeof res?.answer === "string" &&
+		typeof res.end === "number" &&
+		typeof res.score === "number" &&
+		typeof res.start === "number";
+	if (!isValidOutput) {
+		throw new InferenceOutputError("Expected {answer: string, end: number, score: number, start: number}");
```
coyotte508, line 46:
```suggestion
		throw new InferenceOutputError("Expected Array<{answer: string, end: number, score: number, start: number}>");
```

packages/inference/src/tasks/multimodal/visualQuestionAnswering.ts
```diff
@@ -0,0 +1,37 @@
...
+ * Answers a question on an image. Recommended model: dandelin/vilt-b32-finetuned-vqa.
+ */
+export async function visualQuestionAnswering(
+	args: VisualQuestionAnsweringArgs,
+	options?: Options
+): Promise<VisualQuestionAnsweringOutput> {
+	const res = (await request<[VisualQuestionAnsweringOutput]>(args, options))?.[0];
+	const isValidOutput = typeof res?.answer === "string" && typeof res.score === "number";
+	if (!isValidOutput) {
+		throw new InferenceOutputError("Expected {answer: string, score: number}");
```
coyotte508, line 34:
```suggestion
		throw new InferenceOutputError("Expected Array<{answer: string, score: number}>");
```

coyotte508:
I wonder why the test fails despite the tape :thinking: (but not your problem!)

packages/inference/src/tasks/multimodal/documentQuestionAnswering.ts
```diff
@@ -0,0 +1,49 @@
+import { InferenceOutputError } from "../../lib/InferenceOutputError";
+import type { BaseArgs, Options } from "../../types";
+import { request } from "../custom/request";
+
+export type DocumentQuestionAnsweringArgs = BaseArgs & {
+	inputs: {
+		/** Base64 of document image **/
+		image: string;
```
coyotte508, line 8:
```suggestion
		/** 
		 * Raw image
		 * 
		 * You can use native `File` in browsers, or `new Blob([buffer])` in node, or for a base64 image `new Blob([btoa(base64String)])`, or even `await (await fetch('...)).blob()`
		 **/
		image: Blob;
```

I would rather use native `Blob` as the datatype for the image, even if we later convert it to base64 to pass to the inference API.

coyotte508:
> Happy to hear suggestions on that front where it's "obvious" to write a request knowing the API expects something like `{"inputs": image, "parameters": {...}}` .
> 
> This is the reason we went with base64. I must note that the API does accept binary file when you send a single file with no parameters/other input.

@Narsil Just a suggestion, other params (here `question` and `parameters`) could be passed via headers.

packages/inference/src/tasks/multimodal/documentQuestionAnswering.ts
vvmnnnkv: So we'll do blob to base64 conversion for user?
Perhaps we should still allow base64 string, to avoid base64 decoding in case they already have base64 for some reason

coyotte508: > Perhaps we should still allow base64 string, to avoid base64 decoding in case they already have base64 for some reason

Not a fan, it's an optimization for a specific use case. We can revisit later but for now I'd rather just the `Blob` like our other tasks, to keep the API straightforward

coyotte508: by the way there's already a `base64FromBytes` function that can be moved to `packages/shared`

coyotte508:
It seems you prettier'd the tapes.json file with tabs. It's fine, but you should update `vcr.ts` then to use tabs as well whens saving ^^

Narsil:
> @Narsil Just a suggestion, other params (here question and parameters) could be passed via headers.

That requires special custom headers for us, and passing them along to the endpoint from the proxy.
Yes it's doable, I've considered doing that to avoid base64. But headers have their own limitations/caveats which makes them not ideal (and it doesn't really work with multiple inputs types like image comparison or image question answering)

coyotte508 (approved):
Very cool!

PR #148, opened 2023-04-14
♻️ Export individual functions, precise typing

coyotte508:
cc @Grsmto @vvmnnnkv @TimMikeladze 

Inference endpoints do not accept `model` anymore.

Also, individual functions for the various tasks are exported, taking an additional optional `accessToken` parameter.

This is a big refacto even if the API **keeps backwards compatibility**, so I'd like some input :)

---

This moves the Inference API closer to `@huggingface/hub`, and conversely, we can also move the `@huggingface/hub` api closer to the inference module by offering a `new HfHub("hf_...")` class


vvmnnnkv (commented):
Really nice to have each task function in its own file!
Some suggestions:
* Put task functions files to a folder like `tasks`, so they are not mixed with internal methods
* Perhaps even split them to sub-folders like `nlp`, `cv`, `multimodal`, etc. following HF models classification
* Move args/response types to corresponding task function file

coyotte508:
>   * Put task functions files to a folder like `tasks`, so they are not mixed with internal methods

:+1: 

>  * Perhaps even split them to sub-folders like `nlp`, `cv`, `multimodal`, etc. following HF models classification

And we could move `request` and `streamingRequest` to a `custom` folder too !

> * Move args/response types to corresponding task function file

The thing is that some types are reused across tasks. Although we could duplicate the types I guess :) More future-proof

Good suggestions ^^

coyotte508:
Merging to unblock parallel PRs, waiting a bit before release, feel free to comment in the meantime

Issue #165, opened 2023-04-25
Add E2E where the module is downloaded (or linked) to a TS project


coyotte508:
To prevent things like #164

Issue #164, opened 2023-04-25
Cannot find module base64FromBytes when using TypeScript


shadow1349:
Node Version: 18.16.0
TypeScript Version: 5.0.4

Here's a screenshot of the issue:

![CleanShot 2023-04-25 at 12 47 29](https://user-images.githubusercontent.com/9685464/234359831-01c5aac0-2766-4d8a-95d4-3e6b44e083b0.png)

I created a brand new TypeScript project, installed @huggingface/inference, and then tried something really simple:

```typescript
import { HfInference } from "@huggingface/inference";

const hfToken = process.env.token;

const hf = new HfInference(hfToken);

(async () => {
  const sum = await hf.summarization({
    model: "facebook/bart-large-cnn",
    inputs:
      "The tower is 324 metres (1,063 ft) tall, about the same height as an 81-storey building, and the tallest structure in Paris. Its base is square, measuring 125 metres (410 ft) on each side. During its construction, the Eiffel Tower surpassed the Washington Monument to become the tallest man-made structure in the world, a title it held for 41 years until the Chrysler Building in New York City was finished in 1930.",
    parameters: {
      max_length: 100,
    },
  });

  console.log(sum);
})();
```

Here is a repository that recreates the issue
- https://github.com/shadow1349/hfmoduleissues

coyotte508:
Thanks for the report! 

It's fixed in https://github.com/huggingface/huggingface.js/commit/135fd72c4abf33b2f531d03094c40063149e7a83, and in @hugginface/inference v2.1.3 (NPM registry may take some time to update)

shadow1349:
Yeah just tested 2.1.3 and it's been resolved. Thanks for the speedy reply!

PR #166, opened 2023-04-29
♻️ Extract shared code

vvmnnnkv:
New attempt to extract common files into shared package, plus e2e test for inference package (#165) to check it actually works.

Background:
 * Inference package includes .ts source code and references it in `types` of package.json
 * This is because compiled *.d.ts file loses proper doc strings
 * However, when such package is installed in a node TS project, .ts code imports outside of inference package scope result in error (#164)

Approach:
 * Put common code in private "shared" package in the workspace
 * Reference it as dev dependency in other packages using workspace protocol
 * Have it listed in bundledDependencies in package.json so the shared .ts source code is included

Note: in GH action had to update pnpm to v8 because v7 doesn't honor bundledDependencies when installing from a local folder. Can change this to pnpm pack + pnpm install from tgz.


coyotte508:
Thanks a lot @vvmnnnkv , it's really neat and lets us use our own internal packages but also external packages like type-fest.

Can you upgrade all pnpm lock files to pnpm 8 too, as well as in the CI file, and change `@huggingface/hub`'s source code to use the shared package?

Issue #157, opened 2023-04-18
Image-to-Image 


radames:
It depends on this PR https://github.com/huggingface/api-inference-community/pull/223

```
        Args:
            inputs (:obj:`str`):
                a string containing some text
            image (:obj:`PIL.Image.Image`):
                a condition image
        Return:
            A :obj:`PIL.Image.Image` with the raw image representation as PIL.
```
```
{"inputs": {
    image: base64image,
    text: "",
}}
```

coyotte508:
It is like the pose/controlnet models, an image as a pose, a text of what we want generated, and the output is the input image adapted to fit the text?

coyotte508:
https://huggingface.co/tasks/image-to-image

It seems some of those models do not have text input, only image to image :thinking: 

Can you clarify what it does? Should we use the same `imageToImage` task for both, or call it `imageToImageGuided` instead?

osanseviero:
Internal Slack discussion - https://huggingface.slack.com/archives/C016D661PAN/p1680600898941279?thread_ts=1680599959.804889&cid=C016D661PAN

patrickvonplaten:
In short we have the following type of models in `diffusers`:

[...]

PR #175, opened 2023-05-02
Img2img task

radames:
implement #157


packages/inference/test/tapes.json
```diff
@@ -156,13 +156,15 @@
...
-      "statusText": "OK",
+      "body": "{\"error\":\"overloaded\"}",
+      "status": 503,
+      "statusText": "Service Unavailable",
       "headers": {
         "access-control-allow-credentials": "true",
         "connection": "keep-alive",
         "content-type": "application/json",
+        "server": "uvicorn",
+        "transfer-encoding": "chunked",
```
coyotte508, line 167: Let's not add 5XX to tapes.

You can do `it.only` to only record the new test

packages/inference/src/tasks/cv/imageToImage.ts
```diff
@@ -0,0 +1,83 @@
...
+import type { BaseArgs, Options, RequestArgs } from "../../types";
+import { request } from "../custom/request";
+import { base64FromBytes } from "@huggingface/shared";
+
+export type ImageToImageArgs = BaseArgs & {
+	/**
+	 * The initial image condition
+	 *
+	 **/
+	image: Blob | ArrayBuffer;
```
coyotte508, line 11: Nice. Can you make document Q&A and visual Q&A support `ArrayBuffer` as well?

packages/inference/src/types.ts
```diff
@@ -40,7 +40,7 @@ export interface BaseArgs {
 }
 
 export type RequestArgs = BaseArgs &
-	({ data: Blob | ArrayBuffer } | { inputs: unknown }) & {
+	({ data: Blob | ArrayBuffer } | { inputs: unknown } | { image: string }) & {
```
coyotte508, line 43: Arg. See https://huggingface.slack.com/archives/C016D661PAN/p1683012959767189 :(

coyotte508: Apparently we should still accept `inputs: string` cc @Narsil, so let's keep using `inputs` for consistency

packages/inference/test/tapes.json
radames: this is `questionAnswering` test failing since this model https://huggingface.co/deepset/roberta-base-squad2 is overloaded, I can checkout the old tapes.json and add the it.only

radames: I've rolled back tapes.json, for now since we need the api-inference to change the params, to make tests here pass again

radames:
new update depends on https://github.com/huggingface/api-inference-community/pull/244

PR #176, opened 2023-05-02
Utility to parse safetensors remotely

julien-c:


packages/hub/src/lib/parse-safetensors.ts
```diff
@@ -0,0 +1,136 @@
...
+			...(await parseShardedIndex(indexUrl)),
+		};
+	} else {
+		throw new Error("model id does not seem to contain safetensors weights");
+	}
+}
+
+function computeNumOfParamsByDtypeSingleFile(header: FileHeader): Counter<Dtype> {
+	const n = new Counter<Dtype>();
+	for (const [k, v] of Object.entries(header)) {
```
coyotte508, line 113: We can add `typedEntries` and maybe `omit` to the utils, to avoid manual castings below

packages/hub/src/lib/parse-safetensors.ts
```diff
@@ -0,0 +1,136 @@
...
+		if (k === "__metadata__") {
+			continue;
+		}
+		if ((v as TensorInfo).shape.length === 0) {
+			continue;
+		}
+		n.incr(
+			(v as TensorInfo).dtype,
+			(v as TensorInfo).shape.reduce((a, b) => a * b)
+		);
```
coyotte508, line 123:
```suggestion
		n.set(v.dtype, (n.get(v.dtype) ?? 0) + sum(v.shape.map((a, b) => a * b));
```

We can add a `sum` utils. I don't think we need to add a whole `Counter` class

packages/hub/src/lib/parse-safetensors.ts
```diff
@@ -0,0 +1,142 @@
...
+		redirect: "manual",
+		/// ^do not follow redirects to save some time
+	});
+	return res.status >= 200 && res.status < 400;
+}
+
+export async function parseSafetensorsFromModelRepo(params: {
+	modelId: string;
+	hubUrl?: string;
+}): Promise<ParseFromRepo> {
```
coyotte508, line 91:
```suggestion
export async function parseSafetensorsMetadata(params: {
  repo: RepoDesignation; // can just be a model id
	hubUrl?: string;
	credentials?: Credentials;
	revision?: string;
}): Promise<ParseFromRepo> {
```

For consistency with the rest of the exports. We can add checks to make sure that `repo` is a model:

```ts
const repoId = toRepoId(params.repo);

if (repoId.type !== "model) {
  throw new TypeError("...");
}
```

coyotte508 (commented):
The function also needs to be xported/imported from `lib/index.ts`

packages/hub/src/lib/parse-safetensors.ts
julien-c: 0ad1d27209d4f02dfc4d83d52e64a7765ca42b1e

packages/hub/src/lib/parse-safetensors.ts
julien-c: maybe like 6d1124d517d2a1e0b3ac3b58c6a3e192bc8c835c?

packages/hub/src/lib/parse-safetensors.ts
julien-c: Are you sure? Given i'm also using .dict and .total, I'm wondering if it's going to be a bit verbose/error-prone to not use it

julien-c: i've slimmed it down a litte bit in 2cc913d7097ce50fda339fba1f6ad74a1587ebb1

julien-c: i can still remove it if you think it's better, though

julien-c:
btw, is there a place in the repo where i can place a (executable) sample usage script? (which would produce the output of https://github.com/huggingface/safetensors/issues/44#issuecomment-1531030534)

packages/hub/src/lib/parse-safetensors.ts
```diff
@@ -0,0 +1,161 @@
...
+	__metadata__: Record<string, string>;
+};
+
+interface IndexJson {
+	dtype?: string;
+	/// ^there's sometimes a dtype but it looks inconsistent.
+	metadata?: Record<string, string>;
+	/// ^ why the naming inconsistency?
+	weight_map: Record<TensorName, FileName>;
+}
```
coyotte508, line 34: Let's use `camelCase` for what we expose to the users (here `weight_map` and `data_offsets`), for raw API types there is the `types/api` folder

Also let's prefix types (incl. `IndexJson` and `ShardedHeaders`) with `Tensor`

packages/hub/src/lib/parse-safetensors.ts
```diff
@@ -0,0 +1,161 @@
...
+	return n;
+}
+
+export function computeNumOfParamsByDtype(parse: ParseFromRepo): Counter<Dtype> {
+	if (parse.sharded) {
+		return computeNumOfParamsByDtypeSharded(parse.headers);
+	} else {
+		return computeNumOfParamsByDtypeSingleFile(parse.header);
+	}
+}
```
coyotte508, line 161: Imo it's too generic a function name to be exported (when it's not part of a safetensors namespace/module)

Maybe if we keep the function we can just compute it always and include it in the result of `ParseFromRepo`?

Also, exporting a `Counter` class to the users of the lib is not great, we should export the results of `Counter.dict()` or even better, export a `Record<DType, number>` eg:

```
{
  BOOL: 0,
  I8: 0,
  I16: 21892,
  I32: 0,
  ...
}
```

Not sure we need to add a `total` field, a user could just `sum(Object.values(X))`

packages/hub/src/utils/pick.ts
```diff
@@ -0,0 +1,25 @@
...
+ * Return copy of object, omitting blacklisted array of props
+ */
+export function omit<T extends Record<string, unknown>, K extends keyof T>(
+	o: T,
+	props: K[] | K
+): Pick<T, Exclude<keyof T, K>> {
+	const propsArr = Array.isArray(props) ? props : [props];
+	const letsKeep = (Object.keys(o) as (keyof T)[]).filter((prop) => !propsArr.includes(prop as unknown as K));
+	return pick(o, letsKeep);
+}
```
coyotte508, line 25: Two separate files please ^^

packages/hub/src/lib/parse-safetensors.ts
julien-c: > Also let's prefix types (incl. `IndexJson` and `ShardedHeaders`) with `Tensor`

i'll prefix with `Safetensors` then? (`Tensor` is less relevant at the top-level, it's more like a collection of tensors)

or should i nest everything into a namespace?

packages/hub/src/lib/parse-safetensors.ts
julien-c: > Maybe if we keep the function we can just compute it always and include it in the result of `ParseFromRepo`?

I'd rather not, as i prefer the output of parse be the exact data from the files

packages/hub/src/lib/parse-safetensors.ts
coyotte508: `Safetensors` yes :+1:

packages/hub/src/lib/parse-safetensors.ts
julien-c: Ok i will remove `Counter` completely

packages/hub/src/lib/parse-safetensors.ts
julien-c: (it's not a sum BTW, it's really the dot product of everything a * b  * c * ...)

coyotte508: (yes noticed afterwards ^^)

julien-c:
Ok. It's not really a sample script, more like a small utility script. I'll store it elsewhere.

packages/hub/src/utils/pick.ts
julien-c: 365bbc17bc944003f2aec6e496db402d14be906f

packages/hub/src/lib/parse-safetensors.ts
julien-c: 3757fc3598049bf3eba451e5e363a958f00f25b3

packages/hub/src/lib/parse-safetensors.ts
julien-c: 3757fc3598049bf3eba451e5e363a958f00f25b3

packages/hub/src/lib/parse-safetensors.ts
julien-c: feels a bit verbose tbh: ccf2dbea19c7f8726982455819ab89fc5cf0f8c6

julien-c: re. snake_case, i'd rather not transform the data that's actually inside the safetensors files/format, if that's ok

julien-c (commented):
CI error is unrelated:

> test/HfInference.spec.ts > HfInference > questionAnswering

julien-c:
(relauching CI)

julien-c:
(nope, still down)

coyotte508:
> For the documentation, do I need to so something to have auto-generated page? (i would like to link to it from the main safetensors doc)

When publishing a new version it will update the docs on hf.co/docs

You can also edit `packages/hub/README.md` and it will trigger the dev documentation in this PR

coyotte508:
WDYT @julien-c of the changes?

coyotte508 (approved):

julien-c:
Is it easy to get this into a new released version?

coyotte508:
@julien-c go in https://github.com/huggingface/huggingface.js/actions/workflows/hub-publish.yml, click "run wokflow", replace "patch" by "minor" and launch!

julien-c:
Failure 🥲
https://github.com/huggingface/huggingface.js/actions/runs/4946961906/jobs/8845705043

PR #177, opened 2023-05-02
accept image as ArrayBuffer

radames:
do we also want to support an base64 string direct?

```js
image: Blob | ArrayBuffer | string;
```


coyotte508 (approved):
Thanks! And not atm for the base64 string I think, not unless there is real demand

Issue #179, opened 2023-05-03
[Yarn] Install Error: Couldn't find package "@huggingface/shared@1.0.0"


carrotfarmer:
Error while trying to install `@huggingface/inference` using yarn:

<img width="954" alt="image" src="https://user-images.githubusercontent.com/66675022/235894508-ae1190f6-7b17-4088-a245-bfc2e094bbf1.png">

```
yarn add v1.22.17
[1/4] 🔍  Resolving packages...
error Couldn't find package "@huggingface/shared@1.0.0" required by "@huggingface/inference" on the "npm" registry.
info Visit https://yarnpkg.com/en/docs/cli/add for documentation about this command.
```

**OS:** macOS Monterey
**yarn version**: 1.22.17
**npm version**: 8.6.0
**node version**: 18.14.2

carrotfarmer:
It works on npm. Dunno why it isn't working in yarn

coyotte508:
Yarn is really opinionated, see https://yarnpkg.com/getting-started/migration#dont-use-bundledependencies :sweat_smile: 

cc @vvmnnnkv 

They recommend using `file:` instead of bundled dependencies but I'm not sure it's possible with TS types

coyotte508:
Also  the error is because yarn tries to install the dev dependencies of `@huggingface/inference`, you should be able to fix it by telling it not to: https://stackoverflow.com/questions/49530678/why-does-yarn-install-dev-dependencies-but-i-just-need-the-builds

Btw if you're looking for a cool `yarn` replacement there's `pnpm` ;)

carrotfarmer:
Ah alright
Thanks for the quick response!
(btw: actually switched to pnpm after this incident lmao)

PR #188, opened 2023-05-11
🙈 Mask the use of `@huggingface/shared` because of yarn

coyotte508:
cc @vvmnnnkv  ;)

And @carrotfarmer see https://github.com/huggingface/huggingface.js/issues/179

Should fix yarn install, and also improve Deno importing


coyotte508:
Didn't work :(

At least deno instructions are correct now

PR #190, opened 2023-05-12
👷 Remove shared dependency (again!)

vvmnnnkv:
It turned out that shared workspace dependency must be public in order to work with yarn.
Also yarn & deno still try to resolve bundledDependencies.
So we're back to imports from relative path.
However, since package.json's `types` of hub & inference now point to compiled d.ts, this doesn't lead to missing imports problem when using npm packages.
Note that types imported from `type-fest` are resolved into d.ts.

### DIFF
--- a/.github/workflows/lint-and-test.yml
+++ b/.github/workflows/lint-and-test.yml
@@ -57,10 +57,10 @@ jobs:
           HF_ACCESS_TOKEN: ${{ secrets.HF_ACCESS_TOKEN }}
 
       - name: E2E - start mock npm registry
-        working-directory: e2e
         run: |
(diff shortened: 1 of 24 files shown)


coyotte508 (approved):
Amazing, thank you :)

Can you revert my recent change to `packages/inference/README.md`:

> `yarn` doesn't work because it tries to get dev dependencies, one of which is private.

Also, can you edit the deno sections of the main README.md, and `packages/inference/README.md`? to add the `npm:@huggingface/inference` example.

And add a deno section inside `packages/hub/README.md`?

coyotte508 (approved):
:tada: :heart: :boom:

PR #196, opened 2023-05-19
🐛 Adjust feature extraction expected result check

### LINKED ISSUE #194: Not every feature extraction model returns an array of numbers
https://github.com/huggingface/huggingface.js/blob/1e7cda0f14a54b8fac345cb88ded37d8f353d87d/packages/inference/src/tasks/nlp/featureExtraction.ts#L36

Some models like `nielsr/lilt-xlm-roberta-base` doesn't return an array of numbers, but an array of arrays of numbers.
So the output should be valid not only when every response item is a number, but also when every response item is a number array.

vvmnnnkv:
Feature extraction model can have 3-d output.
Fixes #194


coyotte508 (approved):
Thanks!

lisandropuzzolo:
Thank you [vvmnnnkv](https://github.com/vvmnnnkv)!
I've written a PHP SDK for huggingface similar than this one. 

I would like to add to this project (huggingface.js) some of my features not present, like for example ejecute different tasks for one model (right now huggingface.js only allows to ejecute the default task).

Is there any guide to download the project, execute tests, and propose changes?
Thanks in advance!

PR #197, opened 2023-05-19
feature extraction, sentence similarity and text classification now use pipeline url

lisandropuzzolo:
Some models allows them to be use on more than one (default task).
In order to let the huggingface.js use more than the default model task, I've changed featureExtraction, sentenceSimilarity and textClassification to change the model to the pipeline url when model is not a url.


packages/inference/src/lib/makeRequestOptions.ts
```diff
@@ -38,7 +48,7 @@ export function makeRequestOptions(
 		}
 	}
 
-	const url = /^http(s?):/.test(model) || model.startsWith("/") ? model : `${HF_INFERENCE_API_BASE_URL}${model}`;
+	const url = isModelTypeURL(model) ? model : `${HF_INFERENCE_API_MODEL_BASE_URL}${model}`;
```
coyotte508, line 51: The `/^http(s?):/.test(model)` test needs to stay, it's for inference endpoints.

Only the latter part of the ternary expression needs to be changed

coyotte508:
Thanks, that's a neat feature!

I would add a `pipeline` optional string param here: 
image

And do the logic of changing the url in `makeRequestOptions` instead.

In `textClassification` etc, it would be a one-line-change:

```ts
const res = (await request<TextClassificationOutput[]>(args, {
  ...options, 
  pipeline: "text-classification" // HERE!
}))?.[0];
```

coyotte508 (commented):
Thanks, that's a neat feature!

I would add a `pipeline` optional string param here: 
image

And do the logic of changing the url in `makeRequestOptions` instead.

In `textClassification` etc, it would be a one-line-change:

```ts
const res = (await request<TextClassificationOutput[]>(args, {
  ...options, 
  pipeline: "text-classification" // HERE!
}))?.[0];
```

packages/inference/src/lib/makeRequestOptions.ts
lisandropuzzolo: I agree with the previous suggestions, but I'm not getting this one...
I've replaced the first part of the ternary expression with a a function call that just do the same:
```
export function isModelTypeURL(model: string): boolean {
	return /^http(s?):/.test(model) || model.startsWith("/");
}
```

I've done this way in order to not write the same expression in getPipelineURL(...)

Would you please clarify me?
Thanks in advance!

(I'm already implementing the other suggestions you made)

lisandropuzzolo:
Done changes sugested by [coyotte508](https://github.com/coyotte508)

packages/inference/src/lib/makeRequestOptions.ts
```diff
@@ -1,6 +1,20 @@
...
+	return /^http(s?):/.test(model) || model.startsWith("/");
+}
+
+function getPipelineURL(model: string, pipeline: string): string {
+	return `${HF_INFERENCE_API_PIPELINE_BASE_URL}${pipeline}/${model}`;
+}
+
+function getDefaultModelTaskURL(model: string) {
+	return `${HF_INFERENCE_API_MODEL_BASE_URL}${model}`;
+}
```
coyotte508, line 17: Those are one-line local functions used once, can you hardcode them directly in `makeRequestOptions`?

packages/inference/src/tasks/nlp/featureExtraction.ts
```diff
@@ -2,6 +2,8 @@ import { InferenceOutputError } from "../../lib/InferenceOutputError";
 import type { BaseArgs, Options } from "../../types";
 import { request } from "../custom/request";
 
+const HF_INFERENCE_API_PIPELINE_FEATURE_EXTRACTION = "feature-extraction";
```
coyotte508, line 5: The constant is only used once / not worth exporting, can you hardcode it below and remove the declaration here?

(maybe later we'll type `pipeline` as `"feature-extraction" | ... | string` to offer some kind of completion)

Same remark for the two other tasks

coyotte508 (commented):
A few comments left, after they're handled I'll update `tapes.json` and merge !

julien-c (commented):
would be interested in @Narsil (and @OlivierDehaene) feedback and maybe @Wauplin too on this

is this a feature of the Inference API we want to expose?

lisandropuzzolo:
> would be interested in @Narsil (and @OlivierDehaene) feedback and maybe @Wauplin too on this
> 
> is this a feature of the Inference API we want to expose?

it's a very needed inference api feature, because many models can be used for more than one task

Wauplin:
Makes me think about this recent convo (https://github.com/huggingface/huggingface_hub/pull/1474#discussion_r1196255344). In `huggingface_hub` we'll use the /task endpoint to allow this.

coyotte508:
Sorry @lisandropuzzolo wrong git manipulation :grimacing:

PR #198, opened 2023-05-22
:sparkles: Audio to audio task

### LINKED ISSUE #11: Audio-to-audio

vvmnnnkv:
Fixes #11 

Not super happy about returning a base64-encoded blob, but this is consistent with `imageSegmentation` where image mask is also returned as base64


coyotte508 (commented):
Can you add a max of `20kB` for tapes? (after which we only save an empty body)

The output of image segmentation is only ~ 16kB, here it's 686kB.

radames (commented):
hi @vvmnnnkv , yes the base64 there is not great but here is the reason
https://github.com/huggingface/api-inference-community/commit/d1c680979093f40c38646e7aa2545eeac1d6c526

> - Decision to return base64 blobs in JSON. It's not super shiny but it
does the work and enabled easy ways to add annotations for audio streams
(like the instrument labels for source separation).

coyotte508:
yes let's return the raw response for now, maybe later offer alternatives

radames:
I see your point, maybe in the future @coyotte508 we could offer some utils for these conversion right?

packages/inference/test/vcr.ts
```diff
@@ -135,8 +135,40 @@ async function vcr(
...
+							// base64 heuristic
+							value.length % 4 === 0 &&
+							value.match(/^[a-zA-Z0-9+/]+={0,2}$/)
+						) {
+							return value.slice(0, 1_000);
+						} else {
+							return value;
+						}
+					})
+				);
```
coyotte508, line 160: :exploding_head:

PR #200, opened 2023-05-22
Allow different tasks for some models

coyotte508:
Contributed by @lisandropuzzolo

See https://github.com/huggingface/huggingface.js/pull/197


radames:
hi @coyotte508 , sentence transformers all Mini is a good example https://huggingface.co/sentence-transformers/all-MiniLM-L6-v2

packages/inference/src/lib/makeRequestOptions.ts
```diff
@@ -38,7 +45,45 @@ export function makeRequestOptions(
...
+				}
+			}
+
+			if (cachedTask.task !== options.task) {
+				return `${HF_INFERENCE_API_BASE_URL}pipeline/${options.task}/${model}`;
+			}
+		}
+
+		return `${HF_INFERENCE_API_BASE_URL}models/${model}`;
+	})();
```
radames, line 85: hi @coyotte508 , trying to understand this, why do you need to cache the task?

radames: Oh I see, just to check the task tag , but if the user is passing the task, do you even need to check the hub ?

coyotte508: Here the task is automatically passed from the `featureExtraction`, `textClassification` or `sentenceSimilarity` function, it's not passed by the user.

Maybe we should add an param to differentiate direct calls to `request` from calls made by the lib wrappers.

PR #203, opened 2023-05-26
Fix non-unique hashlinks problem

mishig25:
### Problem

doc-builder turns markdown headings into html ids by replacing space with `-` (for example: `## Defined in` -> `id="defined-in"`) and this becomes a problem when there are same markdown headings -> multiple elements with same ids. Right navbar navigation breaks https://huggingface.co/docs/huggingface.js/hub/modules#downloadfile

### Solution

Doc-builder provides [syntax](https://github.com/huggingface/doc-builder#anchor-link) to customize element ids. 

#### Example

before:
```
# Interface: Credentials
## Properties
### accessToken
• **accessToken**: `string`
#### Defined in
hub/src/types/public.d.ts:16
```

after:
```
# Interface: Credentials
## Properties
### accessToken
• **accessToken**: `string`
#### Defined in[[accesstoken.defined-in]]
hub/src/types/public.d.ts:16
```

### DIFF
--- a/packages/doc-internal/package.json
+++ b/packages/doc-internal/package.json
@@ -5,18 +5,19 @@
 	"description": "Package to generate doc for other @huggingface packages",
 	"private": true,
 	"scripts": {
-		"start": "pnpm run fix-cdn-versions && pnpm run doc-hub && pnpm run doc-inference && cp ../../README.md ../../docs/index.md && pnpm run update-toc && pnpm run fix-md-links",
+		"start": "pnpm run fix-cdn-versions && pnpm run doc-hub && pnpm run doc-inference && cp ../../README.md ../../docs/index.md && pnpm run update-toc && pnpm run fix-md-links && pnpm run fix-md-headinghashlinks",
 		"lint": "eslint --quiet --fix --ext .cjs,.ts .",
 		"lint:check": "eslint --ext .cjs,.ts .",
 		"format": "prettier --write .",
 		"format:check": "prettier --check .",
-		"prepublish-hub": "pnpm run fix-cdn-versions && pnpm run doc-hub && cp ../../README.md ../../docs/index.md && pnpm run update-toc && pnpm run fix-md-links",
-		"prepublish-inference": "pnpm run fix-cdn-versions && pnpm run doc-inference && cp ../../README.md ../../docs/index.md && pnpm run update-toc && pnpm run fix-md-links",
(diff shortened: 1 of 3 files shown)


packages/doc-internal/package.json
```diff
@@ -5,18 +5,19 @@
...
+		"prepublish-inference": "pnpm run fix-cdn-versions && pnpm run doc-inference && cp ../../README.md ../../docs/index.md && pnpm run update-toc && pnpm run fix-md-links && pnpm run fix-md-headinghashlinks",
 		"doc-hub": "typedoc --tsconfig ../hub/tsconfig.json --githubPages false --plugin typedoc-plugin-markdown --out ../../docs/hub --hideBreadcrumbs --hideInPageTOC --sourceLinkTemplate https://github.com/huggingface/huggingface.js/blob/main/{path}#L{line} ../hub/index.ts",
 		"doc-inference": "typedoc --tsconfig ../inference/tsconfig.json --githubPages false --plugin typedoc-plugin-markdown --out ../../docs/inference --hideBreadcrumbs --hideInPageTOC --sourceLinkTemplate https://github.com/huggingface/huggingface.js/blob/main/{path}#L{line} ../inference/src/index.ts",
 		"update-toc": "ts-node --esm update-toc.ts",
 		"fix-cdn-versions": "ts-node --esm fix-cdn-versions.ts",
-		"fix-md-links": "ts-node --esm fix-md-links.ts"
+		"fix-md-links": "ts-node --esm fix-md-links.ts",
+		"fix-md-headinghashlinks": "ts-node --esm fix-md-headinghashlinks"
```
coyotte508, line 20:
```suggestion
		"fix-md-headinghashlinks": "ts-node --esm fix-md-headinghashlinks.ts"
```

I think that's what makes the CI fail?

coyotte508:
The "Returns" doesn't seem to work, maybe because it's `h5` and not `h4`? 

It's not in the navigation sidebar, but we can still get links when clicking on hover on the header

mishig25:
> The "Returns" doesn't seem to work, maybe because it's h5 and not h4?

yes. I will fix it

> It's not in the navigation sidebar, but we can still get links when clicking on hover on the header

this decision was made in moon-landing to only show up to h4 level

coyotte508:
Sorry, but all the "returns" have the same link: `#type-declaration.returns`

coyotte508 (approved):

coyotte508:
Thanks!

PR #214, opened 2023-06-19
🐛 Update feature extraction expected type

### LINKED ISSUE #212: featureExtraction: Invalid inference output when using an array of inputs
Trying to use the following:

```ts
    const output = await hf.featureExtraction({
      model: 'intfloat/e5-small-v2',
      inputs: ['test', 'hello'],
    });
```

Get:

```
Invalid inference output: Expected Array<number[][] | number[] | number>. Use the 'request' method with the same parameters to do a custom call with no type checking.
```

Non-array string input does not cause this issue.

vvmnnnkv:
Fixes #212


coyotte508 (approved):
Thanks @vvmnnnkv for looking into it!

PR #218, opened 2023-07-01
:recycle: res.warnings is an optional return

DirectriX01:
API calls to models such as `facebook/blenderbot-400M-distill` do not return any warnings, therefore the package throws an InferenceOutputError. With this commit, `res.warnings` can be either undefined or the said definition.


coyotte508 (approved):

coyotte508:
Thanks!

PR #236, opened 2023-07-27
Get rid of default models on some tools

nsarrazin:
The models I hardcoded were outdated for certain tools so I removed them. Now the tools will use the model suggestions from the API instead so we're always up to date.

I left the hardcoded model on `textToSpeech` since the suggested models in the API don't work for now. (See [this pr](https://github.com/huggingface/transformers/pull/24952))


julien-c:
cool, nice! make sure they're pinned on the API side then

PR #241, opened 2023-08-17
change: content-type compare logic `!==` to startswith

dongs0104:
wrapping with fastapi python return with charsets, so i change compare logic :)


julien-c (approved):
i think this is reasonable but wil let others chime in

julien-c:
thanks!

PR #243, opened 2023-08-19
fix: function param typo

0o001:


nsarrazin (approved):
Thanks for catching this 😆

PR #248, opened 2023-09-21
fix: CF Workers doesn't support credentials so don't provide it

WalshyDev:
Workers don't support `credentials` being passed into fetch. If we detect we're inside Workers, we'll not provide this property.


julien-c:
looks good from quick glance, but ping @xenova and possibly @nsarrazin for a quick review . Thanks!

xenova:
Thanks for the PR @WalshyDev 🤗 

Indeed, I think this is something missing from the spec: the `credentials` property can be `omit`, `same-origin` (default), or `include` (see [here](https://developer.mozilla.org/en-US/docs/Web/API/Request/credentials)), however HF.js only supports `include` or `same-origin`.

Just to confirm, you say that `same-origin` (which is the default) and `omit` don't work for cloudflare workers?

WalshyDev:
Yeah, having the property at all in `fetch` is unsupported in Workers today, it's marked as won't implement here:
https://github.com/cloudflare/workerd/blob/ad540fb826261fc3a95a12bf5a2afc3f65bc7ca1/src/workerd/api/http.h#L579-L585

Usage of it results in `Uncaught Error: The 'credentials' field on 'RequestInitializerDict' is not implemented.`.

Other edge runtimes such as Deno are similar:
https://docs.deno.com/runtime/manual/runtime/web_platform_apis#spec-deviations
```
The referrer, referrerPolicy, mode, credentials, cache, integrity, keepalive, and window properties and their relevant behaviours in RequestInit are not implemented. The relevant fields are not present on the Request object.
```

xenova:
Thanks for the extra info 👍 I'm a little hesitant to include custom logic like this which only targets certain runtimes. For example, `navigator` isn't defined in Node.js (see failing tests).

I think a better option would be to expose `includeCredentials` to the end-user and allow them to specify: `undefined` (CW will do), `omit`, `same-origin` (default), or `include`:
https://github.com/huggingface/huggingface.js/blob/2a7b6837e4d5c977927736856cfe7984cec8de55/packages/inference/src/lib/makeRequestOptions.ts#L21-L22
... unless there's some reason for keeping it internal (@mishig25 @nsarrazin)?

packages/inference/src/lib/makeRequestOptions.ts
```diff
@@ -18,8 +18,6 @@ export async function makeRequestOptions(
 		stream?: boolean;
 	},
 	options?: Options & {
-		/** For internal HF use, which is why it's not exposed in {@link Options} */
-		includeCredentials?: boolean;
```
xenova, line 22: Probably best not to remove this, in case other libraries depend on it.

packages/inference/src/lib/makeRequestOptions.ts
```diff
@@ -29,7 +27,7 @@ export async function makeRequestOptions(
 	// eslint-disable-next-line @typescript-eslint/no-unused-vars
 	const { accessToken, model: _model, ...otherArgs } = args;
 	let { model } = args;
-	const { forceTask: task, includeCredentials, taskHint, ...otherOptions } = options ?? {};
+	const { forceTask: task, include_credentials, taskHint, ...otherOptions } = options ?? {};
```
xenova, line 30:
```suggestion
	const { forceTask: task, includeCredentials, taskHint, ...otherOptions } = options ?? {};
```
any reason for renaming?

packages/inference/src/lib/makeRequestOptions.ts
WalshyDev: Ah, other libraries may use it. I couldn't find any usages in this repo so assumed it wasn't used.

packages/inference/src/lib/makeRequestOptions.ts
WalshyDev: Just following the convention of these other ones:
https://github.com/huggingface/huggingface.js/blob/main/packages/inference/src/types.ts#L1-L26

packages/inference/src/lib/makeRequestOptions.ts
xenova: tbh, I'm not sure either 😅 (since it isn't exposed to end-users). I interpreted the comment as "used by other HF libraries". 🤷‍♂️

WalshyDev:
Ok, I have changed to what was discussed. The comment explains it all pretty well but essentially, it keeps backwards compatibility for user usages, if undefined we keep the usual "same-origin", if true "include". Then `false` becomes undefined. If string, we just pass through.

It's a bit messy but not sure how to make it better while keeping compatibility haha

xenova:
Thanks! 👍

By the way, is there a reason you can't use a custom `fetch` function for this (which strips those headers)? The library has an option for it:
https://github.com/huggingface/huggingface.js/blob/2a7b6837e4d5c977927736856cfe7984cec8de55/packages/inference/src/types.ts#L23-L26

Something like:
```js
function myFetch(input, init) {
    return fetch(input, { ...init, credentials: undefined });
}

WalshyDev:
Ideally, this should be as frictionless as possible for edge runtimes. This would work but it isn't a great user experience having to define their own fetch to just remove this property. Of course, it isn't great to have to specify another param either so maybe my point is moot. I didn't notice the `fetch` option until I was moving the `includeCredentials` so it didn't cross my mind until earlier.

PR #249, opened 2023-09-25
Agents in chat ui

nsarrazin:
This breaks the demo that depend on #246 so I opened a separate PR. These are the changes required for chat-ui agents. still WIP


nsarrazin:
Will merge this if you think the agents stuff is good enough @gary149

gururise:
Are you still working on Agents?  Something like this would be really useful for attaching a VectorDB to Chat-UI to perform RAG.

PR #254, opened 2023-10-05
Safetensor parsing: validation & error handling

SBrandeis:
# TL;DR

- Add some validation in safetensors parsing
- Update the error messages to be more specific about what went wrong
- Wrap errors in a new Error class

### DIFF
--- a/packages/hub/src/lib/parse-safetensors-metadata.ts
+++ b/packages/hub/src/lib/parse-safetensors-metadata.ts
@@ -11,6 +11,9 @@ import { promisesQueue } from "../utils/promisesQueue";
 const SINGLE_FILE = "model.safetensors";
 const INDEX_FILE = "model.safetensors.index.json";
 const PARALLEL_DOWNLOADS = 5;
+const MAX_HEADER_LENGTH = 25_000_000;
+
+class SafetensorParseError extends Error { }
 
 type FileName = string;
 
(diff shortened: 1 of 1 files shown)


packages/hub/src/lib/parse-safetensors-metadata.ts
```diff
@@ -11,6 +11,9 @@ import { promisesQueue } from "../utils/promisesQueue";
 const SINGLE_FILE = "model.safetensors";
 const INDEX_FILE = "model.safetensors.index.json";
 const PARALLEL_DOWNLOADS = 5;
+const MAX_HEADER_LENGTH = 25_000_000;
```
SBrandeis, line 14: No idea if this value (25MB) is reasonable

packages/hub/src/lib/parse-safetensors-metadata.ts
```diff
@@ -70,23 +73,33 @@ async function parseSingleFile(
...
 	}
 
-	// no validation for now, we assume it's a valid FileHeader.
-	const header: SafetensorsFileHeader = await secondResp.json();
+	try {
+		// no validation for now, we assume it's a valid FileHeader.
+		const header: SafetensorsFileHeader = await secondResp.json();
+		return header;
+	} catch (err) {
+		throw new SafetensorParseError(`FFailed to parse file ${path}: safetensors header is not valid JSON.`)
```
SBrandeis, line 100:
```suggestion
		throw new SafetensorParseError(`Failed to parse file ${path}: safetensors header is not valid JSON.`)
```

packages/hub/src/lib/parse-safetensors-metadata.ts
julien-c: i think 25MB is more than enough for the JSON part!! That's one big JSON already.

cc @Narsil for visibility (we had a pathologically crafted safetensors file that crashed the Hub) – probably we don't need to put that in the spec

Kakulukian (approved):

Pierrci:
niiiiiice ❤️

packages/hub/src/lib/parse-safetensors-metadata.ts
Narsil: https://github.com/huggingface/safetensors/blob/main/safetensors/src/tensor.rs#L295

https://github.com/huggingface/safetensors/blob/main/safetensors/src/tensor.rs#L10

There's  a hard cap in the lib to prevent DOS attack (while deserializing super large JSON structures)

PR #260, opened 2023-10-13
add abort signal to streaming requests

### LINKED ISSUE #259: Aborting `generate_stream` on the inference package midway through processing
When using the `generate_stream` API (`hf.textGenerationStream` to be exact), is there any way to abort the SSE stream midway? I'm trying to implement custom stopping logic, and when I CURL from a terminal, I can close the connection midway, and the TGI server will recognize and processing the stream (https://github.com/huggingface/text-generation-inference/issues/1130). 

I noticed this repo uses Azure's fetch-event-source (https://github.com/Azure/fetch-event-source), so I'm wondering if there's a way to handle aborting via this library.

radames:
Add abort signal as option to interrupt fetch streaming. fix #259

### DIFF
--- a/packages/inference/src/tasks/custom/streamingRequest.ts
+++ b/packages/inference/src/tasks/custom/streamingRequest.ts
@@ -66,6 +66,9 @@ export async function* streamingRequest<T>(
 			if (done) return;
 			onChunk(value);
 			for (const event of events) {
+				if (options?.signal?.aborted) {
+					throw new Error("Request aborted");
+				}
 				if (event.data.length > 0) {
 					const data = JSON.parse(event.data);
 					if (typeof data === "object" && data !== null && "error" in data) {
--- a/packages/inference/test/HfInference.spec.ts
+++ b/packages/inference/test/HfInference.spec.ts
@@ -243,8 +243,26 @@ describe.concurrent(
 			});
 
 			await expect(response.next()).rejects.toThrow(
-				"Input validation error: `inputs` tokens + `max_new_tokens` must be <= 4096. Given: 17 `inputs` tokens and 10000 `max_new_tokens`"
+				"Input validation error: `inputs` tokens + `max_new_tokens` must be <= 2048. Given: 17 `inputs` tokens and 10000 `max_new_tokens`"
 			);
+		});
+
+		it("textGenerationStream - Abort", async () => {
+			const controller = new AbortController();
+			const response = hf.textGenerationStream(
+				{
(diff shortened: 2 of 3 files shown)


packages/inference/test/HfInference.spec.ts
```diff
@@ -243,10 +243,28 @@ describe.concurrent(
 			});
 
 			await expect(response.next()).rejects.toThrow(
-				"Input validation error: `inputs` tokens + `max_new_tokens` must be <= 4096. Given: 17 `inputs` tokens and 10000 `max_new_tokens`"
+				"Input validation error: `inputs` tokens + `max_new_tokens` must be <= 2048. Given: 17 `inputs` tokens and 10000 `max_new_tokens`"
```
radames, line 246: change this since got an error, but I'm not sure if it's needed

radames:
Fixing, something, just realized I'm aborting at the wrong place

packages/inference/src/tasks/custom/streamingRequest.ts
```diff
@@ -66,6 +66,9 @@ export async function* streamingRequest<T>(
 			if (done) return;
 			onChunk(value);
 			for (const event of events) {
+				if (options?.signal?.aborted) {
```
vvmnnnkv, line 69: It's probably more correct to pass `signal` into `fetch` to explicitly abort connection (adding it to `info` in `makeRequestOptions`).
And then handle AbortError here (or let it bubble to user, not sure).

radames: yes totally right, I realize something was wrong when I did the VCR 🥲 I was expecting like 2 tokens, and kept getting the 1000 tokens, while the stream was interrupt, the fetch kept going :)

radames:
while I run 
`npx vitest run  --testNamePattern "textGenerationStream - Abort"` I get the test to pass however if fails when I do ` VCR_MODE=record  npx vitest run  --testNamePattern "textGenerationStream - Abort"`  I think it's related to the cache for the same input, and when the response is cached, there is a single fetch, and the abort never happens

vvmnnnkv:
vcr.ts doesn't support signal nor it can properly record & playback streaming responses :)
this is doable but perhaps its more practical to just avoid caching such responses?

swarupsrini:
@radames just wanted to say thank you for the quick PR! I appreciate the amazing turnaround time for this, it will be really helpful  for us :)

PR #265, opened 2023-10-21
📝 Fix Hub Deno Import

ArsalaBangash:
I was going through the Hub documentation and noticed that the Deno import was HInference


ArsalaBangash:
Thank you for the clarification. 

I've incorporated your suggested changes.

PR #283, opened 2023-11-06
♻️ Use HEAD call for fileExists

coyotte508:
cc @julien-c (original implem of `fileExists`) @Pierrci


julien-c:
looks good, which issue is this fixing out of curiosity?

Pierrci:
Thanks! (this will make better use of our cache system)

PR #292, opened 2023-11-10
Languages subpackage

julien-c:


packages/languages/src/types.d.ts
```diff
@@ -0,0 +1,5 @@
+export interface Language {
+	code: string;
```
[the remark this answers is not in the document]
julien-c: i don't think stronger typing is necessary no (and I'm warry of putting too much strain on the typescript compiler given the number of keys...)

.github/workflows/languages-publish.yml
```diff
@@ -0,0 +1,54 @@
...
+          node-version: "18"
+          cache: "pnpm"
+          cache-dependency-path: |
+            packages/languages/pnpm-lock.yaml
+            packages/doc-internal/pnpm-lock.yaml
+          # setting a registry enables the NODE_AUTH_TOKEN env variable where we can set an npm token.  REQUIRED
+          registry-url: "https://registry.npmjs.org"
+      - run: pnpm install
+      - run: git config --global user.name machineuser
+      - run: git config --global user.email tech@huggingface.co
```
julien-c, line 33: by the way, this email probably needs to be changed everywhere to the new one

README.md
```diff
@@ -40,6 +40,7 @@ This is a collection of JS libraries to interact with the Hugging Face API, with
 - [@huggingface/inference](packages/inference/README.md): Use the Inference API to make calls to 100,000+ Machine Learning models, or your own [inference endpoints](https://hf.co/docs/inference-endpoints/)!
 - [@huggingface/agents](packages/agents/README.md): Interact with HF models through a natural language interface
 - [@huggingface/hub](packages/hub/README.md): Interact with huggingface.co to create or delete repos and commit / download files
+- [@huggingface/languages](packages/languages/README.md): List of all languages supported by the huggingface hub
```
julien-c, line 43: does not really need to be documented imo (it's mostly internal) but why not

julien-c:
> I did some changes @julien-c , let me know what you think
> 
> * Split into multiple files so that github lets us review non-data files
> * add tests
> * add a README.md
> 
> I also created a `language(code)` function & changed the arg of `wikiLink` to just be the language code

splitting files + tests is great

I will still need the dicts exported (i need them in the moon codebase) and also merged in a single dict

So personnally i'd remvoe the `language` function which does not bring anythign to my use case, but up to you if you want to keep it

packages/languages/src/language.ts
julien-c, line 1: i'd personally remove this

packages/languages/src/language.ts
```diff
@@ -0,0 +1,20 @@
+import { LANGUAGES_ISO_639_1 } from "./languages_iso_639_1";
+import { LANGUAGES_ISO_639_3 } from "./languages_iso_639_3";
+import type { Language } from "./types";
+
+/**
+ * Be careful importing this function in frontend code, as it will import all languages.
+ */
+export function language(code: keyof typeof LANGUAGES_ISO_639_1): Language;
+export function language(code: keyof typeof LANGUAGES_ISO_639_3): Language;
```
julien-c, line 9: this is probably hard to compute for `tsc`

packages/languages/src/index.ts
```diff
@@ -0,0 +1,5 @@
+export type { Language } from "./types";
+export { language } from "./language";
+export { wikiLink } from "./wikiLink";
+export { LANGUAGES_ISO_639_1 } from "./languages_iso_639_1";
+export { LANGUAGES_ISO_639_3 } from "./languages_iso_639_3";
```
julien-c, line 5:
```suggestion
export { LANGUAGES_ISO_639_3 } from "./languages_iso_639_3";

export const LANGUAGES = { ...LANGUAGES_ISO_639_1, ...LANGUAGES_ISO_639_3 };
```

PR #293, opened 2023-11-10
machineuser email change

julien-c:


julien-c:
Closing given it's being done in #292 too

PR #297, opened 2023-11-19
🎉 @huggingface/widgets

coyotte508:
https://kit.svelte.dev/docs/packaging

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
 		"prettier": "^3.0.0",
 		"prettier-plugin-svelte": "^3.0.0",
 		"typescript": "^5.0.0",
-		"vite": "^4.4.2"
+		"vite": "4.1.4"
 	},
 	"scripts": {
 		"lint": "eslint --quiet --fix --ext .cjs,.ts .eslintrc.cjs",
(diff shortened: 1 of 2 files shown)


julien-c:
cc @mishig25

packages/widgets/src/lib/interfaces/Types.ts
julien-c, line 1: the whole interfaces folder should be in its own package IMO

julien-c: it's more central than `widgets-data`, i had an idea for a name, let me try to find it again

julien-c: aarg can't find it, but what about:

`@huggingface/pipeline-types`
or 
`@huggingface/interfaces`
or 
`@huggingface/types` (this one prob looks weird)

Also ping @osanseviero on naming the root package that will include those files:

image

WDYT?

mishig25:
lgtm !

packages/widgets/src/lib/interfaces/Types.ts
julien-c: `Types.ts` is more central than widgets, which should be mostly self-contained

Otherwise let's move those files to `@huggingface/tasks` then. I think it works too

julien-c: ok let's move the whole interfaces folder to the root of `@huggingface/tasks` ok?

PR #317, opened 2023-11-22
[Task Label & Icon] for Image-to-Video

mishig25:
This is how it looks

image


packages/widgets/src/lib/components/Icons/IconImageToVideo.svelte
```diff
@@ -0,0 +1,5 @@
+<script lang="ts">
+	export let classNames = "";
+</script>
+
+<svg xmlns="http://www.w3.org/2000/svg" class={classNames} width="1em" height="1em" fill="none"><path fill="currentColor" fill-rule="evenodd" d="M1.43 1.2h6.63a.7.7 0 0 1 .7.7v4.44a.79.79 0 0 0-.34-.08h-.36V1.9H1.43v3.55l.9-.9a.7.7 0 0 1 .99 0l1.74 1.74a.79.79 0 0 0-.51.47L2.82 5.04l-1.4 1.4v2.09H4.5v.7H1.43a.7.7 0 0 1-.7-.7V1.9a.7.7 0 0 1 .7-.7ZM6.8 5.94l.32.32H5.5l.32-.32a.7.7 0 0 1 .99 0ZM6.2 4.52a1.05 1.05 0 1 1-1.17-1.75A1.05 1.05 0 0 1 6.2 4.52Zm-.4-1.16a.35.35 0 1 0-.38.58.35.35 0 0 0 .39-.58Zm3.9 5.16 1.57-1.26v3.14L9.7 9.15V10a.79.79 0 0 1-.79.79H5.77a.79.79 0 0 1-.79-.79V7.65a.79.79 0 0 1 .79-.78H8.9a.79.79 0 0 1 .79.78v.87Zm-3.93-.87v2.36H8.9V7.65H5.77Z" clip-rule="evenodd"/></svg>
```
mishig25, line 5: @coyotte508 I think the formatter is not configured properly?
Shouldn't it be multline like https://github.com/huggingface/huggingface.js/blob/ccfb044414a235270cb82d8b586f81785ad04eba/packages/widgets/src/lib/components/Icons/IconTextToImage.svelte#L5-L25 ?

packages/tasks/src/const.ts
```diff
@@ -18,6 +18,7 @@ export const TASKS_MODEL_LIBRARIES: Record<PipelineType, ModelLibraryKey[]> = {
 	"image-segmentation": ["transformers", "transformers.js"],
 	"image-to-image": [],
 	"image-to-text": ["transformers.js"],
+	"image-to-video": [], // TODO: diffusers will soon support image-to-video
```
osanseviero, line 21: Let's add it already, it will be ready this week

osanseviero, line 21: Let's add it already, it will be ready this week

```suggestion
	"image-to-video": ["diffusers"],
```

osanseviero (approved):
Thanks, LGTM!

packages/tasks/src/tasksData.ts
```diff
@@ -54,6 +54,7 @@ export const TASKS_DATA: Record<PipelineType, TaskData | undefined> = {
 	"image-segmentation": getData("image-segmentation", imageSegmentation),
 	"image-to-image": getData("image-to-image", imageToImage),
 	"image-to-text": getData("image-to-text", imageToText),
+	"image-to-video": undefined,
```
pcuenca, line 57: Just a question to help me understand how it works, would creating an entry in `tasks/data` and populating it here would be enough to have the task listed in https://huggingface.co/tasks ?

pcuenca: Awesome, thanks!

packages/tasks/src/const.ts
mishig25: @osanseviero shouldn't this line also be diffusers ? https://github.com/huggingface/huggingface.js/blob/6764b0e2e256fd79cd718ec6d90fbcccedecea9c/packages/tasks/src/const.ts#L39

mishig25: wdyt e1d2b92c9207e7162f423169b4af437966a80ae5 ?

julien-c: yep TBH i didn't even remember the tasks/src/const.ts one

osanseviero: Yes @mishig25 but let's keep this PR for image-to-video? For text-to-image actually I have https://github.com/huggingface/huggingface.js/pull/320 being updated with the migration

osanseviero: They are different though. `LibrariesToTasks` defines which tasks are supported by the Inference API for the specific library. This file is automatically generated by a GitHub Actions workflow in `api-inference-community`. `const.ts` is about library support, not specific to Inference API

mishig25: handled in cb0f27702879889c1e18102113a9ea6764946f5c and 6764b0e2e256fd79cd718ec6d90fbcccedecea9c

PR #319, opened 2023-11-22
[Inference Widget] Show widget displayability on tasks without widgets

mishig25:
Closes https://github.com/huggingface/moon-landing/issues/8165

### Before

There was no error message

### After 

Show the error message

image


osanseviero (approved):
Thanks!

PR #327, opened 2023-11-22
Fix of feature-extraction metrics

osanseviero:
Check how the metrics section show up in https://huggingface.co/tasks/feature-extraction


merveenoyan (approved):
Must be left from placeholder for when external contributors were contributing and I've missed this during reviews

PR #347, opened 2023-11-22
Update TTS Tasks page.

Vaibhavs10:
Wee update to the models listed, just making it a bit more current.

### DIFF
--- a/packages/tasks/src/text-to-speech/about.md
+++ b/packages/tasks/src/text-to-speech/about.md
@@ -25,7 +25,7 @@ def query(payload):
 	response = requests.post(API_URL, headers=headers, json=payload)
 	return response
 
-output = query({"text_inputs": "This is a test"})
+output = query({"text_inputs": "Max is the best doggo."})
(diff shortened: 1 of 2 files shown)


packages/tasks/src/text-to-speech/about.md
```diff
@@ -25,7 +25,7 @@ def query(payload):
 	response = requests.post(API_URL, headers=headers, json=payload)
 	return response
 
-output = query({"text_inputs": "This is a test"})
+output = query({"text_inputs": "Max is the best doggo."})
```
pcuenca, line 28: 😂

pcuenca (approved):
Sounds good!

packages/tasks/src/text-to-speech/about.md
Vaibhavs10: Facts are facts!

Vaibhavs10:
I forgot what's the protocol here. Is this good to merge?

PR #348, opened 2023-11-22
💄 Add bakc fonts to demo

coyotte508:
We can probably do with less fonts, I just translated the old google font links

image

vs old: https://huggingface.co/spaces/huggingfacejs/inference-widgets

(to test lcoally: `pnpm install && pnpm --filter widgets dev`)


mishig25:
there is no changes needed in tailwind.config.cjs ?

julien-c:
what are pros & cons vs. using the same system than in moon-landing (links in layout.html)

mishig25 (approved):

PR #351, opened 2023-11-22
Update Text-generation tasks page.

Vaibhavs10:
I think we can remove some models from here btw.
But good to keep them too. Just some changes to make the page not sound ancient.


osanseviero (approved):
This LGTM for now :fire: Maybe add a link to the leaderboard somewhere

merveenoyan (approved):

PR #356, opened 2023-11-24
🚚 Move all TS exports to @huggingface/tasks

coyotte508:


julien-c (approved):

julien-c (commented):
@coyotte508 if you want you can move the many folders (one for each task) into a `task-pages` subfolder

It's a bit messy to have them at the root IMO

julien-c (commented):
lets try to minimize the file layout diff compared to before

mishig25:
Nit: this [README](https://github.com/huggingface/huggingface.js/blob/refacto-end/packages/tasks/src/README.md?plain=1) is assuming that there is a folder called `Interfaces`

packages/tasks/src/tasks/audio-classification/data.ts
```diff
@@ -1,4 +1,4 @@
-import type { TaskDataCustom } from "../Types";
+import type { TaskDataCustom } from "..";
```
mishig25, line 1: there has to be something here right ?

mishig25: Oh I see that `index.ts` is being inferred

mishig25:
No string preference. But we can leave [Widgets Input/Output types](https://github.com/huggingface/huggingface.js/blob/refacto-end/packages/tasks/src/model-data.ts#L5-L25) in hf.js/widgets rather than moving them to hf.js/tasks ?

PR #357, opened 2023-11-24
Update list of transformers.js-compatible tasks (and add image-to-image subtasks)

xenova:
This PR does two things:
1. Updates the list of tasks which are compatible with Transformers.js
2. Adds a list of subtasks for the `image-to-image` task (list obtained from [here](https://huggingface.co/tasks/image-to-image)).

cc @osanseviero


osanseviero (approved):
Thank you!

PR #359, opened 2023-11-24
Add GGUF to model libraries

krampstudio:


julien-c (changes requested):
see the comments here:

> This is for modeling (= architectures) libraries, not for file formats (like ONNX, etc).
> File formats live in an enum inside the internal codebase.

I don't think this PR is needed!

PR #363, opened 2023-11-24
Add API snippet zero-shot img classification

osanseviero:
TODO (in other PRs)

- [x] Zero shot img classification
- [ ] Image to image
- [ ] Tabular Classification
- [ ] Tabular Regression
- [ ] Visual Question Answering
- [ ] Document Question Answering
- [ ] Zero shot object detection ? (need to investigate this one a bit more)

### DIFF
--- a/packages/tasks/src/snippets/python.ts
+++ b/packages/tasks/src/snippets/python.ts
@@ -15,12 +15,12 @@ output = query({
 export const snippetZeroShotImageClassification = (model: ModelData): string =>
 	`def query(data):
 	with open(data["image_path"], "rb") as f:
-		data = f.read()
+		img = f.read()
 	payload={
 		"parameters": data["parameters"],
-		"inputs": base64.b64encode(data).decode("utf-8")
+		"inputs": base64.b64encode(img).decode("utf-8")
 	}
-	response = requests.post(API_URL, headers=headers,json=payload)
+	response = requests.post(API_URL, headers=headers, json=payload)
 	return response.json()
 
 output = query({
--- a/packages/tasks/src/snippets/python.ts
+++ b/packages/tasks/src/snippets/python.ts
@@ -24,7 +24,7 @@ export const snippetZeroShotImageClassification = (model: ModelData): string =>
 	return response.json()
 
 output = query({
-    "inputs": ${getModelInputSnippet(model)},
+    "image_path": ${getModelInputSnippet(model)},
     "parameters": {"candidate_labels": ["cat", "dog", "llama"]},
 })`;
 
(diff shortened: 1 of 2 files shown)


packages/tasks/src/snippets/python.ts
```diff
@@ -12,6 +12,22 @@ output = query({
...
+		img = f.read()
+	payload={
+		"parameters": data["parameters"],
+		"inputs": base64.b64encode(img).decode("utf-8")
+	}
+	response = requests.post(API_URL, headers=headers, json=payload)
+	return response.json()
+
+output = query({
+    "inputs": ${getModelInputSnippet(model)},
```
merveenoyan, line 27: I'm a little confused, aren't we passing the example input (which is the image path in the other file)?
if so, isn't there supposed to be an `image_path` key in the dict that is passed to the query function? 

if so it should be:
output = query({
    "image_path": ${getModelInputSnippet(model)},

merveenoyan (commented):
thank you!

packages/tasks/src/snippets/python.ts
osanseviero: Good catch! Fixed it

packages/tasks/src/snippets/python.ts
```diff
@@ -12,6 +12,22 @@ output = query({
     "parameters": {"candidate_labels": ["refund", "legal", "faq"]},
 })`;
 
+export const snippetZeroShotImageClassification = (model: ModelData): string =>
+	`def query(data):
+	with open(data["image_path"], "rb") as f:
+		img = f.read()
+	payload={
+		"parameters": data["parameters"],
+		"inputs": base64.b64encode(img).decode("utf-8")
```
pcuenca, line 21: Interesting! So the input image is received as binary data, I guess. Is there a way to test and debug these snippets before merging?

pcuenca (approved):
Looks good!

packages/tasks/src/snippets/python.ts
osanseviero: Yes, you can test the python snippet locally or in Colab

```python
import requests
import base64

API_URL = "https://api-inference.huggingface.co/models/openai/clip-vit-base-patch32"

def query(data):
	with open(data["image_path"], "rb") as f:
		img = f.read()
	payload={
		"parameters": data["parameters"],
		"image":  base64.b64encode(img).decode("utf-8"),
	}
	response = requests.post(API_URL, json=payload)
	return response.json()
 
output = query({
	"image_path": "/content/download.png",
	"parameters": {"candidate_labels": ["cat", "dog", "llama"]},
})
```

for example

pcuenca: Oh, of course! Thanks a lot! 🙌

PR #365, opened 2023-11-26
:arrow_up: upgrade vitest, @vitest/browser

### LINKED ISSUE #267: Upgrade vitest & @vitest/browser
Currently it breaks tests for agents for example, because `process.env` is no longer defined

vvmnnnkv:
Fixes #267 

inference package used to substitute "process.env" with `process.env` contents in `vitest.config.ts` (using `define`), this is broken in more recent version of vitest: https://github.com/vitest-dev/vitest/issues/3900.

So the solution is to use [`import.meta.env`](https://vitejs.dev/guide/env-and-mode.html#env-variables). But since it passes through only vars with "VITE_" prefix, `envPrefix` is needed include vars starting with `HF_` and `VCR_` .

Additionally, moved `expect.closeTo()` to a separate file, it can be removed with update to vitest 1.0.


vvmnnnkv:
> If the demo app of `widgets` still works with vite 5 (`pnpm --filter widgets dev`), good for me!

@coyotte508, it does! :-)
How about moving typescript to common package.json as well?

vvmnnnkv:
Back to vite 4 for widgets, there's no official vite 5 support in sveltekit yet

vvmnnnkv:
Fixes https://github.com/huggingface/huggingface.js/security/dependabot/10

PR #366, opened 2023-11-27
Added doc-vqa snippet

merveenoyan:
@osanseviero as discussed on Slack.
I wanted to repurpose the task page's schema image but I can't find the assets folder where we used to put them

### DIFF
--- a/packages/tasks/src/snippets/inputs.ts
+++ b/packages/tasks/src/snippets/inputs.ts
@@ -31,6 +31,12 @@ const inputsTableQuestionAnswering = () =>
 		}
 	}`;
 
+const inputsDocumentQuestionAnswering = () =>
+	`{
+		"image": "document-question-answering-input.png",
+		"question": "What is the idea behind the consumer relations efficiency team?"
+	}`;
+
 const inputsQuestionAnswering = () =>
 	`{
 		"question": "What is my name?",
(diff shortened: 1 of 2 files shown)


osanseviero (changes requested):
Do you have an example model for which this snippet works? For 3 models, I got

> {'error': ['Error in `inputs.image`: Incorrect padding']}

I think you need to encode the image as done in #363, so something among these lines

```python
def query(data):
  with open(data["image"], "rb") as f:
    img = f.read()
  payload={
    "question": data["question"],
    "image": base64.b64encode(img).decode("utf-8")
  }
  response = requests.post(API_URL, headers=headers,  json=payload)
  return response.json()

output = query({
    "image": "img.png",
    "question": "What is the in this image?",
})
```

outputs

```
[{'score': 0.19945354759693146,
  'answer': 'Locate in Drive',
  'start': 12,
  'end': 14}]
```

merveenoyan:
@osanseviero ah I thought we were providing something minimal, I didn't include encoding. Thanks, I will reflect them.

osanseviero:
> I thought we were providing something minimal

We're aiming to provide a minimal working snippet

```
output = query({
    "image": "img.png",
    "question": "What is the in this image?",
})
```

won't work I think (at least not from a quick test), which would confuse users

osanseviero (approved):
LGTM, thanks! Let's wait for @pcuenca's review

packages/tasks/src/snippets/inputs.ts
```diff
@@ -31,6 +31,12 @@ const inputsTableQuestionAnswering = () =>
 		}
 	}`;
 
+const inputsDocumentQuestionAnswering = () =>
```
osanseviero, line 34: Note that this same input and snippet will work for Visual Question Answering, so we should reuse them if possible

PR #371, opened 2023-11-28
 Exclude Autogenerated Files for Accurate Checksum in CI Workflow

### LINKED ISSUE #361: Add check to make sure `@huggingface/tasks` is published _before_ `@huggingface/widgets`
We could move to changesets (https://github.com/huggingface/huggingface.js/issues/6, but it was getting complicated esp. with publishing to both github packages & npm)

or add a check in `widgets-publish.yml`, that checks the remote package's checksum against the local one, and error if it doesn't match. (preffered solution IMO) (or even, we could then publish both)

kohsheen1234:
Follow up #368 
Fixes #361 
This PR resolves inconsistent checksums in  GitHub Actions for @huggingface/tasks by excluding `node_modules `and `dist` files from checksum calculations. These changes ensure checksum consistency across environments, as demonstrated in multiple workflow runs: 

In both screenshots, it's evident that the checksums are consistently matched, with identical values for Local and Remote in each respective run.

[Run 1 job/19077519008](https://github.com/kohsheen1234/huggingface.js/actions/runs/7012643832/job/19077519008), 
Screenshot 2023-11-27 at 7 54 03 PM

[Run 2 job/19077523195](https://github.com/kohsheen1234/huggingface.js/actions/runs/7012645392/job/19077523195). 
Screenshot 2023-11-27 at 7 54 24 PM

### DIFF
--- a/.github/workflows/widgets-publish.yml
+++ b/.github/workflows/widgets-publish.yml
@@ -38,14 +38,12 @@ jobs:
           git commit . -m "🔖 @huggingface/widgets $BUMPED_VERSION"
           git tag "widgets-v$BUMPED_VERSION"
       - run: |
-
           # Clean up the workspace
           rm -rf ../tasks/node_modules ../tasks/dist
 
           # Calculate checksum of the cleaned package
           echo "Calculating checksum of local @huggingface/tasks package..."
           LOCAL_CHECKSUM=$(cd ../tasks && tar --mtime='1970-01-01' --mode=755 -cf - . | sha256sum | cut -d' ' -f1)
-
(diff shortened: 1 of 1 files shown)


kohsheen1234:
@coyotte508

During my tests on PR #368, I discovered the root cause of inconsistent checksums: the generation of dist files within the automated CI/CD pipeline. This issue arose from the dynamic nature of these files, as they're created anew during each pipeline execution, especially in preparation for deployment. 

To address this, I've implemented a step in this PR to remove both node_modules and dist directories before the checksum calculation. This adjustment ensures that our checksums remain consistent across different runs, aligning them with the static parts of the codebase.

PR #376, opened 2023-11-28
✨ Add "sign-in with HF" in widgets demo

coyotte508:
image

This uses the "inference-api" scope to make calls to the inference API (cc @Wauplin @huggingface/moon-landing @Narsil   for viz)

Once merged, https://huggingface.co/spaces/huggingfacejs/inference-widgets will have a sign-in with HF button

### DIFF
--- a/packages/widgets/src/routes/+page.svelte
+++ b/packages/widgets/src/routes/+page.svelte
@@ -14,6 +14,10 @@
 		window.localStorage.setItem("hf_token", apiToken);
 	}
 
+	/**
+	 * If we are in an iframe, we need to open the auth page in a new tab
+	 * to avoid issues with third-party cookies in a space
+	 */
 	const isIframe = browser && window.self !== window.parent;
 
 	onMount(() => {


packages/widgets/package.json
```diff
@@ -38,6 +38,8 @@
 		"svelte": "^3.59.2"
 	},
 	"devDependencies": {
+		"@auth/core": "^0.18.3",
+		"@auth/sveltekit": "^0.3.14",
```
mishig25, line 42: they should be in dep, not devDep ?

packages/widgets/src/hooks.server.ts
```diff
@@ -0,0 +1,59 @@
+import { env } from "$env/dynamic/private";
```
mishig25, line 1: does this hook file run when a widget is being server from moon ?

mishig25: oh I see

mishig25 (approved):
lgtm !

PR #380, opened 2023-11-29
✨ Use @huggingface/tasks as source of truth in @huggingface/hub and @huggingface/inference

coyotte508:
cc @mishig25 @osanseviero 

With this no need to update `@huggingface/hub` & `@huggingface/inference` each time a task is added

Note: `@huggingface/tasks` is added as a dev dependency, but does not appear in the final bundle:

- in the case of `@huggingface/hub`, the `--dts-resolve` config will copy over the raw types
- in the case of `@huggingface/inference`, the `generate-dts` script will take care of it

### DIFF
--- a/packages/hub/src/types/public.d.ts
+++ b/packages/hub/src/types/public.d.ts
@@ -45,7 +45,7 @@ export type AccessTokenRole = "admin" | "write" | "contributor" | "read";
 
 export type AuthType = "access_token" | "app_token" | "app_token_as_user";
 
-export type Task = PipelineType;
+export type { PipelineType };
 
 export interface SpaceRuntime {
 	stage: SpaceStage;
--- a/packages/hub/src/index.ts
+++ b/packages/hub/src/index.ts
@@ -5,6 +5,7 @@ export type {
 	AccessTokenRole,
 	AuthType,
 	Credentials,
+	PipelineType,
 	RepoDesignation,
 	RepoFullName,
 	RepoId,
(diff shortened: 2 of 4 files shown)


julien-c (commented):
nice DX improvement 🎉

packages/hub/src/types/public.d.ts
```diff
@@ -43,49 +45,7 @@ export type AccessTokenRole = "admin" | "write" | "contributor" | "read";
...
-	| "table-to-text"
-	| "multiple-choice"
-	| "text-retrieval"
-	| "time-series-forecasting"
-	| "visual-question-answering"
-	| "document-question-answering"
-	| "zero-shot-image-classification"
-	| "graph-ml"
-	| "other";
+export type Task = PipelineType;
```
julien-c, line 48: change all the references and only keep `PipelineType` no?.

packages/inference/src/types.ts
```diff
@@ -35,39 +37,7 @@ export interface Options {
...
-	| "text-to-image"
-	| "text-to-speech"
-	| "text-to-video"
-	| "token-classification"
-	| "translation"
-	| "unconditional-image-generation"
-	| "visual-question-answering"
-	| "zero-shot-classification"
-	| "zero-shot-image-classification";
+export type InferenceTask = Exclude<PipelineType, "other">;
```
julien-c, line 40: this could be defined in tasks directly as it might be helpful elsewhere too.

You could name it `ConcretePipelineType` or something (or `InferenceTask` is fine too i guess)

packages/hub/src/types/public.d.ts
julien-c: the canonical name accross the HF ecosystem is more PipelineType than Task

vvmnnnkv (approved):

PR #381, opened 2023-11-29
[Widget] Remove Intel CPU message

mishig25:
Remove Intel CPU message, which makes `WidfetInfo` go back [to this state](https://github.com/huggingface/hub-docs/blob/ec705d08683a2b7171f219342cf6f1fca35fc534/js/src/lib/components/InferenceWidget/shared/WidgetInfo/WidgetInfo.svelte) before https://github.com/huggingface/hub-docs/pull/456


julien-c (approved):
good for me code-wise if @jeffboudier does not object

jeffboudier:
for context we're negotiating the new SOW with Intel with this thing an open point 
Screen Shot 2023-11-29 at 11 29 04 AM

happy to remove this now, there's a probability we want to put it back later.

julien-c:
ok let's merge!

PR #382, opened 2023-11-29
✨ Save login after "sign in with HF"

coyotte508:
image

Also make the cookie expire after 1 hour (until we have a refresh token mechanism maybe)

cc @julien-c for the suggestion (and @osanseviero @SBrandeis who reacted to the suggestion)


mishig25:
🔥

PR #384, opened 2023-11-29
[Widget] Refactor `WidgetWrapper`

mishig25:
### Refactor `WidgetWrapper`

(continuing the widget rewrite)

1. We discussed about removing `WidgetWrapper`. Instead, `WidgetWrapper` is still kept but simplified. Specificlly, `WidgetWrapper` does 2 tasks: 1. common styling for all widgets (just like `App` inside moon); 2. gets a model status (is it loaded, loadable, or too big)
2. With the simplified `WidgetWrapper`, there is no longer `WidgetHeader` or `WidgetFooter` inside `WidgetWrapper`. Therefore, every widget has to explicitly call `WidgetHeader` or `WidgetFooter`.
3. Refactored `WidgetExamples` component: 1. to use less reactive statements `$:`; 2. have better variable names (favouring `exampleXYZ` over `inputSampleXYZ`); 3. `WidgetExamples` has the logic to run a random widget example `onMount` (previously, this logic was inside `WidgetWrapper`)
4. Removed widget maximize feature (cc: @gary149) from `WidgetFooter`. I don't think anyone was using. But happy to put it back.


mishig25:
> not related to this PR, but is there any way to generate the validators from some sort of task-specific JSONschema (one for each task)?

Yes. There is a lib called ajv https://github.com/ajv-validator/ajv. But I think I/we can implement something simple/small in our use case.

> i.e. we'd have a single source of truth to define the shape of inputs/outputs

indeed, that would be great. Can also be used to auto generate the docs page https://huggingface.co/docs/hub/models-widgets-examples#widget-examples

julien-c:
> There is a lib called ajv

lol, going full circle (we used to use ajv before switching to Joi)

> maybe after we switch to using `@huggingface/inference`? Since the vlaidation is also done there

Yes why not, and anyways the single source of truth's idea is that it would also be used in our Python libraries (cc @Wauplin)

> switch to using `@huggingface/inference`

how complex/disruptive would this be? (and what the other advantages?)

mishig25:
> maybe after we switch to using @huggingface/inference? Since the vlaidation is also done there

@huggingface/inference validates api-output. Does it also validate input to the api? Also, we need to validate widget inputs and outputs (which can differ from inference input/output). For instance, on Text2Img widget, inference output is fileBlob while example output is url to file

packages/widgets/src/lib/components/InferenceWidget/shared/WidgetWrapper/WidgetWrapper.svelte
```diff
@@ -1,151 +1,49 @@
...
-				inputGroups.push({ group: inputSample.group as string, inputSamples: [] });
-			}
-			inputGroups.find(({ group }) => group === inputSample.group)?.inputSamples.push(inputSample);
-		}
-		return inputGroups;
-	}
+	// if WidgetHeader determines that there are no examples with outputs AND model.inference !== InferenceDisplayability.Yes
+	// then WidgetHeader will call this function to show InferenceDisplayability error to the user without showing anything else
+	setContext("makeWidgetUndisplayable", () => {
+		widgetUndisplayable = true;
```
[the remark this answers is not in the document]
mishig25: handled in adaea0b900f28b9f07cd740bf5531494f9d52740

packages/widgets/src/lib/components/InferenceWidget/shared/WidgetHeader/WidgetHeader.svelte
```diff
@@ -1,16 +1,46 @@
...
+
+	$: examplesAll = getExamples(isDisabled);
+
+	function getExamples(isDisabled: boolean): TWidgetExample[] {
+		const examples = ((model?.widgetData ?? []) as TWidgetExample[])
+			.filter((sample) => validateExample(sample) && (!isDisabled || sample.output !== undefined))
+			.sort((sample1, sample2) => (sample2.example_title ? 1 : 0) - (sample1.example_title ? 1 : 0));
+
+		if (isDisabled && !examples.length) {
+			const makeWidgetUndisplayable = getContext<() => void>("makeWidgetUndisplayable");
```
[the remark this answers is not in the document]
mishig25: same as https://github.com/huggingface/huggingface.js/pull/384#discussion_r1415598040

packages/widgets/src/lib/components/InferenceWidget/widgets/AudioClassificationWidget/AudioClassificationWidget.svelte
```diff
@@ -156,43 +156,40 @@
...
-			{#if warning}
-				<div class="alert alert-warning mt-2">{warning}</div>
-			{/if}
-		</form>
-	</svelte:fragment>
-	<svelte:fragment slot="bottom">
-		<WidgetOutputChart classNames="pt-4" {output} />
-	</svelte:fragment>
+	<WidgetHeader {noTitle} {model} {isLoading} {isDisabled} {callApiOnMount} {applyWidgetExample} {validateExample} />
+	<form>
```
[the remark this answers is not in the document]
mishig25: handled in 9628f72f3c85cc4556ba5d09682696ba600e1025

> put the form in WidgetWrapper (replace the div by a form),

took this approach

SBrandeis:
Reviewing this now

packages/widgets/src/lib/components/InferenceWidget/shared/WidgetHeader/WidgetHeader.svelte
```diff
@@ -1,16 +1,47 @@
...
+	const pipeline = model?.pipeline_tag;
 
 	$: task = pipeline ? getPipelineTask(pipeline) : undefined;
+
+	$: examplesAll = getExamples(isDisabled);
+
+	function getExamples(isDisabled: boolean): TWidgetExample[] {
+		const examples = ((model?.widgetData ?? []) as TWidgetExample[])
+			.filter((sample) => validateExample(sample) && (!isDisabled || sample.output !== undefined))
+			.sort((sample1, sample2) => (sample2.example_title ? 1 : 0) - (sample1.example_title ? 1 : 0));
```
SBrandeis, line 35: No need to cast here
```suggestion
		const examples = (model?.widgetData ?? [])
			.filter((sample): sample is TWidgetExample => validateExample(sample) && (!isDisabled || sample.output !== undefined))
```

packages/widgets/src/lib/components/InferenceWidget/shared/helpers.ts
```diff
@@ -64,6 +64,10 @@ export async function getBlobFromUrl(url: string): Promise<Blob> {
 	return blob;
 }
 
+export type ApplyExampleFunc = (sample: WidgetExample, opts?: ExampleRunOpts) => Promise<void>;
+
+export type ValidateExampleFunc = (sample: WidgetExample) => sample is WidgetExample;
+
```
SBrandeis, line 70: Unused
```suggestion
```

packages/widgets/src/lib/components/InferenceWidget/shared/WidgetHeader/WidgetHeader.svelte
```diff
@@ -1,16 +1,47 @@
...
+	export let model: WidgetProps["model"];
 	export let noTitle = false;
 	export let title: string | null = null;
-	export let pipeline: PipelineType | undefined;
+	export let isLoading = false;
 	export let isDisabled = false;
+	export let applyWidgetExample: (sample: TWidgetExample, opts?: ExampleRunOpts) => void = () => {};
+	export let validateExample: (sample: WidgetExample) => sample is TWidgetExample = (
+		sample
+	): sample is TWidgetExample => true;
```
SBrandeis, line 22: A bit confused by those defaults, maybe safer to have `undefined` as a default?

packages/widgets/src/lib/components/InferenceWidget/shared/WidgetHeader/WidgetHeader.svelte
```diff
@@ -1,16 +1,47 @@
...
+
+	function getExamples(isDisabled: boolean): TWidgetExample[] {
+		const examples = ((model?.widgetData ?? []) as TWidgetExample[])
+			.filter((sample) => validateExample(sample) && (!isDisabled || sample.output !== undefined))
+			.sort((sample1, sample2) => (sample2.example_title ? 1 : 0) - (sample1.example_title ? 1 : 0));
+
+		// if there are no examples with outputs AND model.inference !== InferenceDisplayability.Yes
+		// then widget will show InferenceDisplayability error to the user without showing anything else
+		if (isDisabled && !examples.length) {
+			$widgetNoInference[model.id] = true;
```
SBrandeis, line 40: Unrelated to the current PR: do stores get persisted when refreshing the page?

What happens when the store value changes? (Are the consuming components getting updated?)

packages/widgets/src/lib/components/InferenceWidget/widgets/AudioClassificationWidget/AudioClassificationWidget.svelte
```diff
@@ -156,43 +156,40 @@
...
 	{model}
-	{modelLoading}
-	{noTitle}
-	{outputJson}
-	{validateExample}
+	let:isDisabled
+	let:modelLoadInfo
+	let:WidgetInfo
+	let:WidgetHeader
+	let:WidgetFooter
```
SBrandeis, line 166: Maybe it would be simpler / cleaner if `isDisabled` / `modelLoadInfo` were in a store?
From what I understand they are only exposed to be "re-injected" in `WidgetHeader` / `WidgetInfo`, correct?

SBrandeis (commented):
Nice refactor.
I haven't tested all the widgets.
I was a bit confused at first by the examples logic living in `WidgetHeader`, but that's probably just a wording issue.

packages/widgets/src/lib/components/InferenceWidget/shared/WidgetFooter/WidgetFooter.svelte
```diff
@@ -22,14 +20,6 @@
...
 		</button>
 	{/if}
-	<button class="ml-auto flex items-center" on:click|preventDefault={() => (isMaximized = !isMaximized)}>
-		<IconMaximize classNames="mr-1" />
-		{#if !isMaximized}
-			Maximize
-		{:else}
-			Minimize
-		{/if}
-	</button>
```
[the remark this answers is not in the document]
gary149: I'm not sure, I would have kept it as it was 👀

packages/widgets/src/lib/components/InferenceWidget/shared/WidgetHeader/WidgetHeader.svelte
mishig25: >do stores get persisted when refreshing the page?

Svelte stores don't persist.

> What happens when the store value changes? (Are the consuming components getting updated?)

if a consuming component had UI or reactive statement dependent on a store, that UI or reactive statement gets updated when store gets a update

mishig25:
> I was a bit confused at first by the examples logic living in WidgetHeader, but that's probably just a wording issue.

WidgetExamples component is inside WidgetHeader. And there is a widget examples logic inside WidgetHeader that does:
```svelte
<script>
...
const validExamples = areThereValidWidgetExamples()
...
</script>

<div>
{if validExamples}
<WidgetExamples>
{/if}
</div>
```

but I think we can move all the logic widgets inside  <WidgetExamples>

packages/widgets/src/lib/components/InferenceWidget/shared/WidgetFooter/WidgetFooter.svelte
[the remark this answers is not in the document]
mishig25: handled in edb463bf1a2bd76cde3092605724d754158c99d6

packages/widgets/src/lib/components/InferenceWidget/shared/WidgetExamples/WidgetExamples.svelte
```diff
@@ -0,0 +1,173 @@
...
+					// @ts-expect-error complicated type
+					exampleFromQueryParams[key] = val;
+				}
+			}
+			if (Object.keys(exampleFromQueryParams).length) {
+				// run widget example from query params
+				applyWidgetExample(exampleFromQueryParams);
+			} else {
+				// run random widget example
+				const example = getWidgetExample<TWidgetExample>(model, validateExample);
```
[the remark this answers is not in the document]
mishig25: handled as part of f22f63ed9420211710a738c4bc228bc189407a1e

packages/widgets/src/lib/components/InferenceWidget/shared/WidgetHeader/WidgetHeader.svelte
mishig25: handled as part of f22f63ed9420211710a738c4bc228bc189407a1e

packages/widgets/src/lib/components/InferenceWidget/widgets/AudioClassificationWidget/AudioClassificationWidget.svelte
mishig25: handled in 9861a29480af975b2b37953aa9f621df87e9833c

SBrandeis (approved):
Ok LGTM
Sorry for the delay in reviewing
Note that different widgets for the same model on the same page will share some state due to stores

PR #389, opened 2023-12-04
[Widget] get model status only on `InferenceDisplayability.Yes` models

mishig25:
Context: [internal slack msg](https://huggingface.slack.com/archives/C02EMARJ65P/p1701705716046049?thread_ts=1701703726.990099&cid=C02EMARJ65P)

To get `model status`, GET request `https://api-inference.huggingface.co/status/{MODEL-ID}` is used to check status of a model on [hf inference api](https://huggingface.co/inference-api).

This PR makes it so that only when a model.inference is [InferenceDisplayability.Yes](https://github.com/huggingface/huggingface.js/blob/get_model_status_oninfyes/packages/tasks/src/model-data.ts#L122-L123), get the `model status`.

### DIFF
--- a/packages/widgets/src/lib/components/InferenceWidget/shared/WidgetWrapper/WidgetWrapper.svelte
+++ b/packages/widgets/src/lib/components/InferenceWidget/shared/WidgetWrapper/WidgetWrapper.svelte
@@ -73,15 +73,17 @@
 
 	onMount(() => {
 		(async () => {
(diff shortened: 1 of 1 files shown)


osanseviero (approved):
Thank you :+1: (conceptual LGTM)

julien-c:
i think it's useful if we wait for a @SBrandeis review (even a quick glance one) for widget-related things from now on. Ok for everyone?

PR #394, opened 2023-12-05
🔧 Add @SBrandeis as codeowners of widgets

coyotte508:
And cc @huggingface/moon-landing-back if anyone wants to be added to any module (like `@huggingface/tasks`)

Being part of CODEOWNERS is mainly to get notifications for only that part of the repo


SBrandeis (approved):
Thanks!

PR #405, opened 2023-12-08
🐛 [Widgets] Fix file/blob input

coyotte508:
Works now:

image

Local file still works:

image

### DIFF
--- a/packages/widgets/src/lib/components/InferenceWidget/shared/helpers.ts
+++ b/packages/widgets/src/lib/components/InferenceWidget/shared/helpers.ts
@@ -101,11 +101,7 @@ export async function callInferenceApi<T>(
 	useCache = true
 ): Promise<Result<T>> {
 	const contentType =
-		"file" in requestBody &&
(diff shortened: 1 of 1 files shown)


packages/widgets/src/lib/components/InferenceWidget/shared/helpers.ts
```diff
@@ -124,8 +120,8 @@ export async function callInferenceApi<T>(
 		headers.set("X-Load-Model", "0");
 	}
 
-	const reqBody: File | string =
-		"file" in requestBody && requestBody["file"] instanceof File ? requestBody.file : JSON.stringify(requestBody);
+	const reqBody: Blob | string =
```
mishig25, line 123:
```suggestion
	// `File` is a subtype of `Blob`: therefore, checking for instanceof `Blob` also checks for instanceof `File`
	const reqBody: Blob | string =
```

I'd just add this comment

mishig25 (approved):
lgtm !

PR #406, opened 2023-12-08
[Widget fix] Check for instance of Blob rather than File

### LINKED ISSUE #404: Audio speech recognition does not send audiofile to backend
cc @mishig25 @SBrandeis for visibility

Can test on https://huggingfacejs-inference-widgets.hf.space/

image

mishig25:
Closes #404 Follow up to #297

Widgets can send work with either `Blob` or `File`. For instance, [here](https://github.com/huggingface/huggingface.js/blob/fix_blob_check/packages/widgets/src/lib/components/InferenceWidget/widgets/ObjectDetectionWidget/ObjectDetectionWidget.svelte#L138-L139) blob is created. However, the previous [instanceof check](https://github.com/huggingface/huggingface.js/pull/406/files) was checking for `File` only. When the instanceof `Blob`, the check was failing;

### DIFF
--- a/packages/widgets/src/lib/components/InferenceWidget/shared/helpers.ts
+++ b/packages/widgets/src/lib/components/InferenceWidget/shared/helpers.ts
@@ -124,8 +124,9 @@ export async function callInferenceApi<T>(
 		headers.set("X-Load-Model", "0");
 	}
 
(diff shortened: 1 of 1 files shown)


mishig25:
Closing in favour for #405

PR #426, opened 2023-12-19
[tasks] Update type of `ModelData.base_model`

julien-c:
needed for internal PR cc @davanstrien

### DIFF
--- a/packages/tasks/src/library-ui-elements.ts
+++ b/packages/tasks/src/library-ui-elements.ts
@@ -72,7 +72,7 @@ model = BaseModel.from_pretrained("${model.id}")`,
 ];
 
 function get_base_diffusers_model(model: ModelData): string {
-	return model.cardData?.base_model ?? "fill-in-base-model";
+	return model.cardData?.base_model?.toString() ?? "fill-in-base-model";
 }
 
 const bertopic = (model: ModelData) => [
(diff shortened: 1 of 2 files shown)


Wauplin (approved):
Looks good.

I just realized `base_model` is not a supported field by default in `huggingface_hub.ModelCardData`. Will open a PR soon.

packages/tasks/src/library-ui-elements.ts
```diff
@@ -72,7 +72,7 @@ model = BaseModel.from_pretrained("${model.id}")`,
 ];
 
 function get_base_diffusers_model(model: ModelData): string {
-	return model.cardData?.base_model ?? "fill-in-base-model";
+	return model.cardData?.base_model?.toString() ?? "fill-in-base-model";
```
osanseviero, line 75: I guess it's ok to assume we only have 1 base model for diffusion models

osanseviero (approved):

packages/tasks/src/library-ui-elements.ts
julien-c: yes, i hesitated with taking the 1st element but 🤷

honestly we should have specified / enforced an array of string from the beginning, it's always kinda messy to do it after the fact

julien-c:
@coyotte508 any way to do a release? 🙏

julien-c:
can you update the npm requirement on my moon-landing PR when it's released? thanks a lot!

PR #427, opened 2023-12-19
Add MLX filter

osanseviero:


packages/tasks/src/library-ui-elements.ts
```diff
@@ -621,6 +626,12 @@ export const MODEL_LIBRARIES_UI_ELEMENTS: Partial<Record<ModelLibraryKey, Librar
 		docsUrl: "https://huggingface.co/docs/hub/keras",
 		snippets: keras,
 	},
+	mlx: {
+		btnLabel: "NeMo",
```
pcuenca, line 630:
```suggestion
		btnLabel: "MLX",
```

packages/tasks/src/library-ui-elements.ts
```diff
@@ -541,6 +541,11 @@ transcriptions = asr_model.transcribe(["file.wav"])`,
 
 const mlAgents = (model: ModelData) => [`mlagents-load-from-hf --repo-id="${model.id}" --local-dir="./downloads"`];
 
+
+const mlx = (model: ModelData) => {
+	return ["A code snippet will be here soon :) stay tuned!"];
```
pcuenca, line 546: Tricky as there are various inference scripts in the repo, let me see what we can do.

pcuenca (commented):
Thanks a lot!

packages/tasks/src/library-ui-elements.ts
osanseviero: There was some convo in Slack about having some automatic way to download models, but yes, it's quite tricky with the current state. We would need a `config.json` that contains the model type.

Vaibhavs10 (commented):
Agreed with @pcuenca! mlx-examples is a bunch of scripts and not quite a library like transformers so there's no standardisation in terms of their overall inference scripts.

I think for now it might be worth demonstrating how to download the checkpoints:
```
pip install huggingface_hub hf_transfer

# Download model
export HF_HUB_ENABLE_HF_TRANSFER=1
huggingface-cli download --local-dir Llama-2-7b-chat-mlx mlx-llama/Llama-2-7b-chat-mlx
```

packages/tasks/src/library-ui-elements.ts
Vaibhavs10: https://github.com/huggingface/huggingface.js/pull/427#pullrequestreview-1790560453

osanseviero:
My idea was to wait for a proper downstream integration before adding a code snippet, but given it's a low hanging fruit, I added a simple `huggingface_hub`-powered snippet

packages/tasks/src/library-ui-elements.ts
```diff
@@ -541,6 +541,13 @@ transcriptions = asr_model.transcribe(["file.wav"])`,
 
 const mlAgents = (model: ModelData) => [`mlagents-load-from-hf --repo-id="${model.id}" --local-dir="./downloads"`];
 
+const mlx = (model: ModelData) => [
+	`pip install huggingface_hub hf_transfer
+
+export HF_HUB_ENABLE_HF_TRANSFER=1
+huggingface-cli download --local-dir ${nameWithoutNamespace(model.id)} ${model.id}`,
```
pcuenca, line 548: Nice!

pcuenca (approved):

packages/tasks/src/library-ui-elements.ts
```diff
@@ -541,6 +541,13 @@ transcriptions = asr_model.transcribe(["file.wav"])`,
 
 const mlAgents = (model: ModelData) => [`mlagents-load-from-hf --repo-id="${model.id}" --local-dir="./downloads"`];
 
+const mlx = (model: ModelData) => [
+	`pip install huggingface_hub hf_transfer
+
+export HF_HUB_ENABLE_HF_TRANSFER=1
```
julien-c, line 547: personally i would remove the HF_HUB_ENABLE_HF_TRANSFER line (it's a bit experimental + not always faster depending on machine)

packages/tasks/src/library-ui-elements.ts
julien-c: if we wanted we could set a filename by convention, like `mlx_model.npz` for instance (like we've done for transformers for instance)

but might be irrelevant if mlx gets built-in safetensors support

PR #428, opened 2023-12-20
MLX: use casing from the GitHub repo's readme

julien-c:


pcuenca (approved):
Makes sense!

julien-c:
(looks better as a filter on hf.co/models too:)

Issue #430, opened 2023-12-23
Inference API "</> JSON Output" button triggers reload of webpage


Lauler:
**Bug description.**
The `</> JSON Output` button is currently broken on every model's inference API. When users click the button to expand the output, the output is first expanded (as intended), but then shortly after another event is triggered that reloads the entire webpage and moves on to the next "example" in the model's inference API. 

**Describe the expected behaviour**
Clicking on `</> JSON Output` should expand the JSON output, and not force reload the webpage. 

**Additional context**
Reproducible on any model with Inference API examples: https://huggingface.co/sentence-transformers/all-mpnet-base-v2?

PR #431, opened 2024-01-03
[Widget] Fix footer

mishig25:
Closes https://github.com/huggingface/huggingface.js/issues/430

Fixes two issues on widget footer.

1. Clicking on `JSON Output` btn was causing reload since it was a button inside `<form>`. In this case, unless a button is a submit button, button should have attribute `type="button"`
2. Minimize btn should minimize the widget

### DIFF
--- a/packages/widgets/src/lib/components/InferenceWidget/shared/WidgetFooter/WidgetFooter.svelte
+++ b/packages/widgets/src/lib/components/InferenceWidget/shared/WidgetFooter/WidgetFooter.svelte
@@ -30,6 +30,7 @@
 	{/if}
 	<button
 		class="ml-auto flex items-center"
+		type="button"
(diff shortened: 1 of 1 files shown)


packages/widgets/src/lib/components/InferenceWidget/shared/WidgetFooter/WidgetFooter.svelte
```diff
@@ -29,7 +30,8 @@
 	{/if}
 	<button
 		class="ml-auto flex items-center"
-		on:click|preventDefault={() => updateWidgetState(model.id, "isMaximized", true)}
+		type="button"
+		on:click|preventDefault={() => updateWidgetState(model.id, "isMaximized", !isMaximized)}
```
pcuenca, line 34: lol nice

Issue #408, opened 2023-12-08
Add signInWithHf() frontend helper


coyotte508:
Add a helper that uses the PKCE oauth flow to sign in with HF.

It should work only on the frontend. For backend, there are already examples including using `Auth.js` https://huggingface.co/docs/hub/spaces-oauth - and it's more framework-dependent.

The frontend one can just be a redirect / checking the url, eg:

```ts
// checks the url and redirect if needed
const info = await signInWithHf({clientId: ...});
```

PR #433, opened 2024-01-05
✨ Introduce oauthLogin function

coyotte508:
Fix https://github.com/huggingface/huggingface.js/issues/408

```ts
import { oauthLogin, oauthHandleRedirectIfPresent } from "@huggingface/hub";
import {HfInference} from "@huggingface/inference";

const oauthResult = await oauthHandleRedirectIfPresent();

if (!oauthResult) {
  // If the user is not logged in, redirect to the login page
  oauthLogin();
}

// You can use oauthResult.accessToken and oauthResult.userInfo
console.log(oauthResult);

const inference = new HfInference(oauthResult.accessToken);

await inference.textToImage({
  model: 'stabilityai/stable-diffusion-2',
  inputs: 'award winning high resolution photo of a giant tortoise/((ladybird)) hybrid, [trending on artstation]',
  parameters: {
    negative_prompt: 'blurry',
  }
})
```

Tested inside a space: https://huggingface.co/spaces/coyotte508/client-side-oauth

Mainly looking for reviews regarding the APIs / usability cc @xenova @radames @vvmnnnkv @jbilcke-hf @Wauplin .

I started with a single `oauthLogin` that was split into `oauthLogin` and `oauthHandleRedirect` with an extra `oauthHandleRedirectIfPresent` for convenience.

### DIFF
--- a/packages/hub/src/lib/oauth-login-url.ts
+++ b/packages/hub/src/lib/oauth-login-url.ts
@@ -87,9 +87,9 @@ export async function oauthLoginUrl(opts?: {
 		userinfo_endpoint: string;
 	} = await openidConfigRes.json();
 
-	const newNonce = crypto.randomUUID();
+	const newNonce = globalThis.crypto.randomUUID();
 	// Two random UUIDs concatenated together, because min length is 43 and max length is 128
-	const newCodeVerifier = crypto.randomUUID() + crypto.randomUUID();
+	const newCodeVerifier = globalThis.crypto.randomUUID() + globalThis.crypto.randomUUID();
 
 	localStorage.setItem("huggingface.co:oauth:nonce", newNonce);
 	localStorage.setItem("huggingface.co:oauth:code_verifier", newCodeVerifier);
--- a/packages/hub/src/utils/randomUUID.ts
+++ b/packages/hub/src/utils/randomUUID.ts
@@ -1,7 +1,0 @@
-export async function randomUUID(): Promise<string> {
-	if (globalThis.crypto) {
-		return globalThis.crypto.randomUUID();
-	} else {
-		return (await import("node:crypto")).randomUUID();
-	}
-}


Wauplin (commented):
Interesting read, thanks for the ping. DX-wise looks good to me. Just to be sure, is `oauthHandleRedirect` intended to be used or is it more like an internal helper and `oauthHandleRedirectIfPresent` is meant to be used more often?  


Small comment in the documentation, I would mention `accessTokenExpiresAt`. In the Python+Gradio integration I think that the token expiration is one of the blocking aspect (or at least one frustrating aspect). Having it explicitly documented should help IMO (if was not the case in the gradio integration).

packages/hub/src/lib/oauth-login.ts
```diff
@@ -0,0 +1,119 @@
...
+	}
+
+	const challenge = base64FromBytes(
+		new Uint8Array(await globalThis.crypto.subtle.digest("SHA-256", new TextEncoder().encode(newCodeVerifier)))
+	)
+		.replace(/[+]/g, "-")
+		.replace(/[/]/g, "_")
+		.replace(/=/g, "");
+
+	window.location.href = `${opendidConfig.authorization_endpoint}?${new URLSearchParams({
```
radames, line 110: would it make sense to have an option to use `window.open`  and optionally open the oauth login as an external window or tab?

radames:
> > Overall the API looks good! can we test it already?
> 
> https://huggingface.co/spaces/coyotte508/client-side-oauth uses the lib, checkout https://huggingface.co/spaces/coyotte508/client-side-oauth/blob/main/index.html

cool, testing this https://huggingface.co/spaces/coyotte508/client-side-oauth , for some reason if I refresh the page I have to re-login to hf, but accessing it via https://coyotte508-client-side-oauth.static.hf.space/index.html I just need to click sign-in and no need to re-login, do you know why?

PR #439, opened 2024-01-17
Change ML-Agents to Unity-ML-Agents + update library link

simoninithomas:
Context:
- Unity team wants to change the library name in Hugging Face from ML-Agents to Unity-ML-Agents.

### DIFF
--- a/packages/tasks/src/library-ui-elements.ts
+++ b/packages/tasks/src/library-ui-elements.ts
@@ -767,7 +767,7 @@ export const MODEL_LIBRARIES_UI_ELEMENTS: Partial<Record<ModelLibraryKey, Librar
 		docsUrl: "https://huggingface.co/docs/hub/stable-baselines3",
 		snippets: stableBaselines3,
 	},
-	"unity-ml-agents": {
+	"ml-agents": {
 		btnLabel: "unity-ml-agents",
 		repoName: "ml-agents",
 		repoUrl: "https://github.com/Unity-Technologies/ml-agents",
--- a/packages/tasks/src/library-ui-elements.ts
+++ b/packages/tasks/src/library-ui-elements.ts
@@ -770,7 +770,7 @@ export const MODEL_LIBRARIES_UI_ELEMENTS: Partial<Record<ModelLibraryKey, Librar
 	"unity-ml-agents": {
 		btnLabel: "unity-ml-agents",
 		repoName: "ml-agents",
-		repoUrl: "https://github.com/huggingface/ml-agents",
+		repoUrl: "https://github.com/Unity-Technologies/ml-agents",
 		docsUrl: "https://huggingface.co/docs/hub/ml-agents",
 		snippets: mlAgents,
 	},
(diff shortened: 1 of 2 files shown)


packages/tasks/src/model-libraries.ts
```diff
@@ -37,7 +37,7 @@ export enum ModelLibrary {
 	"stanza" = "Stanza",
 	"fasttext" = "fastText",
 	"stable-baselines3" = "Stable-Baselines3",
-	"ml-agents" = "ML-Agents",
+	"unity-ml-agents" = "Unity ML-Agents",
```
julien-c, line 40:
```suggestion
	"ml-agents" = "Unity ML-Agents",
```

i would suggest changing only the pretty name (the one that's actually displayed) otherwise 6,414 models will be broken

julien-c (changes requested):

packages/tasks/src/model-libraries.ts
osanseviero: +1 this is what was suggested in the internal channel yesterday with the ML Agent's team

simoninithomas: Thanks, Omar just told me the same thing. I had some doubts about whether it would retrieve them or not. I update it.

packages/tasks/src/library-ui-elements.ts
```diff
@@ -767,8 +767,8 @@ export const MODEL_LIBRARIES_UI_ELEMENTS: Partial<Record<ModelLibraryKey, Librar
 		docsUrl: "https://huggingface.co/docs/hub/stable-baselines3",
 		snippets: stableBaselines3,
 	},
-	"ml-agents": {
-		btnLabel: "ml-agents",
+	"unity-ml-agents": {
```
osanseviero, line 770: Same here

packages/tasks/src/library-ui-elements.ts
```diff
@@ -767,8 +767,8 @@ export const MODEL_LIBRARIES_UI_ELEMENTS: Partial<Record<ModelLibraryKey, Librar
...
+		btnLabel: "unity-ml-agents",
```
osanseviero, line 771: Can we change the link below (`repoUrl`) to the actual Unity repo rather than the fork? I think they merged the changes

simoninithomas: ✅

osanseviero (approved):
LGTM! Thanks!

julien-c (approved):

packages/tasks/src/library-ui-elements.ts
```diff
@@ -768,9 +768,9 @@ export const MODEL_LIBRARIES_UI_ELEMENTS: Partial<Record<ModelLibraryKey, Librar
 		snippets: stableBaselines3,
 	},
 	"ml-agents": {
-		btnLabel: "ml-agents",
+		btnLabel: "unity-ml-agents",
```
julien-c, line 771:
```suggestion
		btnLabel: "ml-agents",
```

to be honest i don't remember how btnLabel is used, i think it's just analytics, so i would keep the current one here too

julien-c:
feel free to merge!

PR #442, opened 2024-01-17
Add sentis library tag

pauldog:
To add the unity-sentis library tag to sentis models.

### DIFF
--- a/packages/tasks/src/library-ui-elements.ts
+++ b/packages/tasks/src/library-ui-elements.ts
@@ -783,9 +783,9 @@ export const MODEL_LIBRARIES_UI_ELEMENTS: Partial<Record<ModelLibraryKey, Librar
 		snippets: mlAgents,
 	},
 	"unity-sentis": {
-		btnLabel: "unity sentis",
+		btnLabel: "unity-sentis",
 		repoName: "unity-sentis",
-		repoUrl: "https://github.com/huggingface/unity-sentis",
+		repoUrl: "https://github.com/Unity-Technologies/sentis-samples",
 		snippets: sentis,
 	},
 	pythae: {
--- a/packages/tasks/src/library-ui-elements.ts
+++ b/packages/tasks/src/library-ui-elements.ts
@@ -774,6 +774,12 @@ export const MODEL_LIBRARIES_UI_ELEMENTS: Partial<Record<ModelLibraryKey, Librar
 		docsUrl: "https://huggingface.co/docs/hub/ml-agents",
 		snippets: mlAgents,
 	},
+	"unity-sentis": {
+		btnLabel: "unity sentis",
+		repoName: "unity-sentis",
+		repoUrl: "https://github.com/huggingface/unity-sentis",
+		snippets: sentis,
+	},
 	pythae: {
 		btnLabel: "pythae",
 		repoName: "pythae",
--- a/packages/tasks/src/library-ui-elements.ts
+++ b/packages/tasks/src/library-ui-elements.ts
@@ -542,8 +542,10 @@ transcriptions = asr_model.transcribe(["file.wav"])`,
 const mlAgents = (model: ModelData) => [`mlagents-load-from-hf --repo-id="${model.id}" --local-dir="./downloads"`];
 
 const sentis = (model: ModelData) => [
-	`Model model = ModelLoader.Load(Application.streamingAssetsPath + "/${model.id}");
+	`string modelName = "[Your model name here].sentis";
+	Model model = ModelLoader.Load(Application.streamingAssetsPath + "/" + modelName);
 IWorker engine = WorkerFactory.CreateWorker(BackendType.GPUCompute, model);
+// Please see provided C# file for more details
 `
 ];
 
--- a/packages/tasks/src/library-ui-elements.ts
+++ b/packages/tasks/src/library-ui-elements.ts
@@ -543,7 +543,7 @@ const mlAgents = (model: ModelData) => [`mlagents-load-from-hf --repo-id="${mode
 
 const sentis = (model: ModelData) => [
 	`string modelName = "[Your model name here].sentis";
-	Model model = ModelLoader.Load(Application.streamingAssetsPath + "/" + modelName);
+Model model = ModelLoader.Load(Application.streamingAssetsPath + "/" + modelName);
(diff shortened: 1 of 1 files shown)


packages/tasks/src/library-ui-elements.ts
```diff
@@ -774,6 +782,12 @@ export const MODEL_LIBRARIES_UI_ELEMENTS: Partial<Record<ModelLibraryKey, Librar
 		docsUrl: "https://huggingface.co/docs/hub/ml-agents",
 		snippets: mlAgents,
 	},
+	"unity-sentis": {
+		btnLabel: "unity sentis",
+		repoName: "unity-sentis",
+		repoUrl: "https://github.com/huggingface/unity-sentis",
```
pcuenca, line 788: This doesn't exist (the `ml-agents` example above is a fork, and it's going to be replaced in #439. Maybe you'd wnat to use https://github.com/Unity-Technologies/sentis-samples instead?

packages/tasks/src/library-ui-elements.ts
```diff
@@ -541,6 +541,14 @@ transcriptions = asr_model.transcribe(["file.wav"])`,
 
 const mlAgents = (model: ModelData) => [`mlagents-load-from-hf --repo-id="${model.id}" --local-dir="./downloads"`];
 
+const sentis = (model: ModelData) => [
+	`string modelName = "[Your model name here].sentis";
+	Model model = ModelLoader.Load(Application.streamingAssetsPath + "/" + modelName);
+IWorker engine = WorkerFactory.CreateWorker(BackendType.GPUCompute, model);
+// Please see provided C# file for more details
```
pcuenca, line 548: Where is the C# file provided? In the model repo?

packages/tasks/src/library-ui-elements.ts
```diff
@@ -774,6 +782,12 @@ export const MODEL_LIBRARIES_UI_ELEMENTS: Partial<Record<ModelLibraryKey, Librar
 		docsUrl: "https://huggingface.co/docs/hub/ml-agents",
 		snippets: mlAgents,
 	},
+	"unity-sentis": {
+		btnLabel: "unity sentis",
```
pcuenca, line 786: I'd maybe use `unity-sentis` for consistency with the tag. This is how it's displayed in the case of `ml-agents`:

Screenshot 2024-01-17 at 20 11 00

packages/tasks/src/library-ui-elements.ts
simoninithomas: Yes it can be interesting to add a Sentis documentation, as we did with ML-Agents: https://huggingface.co/docs/hub/ml-agents

packages/tasks/src/library-ui-elements.ts
pauldog: Hi yes, we provide it as a single file in the repo.

packages/tasks/src/library-ui-elements.ts
pauldog: Hi, yes I can change the github link to the samples link. (I wasn't sure if that github gets created automatically... apparently not) .

It would be good to have Sentis documentation (how do we do this?). We have our own documentation too: https://docs.unity3d.com/Packages/com.unity.sentis@latest

pcuenca: If you can provide the text that's awesome 🔥 We can add a page similar to the one @simoninithomas mentioned and link from there to your existing docs.

packages/tasks/src/library-ui-elements.ts
```diff
@@ -541,6 +541,14 @@ transcriptions = asr_model.transcribe(["file.wav"])`,
 
 const mlAgents = (model: ModelData) => [`mlagents-load-from-hf --repo-id="${model.id}" --local-dir="./downloads"`];
 
+const sentis = (model: ModelData) => [
+	`string modelName = "[Your model name here].sentis";
```
Wauplin, line 545: Out of curiosity, what is the "model name" referred to here? Can it be inferred from the repo_id on the Hub? Asking because if you can infer it from `ModelData`, that would make the snippet even better. In `ModelData` class you have access to the model id, the tags, the config, etc.

(if not possible, just ignore my comment :) )

pauldog: Hi, unfortunately this can't be inferred from the ModelData as in a lot of situations (e.g. Whisper or Stable Diffusion) it requires multiple *.sentis files. So I think the best we can do is give a general example here. We also provide a full C# code along side each model which most people will use instead of the snippet. 

I think there are two other option which might be better?: (1) Just remove this snippet entirely. (2) Do some fancy code which outputs the entire c# file as the snippet - which is probably overkill.

Wauplin: Thanks for the reply @pauldog! In the meantime I saw the discussion you've already had on slack about it so I'll defer to them for what's the best solution.

packages/tasks/src/library-ui-elements.ts
pauldog: I put together[ some text for the doc page](https://docs.google.com/document/d/1Er5PFY5oy9j_OWeQKjyhLrZZyiTWPM1HsRsQUwR_BdM/edit?usp=sharing)

Is this OK? Do we add it ourselves or do you add it? Thanks.

simoninithomas:
LGTM I let @osanseviero have the last word on it

packages/tasks/src/library-ui-elements.ts
```diff
@@ -541,6 +541,14 @@ transcriptions = asr_model.transcribe(["file.wav"])`,
...
+	Model model = ModelLoader.Load(Application.streamingAssetsPath + "/" + modelName);
```
osanseviero, line 546:
```suggestion
Model model = ModelLoader.Load(Application.streamingAssetsPath + "/" + modelName);
```

packages/tasks/src/library-ui-elements.ts
osanseviero: You can open a PR directly to `hub-docs` such as this file https://github.com/huggingface/hub-docs/blob/main/docs/hub/asteroid.md

osanseviero (approved):
LGTM after the suggested change

PR #447, opened 2024-01-19
🚨 Fix lint

coyotte508:
Introduced by #442


julien-c:
probably we could block merge (or at least publish) based on ci failures

osanseviero:
Thanks!

PR #451, opened 2024-01-21
📝 Move hub before agents in doc

coyotte508:
Due to `oauthLogin` recently added to `hub` & low activity on `agents`

cc @nsarrazin @osanseviero


julien-c:
sounds good to me, Hub could even be first IMO

nsarrazin (approved):
Sounds good to me!

PR #459, opened 2024-01-25
:recycle: Change library name: Adapter-Transformers -> Adapters

calpt:
Updated from `adapter-transformers` library to the newly released _Adapters_ library. Related to https://github.com/huggingface/hub-docs/pull/1184.

Open Questions:

I haven't updated the library key from "adapter-transformers" as this is also the tag used on the hub. I'm not sure what the best way to handle this is? Should the tag stay the same, can we support both "adapters" and "adapter-transformers" or is it possible to automatically rename all tags?

### DIFF
--- a/packages/tasks/src/library-ui-elements.ts
+++ b/packages/tasks/src/library-ui-elements.ts
@@ -35,11 +35,11 @@ function nameWithoutNamespace(modelId: string): string {
 
 //#region snippets
 
-const adapter_transformers = (model: ModelData) => [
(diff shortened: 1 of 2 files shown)


julien-c:
Do you really want to update the library key, or is it fine to keep the old one?

in case we want to change it, it'll be hackish, but i think we could code some kind of "redirect" for the library_name in our internal codebase 🤔

calpt:
I think we can keep the old library key for now

julien-c (approved):

PR #464, opened 2024-01-29
👷 Changes to build inference types

coyotte508:
Follow up to #449 

Review with whitespaces off


packages/tasks/scripts/inference-codegen.ts
```diff
@@ -139,54 +141,44 @@ async function postProcessOutput(path2generated: string, outputSpec: Record<stri
 	return;
 }
 
-async function main() {
```
SBrandeis, line 142: 😢

SBrandeis (approved):
thx!

PR #467, opened 2024-02-05
Fix trust_remote_code in snippets

osanseviero:
At the moment, we only add `trust_remote_code` in the cases where there's a custom class, but this will fail when there are other custom components, such as tokenizer/config/etc. Instead, we just use `trust_remote_code` if there's an `auto_map` in the `config`, which is the same condition to add the `custom_code` tag to repos. This fixes:

* https://huggingface.co/allenai/OLMo-7B
* https://huggingface.co/togethercomputer/m2-bert-80M-2k-retrieval


pcuenca (approved):
LGTM! (Note, however, that OLMo technically does not require `trust_remote_code` because it depends on the `ai2-olmo` package being installed).

SBrandeis (approved):

bitsnaps:
It would be nice if the API inference support this param as well?

packages/tasks/src/library-ui-elements.ts
```diff
@@ -422,7 +422,7 @@ const transformers = (model: ModelData) => {
 	if (!info) {
 		return [`# ⚠️ Type of model unknown`];
 	}
-	const remote_code_snippet = info.custom_class ? ", trust_remote_code=True" : "";
+	const remote_code_snippet = model.config?.auto_map ? ", trust_remote_code=True" : "";
```
julien-c, line 425:
```suggestion
	const remote_code_snippet = model.tags.includes(TAG_CUSTOM_CODE) ? ", trust_remote_code=True" : "";
```

or just this then? might be more expressive?

osanseviero: Ah yes, that's cleaner and delegates all to the Hub

osanseviero (commented):
Thanks everyone!

PR #472, opened 2024-02-07
🩹 [Tasks] Fix depth-estimation spec

SBrandeis:


packages/tasks/src/tasks/depth-estimation/inference.ts
```diff
@@ -17,19 +15,21 @@ export interface DepthEstimationInput {
...
+export interface DepthEstimationOutput {
+	/**
+	 * The predicted depth as an image
+	 */
+	depth?: unknown;
 	/**
-	 * When specified, limits the output to the top K most probable classes.
+	 * The predicted depth as a tensor
 	 */
-	topK?: number;
```
osanseviero, line 33: Shouldn't we keep `topk` in parameters?

osanseviero (approved):
👍 LGTM!

packages/tasks/src/tasks/depth-estimation/inference.ts
SBrandeis: It has been removed from the docstring in https://github.com/huggingface/transformers/pull/28733/files so I'm not sure

Ping @ydshieh who implemented the change in transformers

osanseviero: Ah yes! You're right hehe

Issue #469, opened 2024-02-07
Can't run on Cloudflare Workers


TranquilMarmot:
Following the example...

```ts
const hf = new HfInference('your access token')

await hf.summarization({
  model: 'facebook/bart-large-cnn',
  inputs:
    'The tower is 324 metres (1,063 ft) tall, about the same height as an 81-storey building, and the tallest structure in Paris. Its base is square, measuring 125 metres (410 ft) on each side. During its construction, the Eiffel Tower surpassed the Washington Monument to become the tallest man-made structure in the world, a title it held for 41 years until the Chrysler Building in New York City was finished in 1930.',
  parameters: {
    max_length: 100
  }
})
```

Running this in a Cloudflare Worker will fail with:

> Error: The 'credentials' field on 'RequestInitializerDict' is not implemented.

There is a closed issue in their repo (https://github.com/cloudflare/workers-sdk/issues/2514) about this error.

Maybe there's another way to pass the token that would play nicely with it 🤔

coyotte508:
You can pass a custom fetch function to `HfInference`:

```
const hf = new HfInference(env.HF_TOKEN, {
	fetch: (input, init) => {
		if (init) {
			// Remove "credentials" from the request
			const { credentials, ...rest } = init;
			void credentials;

			return fetch(input, rest);
		}

		return fetch(input);
	},
});
```

TranquilMarmot:
I tried that, but when calling the `HfInference` constructor the error still happens (even with an `undefined` token)

coyotte508:
The token is unrelated, `credentials` in the error is a fetch option.

Did you try 

```ts
const hf = new HfInference(undefined, {includeCredentials: false})
```

?

If it still errors, it's probably another part of the code or another hf.js module

TranquilMarmot:
Okay, that seems to be working. For reference, this is the code that I used:

```ts
const hf = new HfInference(undefined, {
includeCredentials: false,
fetch: (url, init) =>
    fetch(url, {
    ...init,
    headers: {
        ...init?.headers,
        Authorization: `Bearer ${env.HUGGINGFACE_TOKEN}`,
    },
    }) as unknown as any,
});
```

PR #475, opened 2024-02-07
💥 By default, do not set `credentials` in `fetch`

coyotte508:
Follow up to #248 

See #469

Do not pass `credentials` to `fetch` unless explicitly requrested.

cc @WalshyDev for visibility


coyotte508:
wow this was forgotten sorry

PR #477, opened 2024-02-08
Add image-text-to-text + multimodal fixes

osanseviero:
Tentatively

* We align better what multimodal is
* We add a new type for llava


packages/tasks/src/pipelines.ts
```diff
@@ -589,9 +589,15 @@ export const PIPELINE_DATA = {
 	},
 	"text-to-video": {
 		name: "Text-to-Video",
-		modality: "multimodal",
+		modality: "cv",
```
SBrandeis, line 592: Same comment here

osanseviero: Check the discussion in https://huggingface.slack.com/archives/C01N44FJDHT/p1706796001731059?thread_ts=1706792789.298009&cid=C01N44FJDHT

This model only has one type of model input (same as text-to-speech).

SBrandeis: That makes sense, thank you for the link!

SBrandeis (approved):

Wauplin (approved):

packages/tasks/src/pipelines.ts
```diff
@@ -430,7 +430,7 @@ export const PIPELINE_DATA = {
 				name: "Image Captioning",
 			},
 		],
-		modality: "multimodal",
+		modality: "cv",
```
amyeroberts, line 433: Should this be "nlp"? The other labels seem to match the output modality.

osanseviero: Hmm that would be a bit strange IMO. Both image generation and image captioning seem fundamentally computer vision tasks to me

Reeray:
Hi there, here is an image-text-to-text icon suggestion: 
Image-text-to-text

three other icon suggestions for better icon consistency:
1. Image-to-text, 
image-to-text

2. Image-to-video, 
image-to-video

3. text-to-video,
text-to-video

SBrandeis (approved):

PR #481, opened 2024-02-13
`speechbrain_interface` instead of `speechbrain?.interface`

pcuenca:
Models seem to use `speechbrain_interface`, for example: https://huggingface.co/TalTechNLP/voxlingua107-epaca-tdnn/blob/main/config.json?library=speechbrain#L2, https://huggingface.co/speechbrain/asr-crdnn-commonvoice-de/blob/main/config.json

### DIFF
--- a/packages/tasks/src/library-ui-elements.ts
+++ b/packages/tasks/src/library-ui-elements.ts
@@ -400,14 +400,14 @@ const speechBrainMethod = (speechbrainInterface: string) => {
 };
 
 const speechbrain = (model: ModelData) => {
-	const speechbrainInterface = model.config?.speechbrain?.interface;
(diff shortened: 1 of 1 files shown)


julien-c:
hmm yeah but take a look at the code in moon-landing; i think we rename/nest the config object in `model.config`

Is it not currently working?

pcuenca:
Oh, that would explain it. But no, it's not working.

Screenshot 2024-02-13 at 13 12 51

pcuenca:
Setting to `draft`, will debug in codespaces.

pcuenca:
Reference (internal): https://github.com/huggingface/moon-landing/blob/8634ab6d6098164176af69e9abef47a83fa66561/server/repo_types/Models.ts#L494

PR #482, opened 2024-02-14
Simpler support for new model libraries

julien-c:
close this [internal issue](https://github.com/huggingface/moon-landing/issues/8791)

### DIFF
--- a/packages/tasks/src/model-libraries.ts
+++ b/packages/tasks/src/model-libraries.ts
@@ -1,52 +1,333 @@
+import * as snippets from "./library-ui-elements";
+import type { ModelData } from "./model-data";
+
+/**
+ * Elements configurable by a model library.
+ */
+export interface LibraryUiElement {
+	/**
+	 * Pretty name of the library.
+	 * displayed in tags, and on the main
+	 * call-to-action button on the model page.
+	 */
+	btnLabel: string;
+	/**
+	 * Repo name of the library's (usually on GitHub) code repo
+	 */
+	repoName: string;
+	/**
+	 * URL to library's (usually on GitHub) code repo
+	 */
+	repoUrl: string;
+	/**
+	 * URL to library's docs
+	 */
+	docsUrl?: string;
+	/**
+	 * Code snippet(s) displayed on model page
+	 */
+	snippets?: (model: ModelData) => string[];
+	/**
+	 * should we display this library in hf.co/models filter
+	 * (only for popular libraries with > 100 models)
+	 */
+	filter?: boolean;
+}
+
 /**
  * Add your new library here.
  *
  * This is for modeling (= architectures) libraries, not for file formats (like ONNX, etc).
- * File formats live in an enum inside the internal codebase.
+ * (unlike libraries, file formats live in an enum inside the internal codebase.)
+ *
+ * Doc on how to add a library to the Hub:
+ *
+ * https://huggingface.co/docs/hub/models-adding-libraries
+ *
+ * /!\ IMPORTANT
+ *
+ * The key you choose is the tag your models have in their library_name on the Hub.
  */
-export enum ModelLibrary {
-	"adapter-transformers" = "Adapters",
-	"allennlp" = "allenNLP",
-	"asteroid" = "Asteroid",
-	"bertopic" = "BERTopic",
-	"diffusers" = "Diffusers",
-	"doctr" = "docTR",
-	"espnet" = "ESPnet",
-	"fairseq" = "Fairseq",
-	"flair" = "Flair",
-	"keras" = "Keras",
-	"k2" = "K2",
-	"mlx" = "MLX",
-	"nemo" = "NeMo",
-	"open_clip" = "OpenCLIP",
-	"paddlenlp" = "PaddleNLP",
-	"peft" = "PEFT",
-	"pyannote-audio" = "pyannote.audio",
-	"sample-factory" = "Sample Factory",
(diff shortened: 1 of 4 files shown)


packages/tasks/src/model-libraries.ts
```diff
@@ -1,52 +1,333 @@
...
+} satisfies Record<string, LibraryUiElement>;
+
+export type ModelLibraryKey = keyof typeof MODEL_LIBRARIES_UI_ELEMENTS;
 
-export const ALL_MODEL_LIBRARY_KEYS = Object.keys(ModelLibrary) as ModelLibraryKey[];
+export const ALL_MODEL_LIBRARY_KEYS = Object.keys(MODEL_LIBRARIES_UI_ELEMENTS) as ModelLibraryKey[];
 
-export const ALL_DISPLAY_MODEL_LIBRARY_KEYS = ALL_MODEL_LIBRARY_KEYS.filter(
-	(k) => !["doctr", "k2", "mindspore", "tensorflowtts"].includes(k)
-);
```
julien-c, line 52: there was already this @osanseviero but i wrapped it inside the main dict

Note that i used an opt-in "filter: true" which seems friendlier to me

The elastic queries i will add in a second PR

packages/tasks/src/model-libraries.ts
```diff
@@ -1,52 +1,333 @@
...
+		repoUrl: "https://github.com/huggingface/huggingface_sb3",
+		docsUrl: "https://huggingface.co/docs/hub/stable-baselines3",
+		snippets: snippets.stableBaselines3,
+		filter: true,
+	},
+	mindspore: {
+		btnLabel: "MindSpore",
+		repoName: "mindspore",
+		repoUrl: "https://github.com/mindspore-ai/mindspore",
+	},
```
julien-c, line 296: added this one + k2 and doctr to the main dict, as they were not there

osanseviero: Thanks!

packages/tasks/src/model-libraries.ts
```diff
@@ -1,52 +1,333 @@
...
+		snippets: snippets.adapters,
+		filter: true,
+	},
+	allennlp: {
+		btnLabel: "AllenNLP",
+		repoName: "AllenNLP",
+		repoUrl: "https://github.com/allenai/allennlp",
+		docsUrl: "https://huggingface.co/docs/hub/allennlp",
+		snippets: snippets.allennlp,
+		filter: true,
```
osanseviero, line 67: Let's remove this one too as it's archived 

```suggestion
```

packages/tasks/src/model-libraries.ts
osanseviero: Awesome, thanks a lot!

osanseviero: Was `ALL_MODEL_LIBRARY_KEYS` being used? They were still shown :thinking:

osanseviero (approved):
Thanks!

coyotte508 (approved):

Wauplin (approved):
PR looks good to me but I don't think it closes the issue. In https://github.com/huggingface/moon-landing/issues/8791#issuecomment-1912339361 @osanseviero suggested to:
1. be able to remove the library from the filter list (done in this PR)
2. move the Elastic queries to have everything in a single file (and public) => still to be done or it's already decided that we don't do it?

packages/tasks/src/model-libraries.ts
```diff
@@ -1,52 +1,333 @@
+import * as snippets from "./library-ui-elements";
```
SBrandeis, line 1: Let's rename the file to make it more explicit it only contains snippets maybe

SBrandeis (approved):

packages/tasks/src/model-libraries.ts
julien-c: yes, looks like `ALL_DISPLAY_MODEL_LIBRARY_KEYS` (the filtered one) was only used in the suggestion dropdown in the GUI editor:

image

julien-c:
@Wauplin https://github.com/huggingface/huggingface.js/pull/482#discussion_r1489188050

i propose to add the elastic queries in a subsequent PR

packages/tasks/src/model-libraries.ts
julien-c: even if it leads to 146 models?

image

osanseviero: Yeah, with the library being archived I don't think it makes much sense anymore. It's 77 public models

packages/tasks/src/model-libraries.ts
osanseviero: Ok, we need a follow up PR so `ALL_DISPLAY_MODEL_LIBRARY_KEYS` is also used to control the filters at left of hf.co/models

packages/tasks/src/model-libraries.ts
julien-c: will do a more comprehensive cleanup in a later PR

packages/tasks/src/model-libraries.ts
julien-c: a4c7c52a9d179252ed7d5c2657a50fe11ed6ea2c

julien-c:
(cc @lhoestq also for dataset libraries)

julien-c:
Second part is in #485, please review 🙏

PR #491, opened 2024-02-20
Fix parseShardedIndex validation error, cap maximum sharded json file

Kakulukian:


coyotte508 (approved):
Slightly more efficient alternative is to rely on the content-length header of the response

Since the fetch is in two stages:

- get the headers
- download the content

So if we check the content-length between the two stages, we don't have to download any data (and don't need to rely on the range)

packages/hub/src/lib/parse-safetensors-metadata.ts
```diff
@@ -119,14 +119,20 @@ async function parseShardedIndex(
...
+		range: [0, 10_000_000],
 	});
 
 	if (!indexResp) {
 		throw new SafetensorParseError(`Failed to parse file ${path}: failed to fetch safetensors index.`);
 	}
 
 	// no validation for now, we assume it's a valid IndexJson.
-	const index: SafetensorsIndexJson = await indexResp.json();
+	let index: SafetensorsIndexJson;
```
julien-c, line 130: remove the comment one line above then?

Kakulukian: It's still true we don't enforce any validation on the payload

PR #492, opened 2024-02-21
🦺 Make sure latest version of jina is published when pubishing widgets

coyotte508:
cc @Wauplin 

@SBrandeis will probably need to do the same with inference if/when adding it as a dep to widgets (we could probably streamline it with a matrix or separate github action that we can call)


SBrandeis (approved):

xenova (approved):
LGTM! (same workflow for `@huggingface/tasks`)

Issue #497, opened 2024-02-22
Hf-Hub - listModels filter models by library type or custom tag


phudtran:
Hi, I've been using the listModels from huggingface/hub to search for and download models, and was wondering if there is a way to filter by Libraries, or "Other" custom tags like on [huggingface.co ](huggingface.co/)

<img width="229" alt="image" src="https://github.com/huggingface/huggingface.js/assets/22720475/094f036f-825d-4ba2-8f4d-50f4f47d106a">

<img width="229" alt="image" src="https://github.com/huggingface/huggingface.js/assets/22720475/cd82a49a-8972-466c-868a-7ae40950e291">

coyotte508:
Fixed in `@huggingface/hub@0.14.0`

You can do `listModels({search: {tags: ["gguf", "lora]}})` to search models with both gguf & lora.

PR #498, opened 2024-02-22
✨ Query models / datasets / spaces by tags

coyotte508:
Fix https://github.com/huggingface/huggingface.js/issues/497

## Query repos by tags

Eg for gguf model, `listModels({search: {tags: ["gguf"]}})`

For indonesian models, `listModels({search: {tags: ["id"]}})`

For models stored in the EU, `listModels({search: {tags: ["region:eu"]}})`

And so on.

## Request extra fields

The `additionalFields` param, previously only available for `listSpaces`, is also available for `listModels` and `listDatasets`.

It's also now strongly typed.

So if you want to get the tags of the models you're listing:

```ts
for await (const model of listModels({additionalFields: ["tags"]})) {
  console.log(model.tags);
}
```

## Limit the number of models requested

You can specify the number of models to fetch, eg:

```ts
for await (const model of listModels({limit: 2})) {
  console.log(model.tags);
}
```

This will only fetch two models


packages/hub/src/lib/list-datasets.ts
```diff
@@ -17,24 +44,35 @@ export interface DatasetEntry {
 	updatedAt: Date;
 }
 
-export async function* listDatasets(params?: {
+export async function* listDatasets<
+	const T extends Exclude<(typeof EXPANDABLE_KEYS)[number], (typeof EXPAND_KEYS)[number]> = never,
+>(params?: {
 	search?: {
 		owner?: string;
+		tags?: string[];
```
julien-c, line 52: or `filter` to stay closer to the actual param name?

packages/hub/src/types/public.d.ts
```diff
@@ -79,3 +79,73 @@ export interface SpaceResourceConfig {
 	throttled?: boolean;
 	is_custom?: boolean;
 }
+
+export type License =
```
julien-c, line 83: will get stale over time but i guess it's no big deal?

coyotte508: Ideally I'd like to move it to `@huggingface/tasks` eventually

packages/hub/src/lib/list-datasets.ts
coyotte508: It's `filter` for the API, but what users see is more something like https://huggingface.co/models?language=ja&license=license:other&sort=trending where there's `language`, `library`, `license`, `other` etc.

(See original issue #497)

Since the API returns the `tags` field in the `listModels`, etc calls, to me it's more user-friendly to also request on `tags`.

Note that there are also other search params such as `search: {owner: ...}`. If we added `filter` it would be more general IMO, like `owner:xxx`, etc, like a query language

packages/hub/src/types/public.d.ts
julien-c: i'm not personally in favor of moving more stuff to `@huggingface` tbh – at least with current tooling (it's very painful to develop against)

PR #500, opened 2024-02-23
update widgets-publish workflow

SBrandeis:


coyotte508 (approved):
ok, we could also use [matrixes](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs) to reduce LOCs :)

(now that we have 3 identical blocks)

PR #507, opened 2024-02-24
✨ Allow search on model name

coyotte508:
For https://github.com/huggingface/huggingface.js/discussions/506

```ts
		for await (const entry of listModels({
			search: { query: "t5" },
			limit: 10,
		})) {
			count++;
			expect(entry.name).to.include("t5");
		}
```

cc @julien-c @Wauplin if you have ideas about the parameter name. It's `search.query`, maybe `search.name` or `search.nameMatch`?

or `search.search`...

### DIFF
--- a/packages/hub/src/lib/list-datasets.ts
+++ b/packages/hub/src/lib/list-datasets.ts
@@ -73,7 +73,7 @@ export async function* listDatasets<
 		...Object.entries({
 			limit: String(Math.min(totalToFetch, 500)),
 			...(params?.search?.owner ? { author: params.search.owner } : undefined),
-			...(params?.search?.query ? { query: params.search.query } : undefined),
+			...(params?.search?.query ? { search: params.search.query } : undefined),
 		}),
 		...(params?.search?.tags?.map((tag) => ["filter", tag]) ?? []),
 		...EXPAND_KEYS.map((val) => ["expand", val] satisfies [string, string]),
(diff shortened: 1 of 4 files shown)


julien-c (commented):
`search.query` is fine i guess 👍 

note again that i would have called the toplevel object `filter` and inside of it `filter.author` (vs. `owner`) etc for consistency with huggingface_hub

packages/hub/src/lib/list-datasets.ts
```diff
@@ -69,6 +73,7 @@ export async function* listDatasets<
 		...Object.entries({
 			limit: String(Math.min(totalToFetch, 500)),
 			...(params?.search?.owner ? { author: params.search.owner } : undefined),
+			...(params?.search?.query ? { query: params.search.query } : undefined),
```
julien-c, line 76:
```suggestion
			...(params?.search?.query ? { search: params.search.query } : undefined),
```
this is search here no?

packages/hub/src/lib/list-spaces.ts
```diff
@@ -56,7 +60,11 @@ export async function* listSpaces<
 }): AsyncGenerator<SpaceEntry> {
 	checkCredentials(params?.credentials);
 	const search = new URLSearchParams([
-		...Object.entries({ limit: "500", ...(params?.search?.owner ? { author: params.search.owner } : undefined) }),
+		...Object.entries({
+			limit: "500",
+			...(params?.search?.owner ? { author: params.search.owner } : undefined),
+			...(params?.search?.query ? { query: params.search.query } : undefined),
```
julien-c, line 66:
```suggestion
			...(params?.search?.query ? { search: params.search.query } : undefined),
```
same here

PR #512, opened 2024-02-27
💚 Fix sync of inference widgets demo

coyotte508:
https://huggingface.co/spaces/huggingfacejs/inference-widgets is broken , due to https://github.com/huggingface/huggingface.js/issues/486


SBrandeis (approved):
Thx

PR #517, opened 2024-02-29
`model.config`: Align with breaking changes in the Hub

julien-c:
following (internal) https://github.com/huggingface/moon-landing/pull/8978

### DIFF
--- a/packages/tasks/src/model-data.ts
+++ b/packages/tasks/src/model-data.ts
@@ -42,6 +42,21 @@ export interface ModelData {
 	 */
 	config?: {
 		architectures?: string[];
+		/**
+		 * Dict of AutoModel or Auto… class name to local import path in the repo
+		 */
+		auto_map?: {
+			/**
+			 * String Property
+			 */
+			[x: string]: string;
+		};
+		model_type?: string;
+		quantization_config?: {
+			bits?: number;
+			load_in_4bit?: boolean;
+			load_in_8bit?: boolean;
+		};
 		tokenizer_config?: TokenizerConfig;
 		adapter_transformers?: {
 			model_name?: string;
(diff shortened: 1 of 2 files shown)


SBrandeis (approved):

packages/tasks/src/model-data.ts
```diff
@@ -40,21 +40,29 @@ export interface ModelData {
...
+		adapter_transformers?: {
+			model_name?: string;
+			model_class?: string;
+		};
+		diffusers?: {
+			_class_name?: string;
+		};
 		sklearn?: {
-			filename?: string;
+			model?: {
```
osanseviero, line 54: Why this change? Isn't it `filename` in the Hub?

julien-c: nope not after https://github.com/huggingface/moon-landing/pull/8978 where we enforce that we always use the prop names from the actual JSON files (will make things simpler in the long run)

packages/tasks/src/model-data.ts
```diff
@@ -40,21 +40,29 @@ export interface ModelData {
...
+				file?: string;
+			};
 			model_format?: string;
 		};
 		speechbrain?: {
-			interface?: string;
+			speechbrain_interface?: string;
```
osanseviero, line 60: Btw I just realized the `speechbrain` API needs some additional fields https://github.com/huggingface/api-inference-community/blob/main/docker_images/speechbrain/app/common.py#L39

julien-c: `vocoder_model_id` is also in there, i'll add it to types

julien-c: see 14ffcb7b740378e16a3a2bdf93325235528f70d7

osanseviero: Ah nice, nothing to worry then!

packages/tasks/src/model-data.ts
osanseviero: Related PR https://github.com/huggingface/moon-landing/pull/9104

PR #525, opened 2024-03-03
[`@huggingface/jinja`] Add more features and improve whitespace handling

xenova:
This PR adds support for:
- [Tests](https://jinja.palletsprojects.com/en/3.0.x/templates/#tests) (e.g., `{{ var is defined }}`)
- Improved handling and parsing of templates with whitespace (including `{%-`, `-%}`, `{{-`, and `-}}` control sequences)
- Add support for `namespace` construct
- Add additional loop info to body of for loop (e.g., `previtem`)
- Additional end-to-end unit tests with popular chat templates on the HF Hub ([link](https://huggingface.co/models?other=conversational)). An external script is being run to guarantee compatibility with all models that have at least 10 downloads (~3k tested).


Rocketknight1:
The automated test for popular templates is amazing! We may also need to do the same thing for `minijinja` in TGI at some point, which would have turned up their `namespace` issue sooner.

PR #529, opened 2024-03-04
🏷️ Less restrictive typing on model data passed to snippets

coyotte508:
https://github.com/huggingface/moon-landing/pull/9115 fails because of a typing conlict on an uused field (`tokenizer_config`)

This PR makes the typing restrictive, no need to pass a full model data, only a specific fields


julien-c:
where is the typing incompatible?

coyotte508:
> where is the typing incompatible?

w/ the PR, on `AddedToken` (required vs non-required in `__type`)

SBrandeis (approved):

julien-c:
ok, why not then 🫡

PR #531, opened 2024-03-04
fix SpecialTokenMap type

SBrandeis:


coyotte508 (approved):
See also https://github.com/huggingface/huggingface.js/pull/529

SBrandeis:
Ah nice - let's only merge #529 then maybe?

julien-c (approved):

Wauplin (approved):

PR #532, opened 2024-03-04
🩹 Fix focus issue with reset btn in `ConversationalWidget`

SBrandeis:
Currently, hitting `enter` to submit a new message triggers the `reset` button when visible


coyotte508 (approved):
Note that `type=reset` exists for buttons - not sure if appropriate here

PR #536, opened 2024-03-06
🚀 `@huggingface/jinja` 100% chat template coverage.

xenova:
With this PR, I'm pleased to announce that `@huggingface/jinja` is now compatible with every single _valid_[^1] chat template on the HF Hub (as of 2024/03/05), including the very complex function calling ones (like [fireworks-ai/firefunction-v1](https://huggingface.co/fireworks-ai/firefunction-v1))! 🥳, and we match the python output exactly in each case! Interestingly, of the ~11k public conversational models, there were only ~250 unique templates. This will also ensure all models tagged with `conversational` will have working conversational widgets!

[^1]: Only 1 invalid chat template was discovered (containing invalid quote characters), and a PR has been made for it [here](https://huggingface.co/YokaiKoibito/llama2_70b_chat_uncensored-fp16/discussions/1#65e8686f034b83aeb8d40528).

Of course, future models may introduce more complex chat templates, and we'll continue to add support for them!

cc @Wauplin @Rocketknight1 @osanseviero


drbh:
🚀 woo great job!

Rocketknight1:
Amazing!

PR #540, opened 2024-03-08
a GGUF parser that works on remotely hosted files (over HTTP range requests)

julien-c:
GGUF is the new single-file weights format that's been taking the Hub by storm...

image

## Spec

https://github.com/ggerganov/ggml/blob/master/docs/gguf.md

Reference implementation (Python): https://github.com/ggerganov/llama.cpp/blob/master/gguf-py/gguf/gguf_reader.py

## Acknowledgements & Inspirations

- https://github.com/hyparam/hyllama by @platypii (MIT license)
- https://github.com/ahoylabs/gguf.js by @biw @dkogut1996 @spencekim (MIT license)
- might also be of interest to: @yagil

### DIFF
--- a/.github/workflows/agents-publish.yml
+++ b/.github/workflows/agents-publish.yml
@@ -33,7 +33,7 @@ jobs:
       - run: corepack enable
       - uses: actions/setup-node@v3
         with:
-          node-version: "18"
+          node-version: "20"
           cache: "pnpm"
           cache-dependency-path: |
             packages/agents/pnpm-lock.yaml
@@ -59,7 +59,7 @@ jobs:
       # hack - reuse actions/setup-node@v3 just to set a new registry
       - uses: actions/setup-node@v3
         with:
-          node-version: "18"
+          node-version: "20"
           registry-url: "https://npm.pkg.github.com"
       - run: pnpm publish --no-git-checks .
         env:
--- a/packages/gguf/src/gguf.spec.ts
+++ b/packages/gguf/src/gguf.spec.ts
@@ -1,0 +1,139 @@
+import { describe, expect, it } from "vitest";
+import { GGMLQuantizationType, gguf } from "./gguf";
+
+const URL_LLAMA = "https://huggingface.co/TheBloke/Llama-2-7B-Chat-GGUF/resolve/main/llama-2-7b-chat.Q2_K.gguf";
+const URL_MISTRAL_7B =
+	"https://huggingface.co/TheBloke/Mistral-7B-Instruct-v0.2-GGUF/resolve/main/mistral-7b-instruct-v0.2.Q5_K_M.gguf";
+const URL_GEMMA_2B = "https://huggingface.co/lmstudio-ai/gemma-2b-it-GGUF/resolve/main/gemma-2b-it-q4_k_m.gguf";
+
+describe("gguf", () => {
+	it("should parse a llama2 7b", async () => {
+		const { metadata, tensorInfos } = await gguf(URL_LLAMA);
+
+		/// metadata
+
+		expect(metadata).toMatchObject({
+			version: 2,
+			tensor_count: 291n,
+			kv_count: 19n,
+			"general.architecture": "llama",
+			"general.file_type": 10,
+			"general.name": "LLaMA v2",
+			"general.quantization_version": 2,
+			"llama.attention.head_count": 32,
+			"llama.attention.head_count_kv": 32,
+			"llama.attention.layer_norm_rms_epsilon": 9.999999974752427e-7,
+			"llama.block_count": 32,
+			"llama.context_length": 4096,
+			"llama.embedding_length": 4096,
+			"llama.feed_forward_length": 11008,
+			"llama.rope.dimension_count": 128,
+		});
+
+		const tokens = metadata["tokenizer.ggml.tokens"];
+		if (!Array.isArray(tokens)) {
+			throw new Error();
+		}
+		expect(tokens.slice(0, 10)).toEqual([
+			"<unk>",
+			"<s>",
+			"</s>",
+			"<0x00>",
+			"<0x01>",
+			"<0x02>",
+			"<0x03>",
+			"<0x04>",
+			"<0x05>",
+			"<0x06>",
+		]);
+
+		/// Tensor infos
+
+		expect(tensorInfos.length).toEqual(291);
+		expect(tensorInfos[0]).toMatchObject({
+			name: "token_embd.weight",
+			shape: [4096n, 32000n],
+			type: GGMLQuantizationType.Q2_K,
+		});
+		expect(tensorInfos[tensorInfos.length - 1]).toMatchObject({
+			name: "output_norm.weight",
+			shape: [4096n],
+			type: GGMLQuantizationType.F32,
+		});
+	});
+
+	it("should parse a mistral 7b", async () => {
+		const { metadata, tensorInfos } = await gguf(URL_MISTRAL_7B);
+
+		/// metadata
+
+		expect(metadata).toMatchObject({
+			version: 3,
+			tensor_count: 291n,
+			kv_count: 24n,
+			"general.architecture": "llama",
+			"general.file_type": 17,
+			"general.name": "mistralai_mistral-7b-instruct-v0.2",
+			"general.quantization_version": 2,
+			"llama.attention.head_count": 32,
+			"llama.attention.head_count_kv": 8,
+			"llama.attention.layer_norm_rms_epsilon": 0.000009999999747378752,
+			"llama.block_count": 32,
+			"llama.context_length": 32768,
+			"llama.embedding_length": 4096,
+			"llama.feed_forward_length": 14336,
+			"llama.rope.dimension_count": 128,
+		});
+
+		/// Tensor infos
+
+		expect(tensorInfos.length).toEqual(291);
+		expect(tensorInfos[0]).toMatchObject({
+			name: "token_embd.weight",
+			shape: [4096n, 32000n],
+			type: GGMLQuantizationType.Q5_K,
+		});
+		expect(tensorInfos[tensorInfos.length - 1]).toMatchObject({
+			name: "output.weight",
+			shape: [4096n, 32000n],
+			type: GGMLQuantizationType.Q6_K,
+		});
+	});
+
+	it("should parse a gemma 2b", async () => {
+		const { metadata, tensorInfos } = await gguf(URL_GEMMA_2B);
+
+		/// metadata
+
+		expect(metadata).toMatchObject({
+			version: 3,
+			tensor_count: 164n,
+			kv_count: 21n,
+			"general.architecture": "gemma",
+			"general.file_type": GGMLQuantizationType.Q8_K, // 15
+			"general.name": "gemma-2b-it",
+			"general.quantization_version": 2,
+			"gemma.attention.head_count": 8,
+			"gemma.attention.head_count_kv": 1,
+			"gemma.attention.layer_norm_rms_epsilon": 9.999999974752427e-7,
+			"gemma.block_count": 18,
+			"gemma.context_length": 8192,
+			"gemma.embedding_length": 2048,
+			"gemma.feed_forward_length": 16384,
+		});
+
+		/// Tensor infos
+
+		expect(tensorInfos.length).toEqual(164);
+		expect(tensorInfos[0]).toMatchObject({
+			name: "token_embd.weight",
+			shape: [2048n, 256128n],
+			type: GGMLQuantizationType.Q4_K,
+		});
+		expect(tensorInfos[tensorInfos.length - 1]).toMatchObject({
+			name: "blk.9.ffn_norm.weight",
+			shape: [2048n],
+			type: GGMLQuantizationType.F32,
+		});
+	});
+});
--- a/packages/gguf/src/gguf.spec.ts
+++ b/packages/gguf/src/gguf.spec.ts
@@ -6,6 +6,9 @@ const URL_MISTRAL_7B =
 	"https://huggingface.co/TheBloke/Mistral-7B-Instruct-v0.2-GGUF/resolve/main/mistral-7b-instruct-v0.2.Q5_K_M.gguf";
 const URL_GEMMA_2B = "https://huggingface.co/lmstudio-ai/gemma-2b-it-GGUF/resolve/main/gemma-2b-it-q4_k_m.gguf";
 
+const URL_BIG_ENDIAN =
+	"https://huggingface.co/ggml-org/models/resolve/main/bert-bge-small/ggml-model-f16-big-endian.gguf";
+
 describe("gguf", () => {
 	it("should parse a llama2 7b", async () => {
 		const { metadata, tensorInfos } = await gguf(URL_LLAMA);
@@ -13,6 +16,7 @@ describe("gguf", () => {
 		/// metadata
 
 		expect(metadata).toMatchObject({
+			// partial list, do not exhaustively list (tokenizer is quite big for instance)
 			version: 2,
 			tensor_count: 291n,
 			kv_count: 19n,
@@ -48,6 +52,7 @@ describe("gguf", () => {
 		]);
 
 		/// Tensor infos
+		/// By convention we test the first and last tensor.
 
 		expect(tensorInfos.length).toEqual(291);
 		expect(tensorInfos[0]).toMatchObject({
@@ -134,6 +139,43 @@ describe("gguf", () => {
 			name: "blk.9.ffn_norm.weight",
 			shape: [2048n],
 			dtype: GGMLQuantizationType.F32,
+		});
+	});
+
+	it("should parse a big-endian file", async () => {
+		const { metadata, tensorInfos } = await gguf(URL_BIG_ENDIAN);
+
+		/// metadata
+
+		expect(metadata).toMatchObject({
+			version: 3,
+			tensor_count: 197n,
+			kv_count: 23n,
+			"general.architecture": "bert",
+			"general.file_type": GGMLQuantizationType.F16,
+			"general.name": "bge-small-en-v1.5",
+			"bert.attention.causal": false,
+			"bert.attention.head_count": 12,
+			"bert.attention.layer_norm_epsilon": 9.999999960041972e-13,
+			"bert.block_count": 12,
+			"bert.context_length": 512,
+			"bert.embedding_length": 384,
+			"bert.feed_forward_length": 1536,
+			"bert.pooling_type": 2,
+		});
+
+		/// Tensor infos
+
+		expect(tensorInfos.length).toEqual(197);
+		expect(tensorInfos[0]).toMatchObject({
+			name: "token_embd_norm.bias",
+			shape: [384n],
+			dtype: GGMLQuantizationType.F32,
+		});
+		expect(tensorInfos[tensorInfos.length - 1]).toMatchObject({
+			name: "blk.9.ffn_down.weight",
+			shape: [1536n, 384n],
+			dtype: GGMLQuantizationType.F16,
 		});
 	});
 });
--- a/packages/gguf/src/gguf.spec.ts
+++ b/packages/gguf/src/gguf.spec.ts
@@ -53,12 +53,12 @@ describe("gguf", () => {
 		expect(tensorInfos[0]).toMatchObject({
 			name: "token_embd.weight",
 			shape: [4096n, 32000n],
-			type: GGMLQuantizationType.Q2_K,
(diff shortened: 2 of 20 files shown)


packages/gguf/src/gguf.spec.ts
```diff
@@ -0,0 +1,139 @@
...
+			"general.name": "LLaMA v2",
+			"general.quantization_version": 2,
+			"llama.attention.head_count": 32,
+			"llama.attention.head_count_kv": 32,
+			"llama.attention.layer_norm_rms_epsilon": 9.999999974752427e-7,
+			"llama.block_count": 32,
+			"llama.context_length": 4096,
+			"llama.embedding_length": 4096,
+			"llama.feed_forward_length": 11008,
+			"llama.rope.dimension_count": 128,
```
julien-c, line 30: partial list, did not exhaustively list (tokenizer is quite big for instance)

packages/gguf/src/gguf.spec.ts
```diff
@@ -0,0 +1,139 @@
...
+		expect(tensorInfos[0]).toMatchObject({
+			name: "token_embd.weight",
+			shape: [4096n, 32000n],
+			type: GGMLQuantizationType.Q5_K,
+		});
+		expect(tensorInfos[tensorInfos.length - 1]).toMatchObject({
+			name: "output.weight",
+			shape: [4096n, 32000n],
+			type: GGMLQuantizationType.Q6_K,
+		});
```
julien-c, line 100: each time i'm checking the first and the last tensors

.github/workflows/agents-publish.yml
```diff
@@ -33,7 +33,7 @@ jobs:
       - run: corepack enable
       - uses: actions/setup-node@v3
         with:
-          node-version: "18"
+          node-version: "20"
```
julien-c, line 36: ArrayBuffer resizing is only in Node 20+

packages/gguf/src/gguf.ts
```diff
@@ -0,0 +1,275 @@
...
+function isGGUFValueType(n: number): n is GGUFValueType {
+	return typeof GGUFValueType[n] === "string";
+}
+
+const HTTP_CHUNK_SIZE = 2 * 10 ** 6; /// 2MB
+const HTTP_DATA_LEEWAY = 5 * 10 ** 5; /// 500kb
+
+/**
+ * Internal stateful instance to fetch ranges of HTTP data when needed
+ */
```
julien-c, line 58: GGUF is more complex to parse remotely than safetensors because in GGUF, we don't know the header size before parsing the whole header.

Hence i'm doing requests for chunks of file ranges, on demand.

biw:
Great work on the library and thanks for the acknowledgement @julien-c! I'm excited to see GGUF get some more support. 

One thing that we've seen a lot is a lack of optional general metadata (`general.author`, `general.url`, `general.description`, `general.license`, `general.source.url`, & `general.source.huggingface.repository`) included in most GGUF files. Do you have any thoughts on adding that as a first-class citizen (typescript types) to the PR? I'm sure having that in `huggingface.js` would help bring attention to model authors that the fields are there (it's very useful to have for model frontends wanting to support arbitrary models)

packages/gguf/src/gguf.ts
```diff
@@ -0,0 +1,275 @@
+type MetadataBaseValue = string | number | bigint | boolean;
+type MetadataValue = MetadataBaseValue | MetadataBaseValue[] | MetadataValue[]; /// recursive as arrays can be nested.
+
+type Version = 1 | 2 | 3;
+const isVersion = (version: number): version is Version => version === 1 || version === 2 || version === 3;
```
mishig25, line 5:
```suggestion
const VALID_VERSIONS: number[] = [1, 2, 3];
type Version = (typeof VALID_VERSIONS)[number];
const isVersion = (version: number): version is Version => VALID_VERSIONS.includes(version);
```

how about this?

packages/gguf/src/gguf.ts
```diff
@@ -0,0 +1,275 @@
...
+	FLOAT32 = 6,
+	BOOL = 7,
+	STRING = 8,
+	ARRAY = 9,
+	UINT64 = 10,
+	INT64 = 11,
+	FLOAT64 = 12,
+}
+function isGGUFValueType(n: number): n is GGUFValueType {
+	return typeof GGUFValueType[n] === "string";
```
mishig25, line 50:
```suggestion
	return !!GGUFValueType[n];
```

this can be just this

packages/gguf/src/gguf.ts
```diff
@@ -0,0 +1,275 @@
...
+						Range: `bytes=${range[0]}-${range[1]}`,
+					},
+				})
+			).arrayBuffer()
+		);
+		/// TODO(fix typing)
+		// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+		// @ts-ignore
+		this.buffer.resize((this.chunk + 1) * HTTP_CHUNK_SIZE);
+		new Uint8Array(this.buffer).set(buf, this.chunk * HTTP_CHUNK_SIZE);
```
mishig25, line 91: is this line needed?

packages/gguf/src/gguf.ts
```diff
@@ -0,0 +1,275 @@
...
+		case 3: {
+			return view.getBigUint64(byteOffset, true);
+		}
+	}
+}
+
+function readString(view: DataView, offset: number): { value: string; newOffset: number } {
+	const length = view.getBigUint64(offset, true);
+	const value = new TextDecoder().decode(view.buffer.slice(offset + 8, offset + 8 + Number(length)));
+	return { value, newOffset: offset + 8 + Number(length) };
```
mishig25, line 120: instead of returning `{value, newOffset}`, imo it should return `{value, length}`
and then instead of using below `offset = keyResult.newOffset;` we can use `offset += keyResult.length;`

same applied to `readMetadataValue` below

packages/gguf/src/gguf.ts
```diff
@@ -0,0 +1,275 @@
...
+	const r = new RangeView(url);
+	await r.fetchChunk();
+
+	if (r.view.getUint32(0, true) !== new DataView(ggufMagicNumber.buffer).getUint32(0, true)) {
+		throw new Error("not a valid gguf file: not starting with GGUF magic number");
+	}
+
+	const version = r.view.getUint32(4, true);
+	if (!isVersion(version)) {
+		throw new Error("not a valid gguf file: unsupported version");
```
mishig25, line 197:
```suggestion
		throw new Error(`not a valid gguf file: unsupported version "${version}"`);
```
make it little more descriptive

packages/gguf/src/gguf.ts
platypii: `length` is better than `newOffset` I agree. I will be making that change to `hyllama` too.

coyotte508:
Unless I'm mistaken it only supports public non-gated repos

packages/gguf/src/gguf.ts
```diff
@@ -0,0 +1,275 @@
...
+/**
+ * Internal stateful instance to fetch ranges of HTTP data when needed
+ */
+class RangeView {
+	private chunk: number;
+	private buffer: ArrayBuffer;
+
+	readonly view: DataView;
+
+	constructor(public url: string) {
```
mishig25, line 65:
```suggestion
	constructor(public url: string, private headers?: Record<string, string>) {
```

regarding [this comment](https://github.com/huggingface/huggingface.js/pull/540#issuecomment-1986870330), `RangeView` can have private property `headers` where Authorization header, amongst other header options, could be passed. Afterwards, `this.headers` could be used inside `fetch` inside `fetchChunk`

packages/gguf/src/gguf.ts
```diff
@@ -0,0 +1,275 @@
...
+	UINT64 = 10,
+	INT64 = 11,
+	FLOAT64 = 12,
+}
+function isGGUFValueType(n: number): n is GGUFValueType {
+	return typeof GGUFValueType[n] === "string";
+}
+
+const HTTP_CHUNK_SIZE = 2 * 10 ** 6; /// 2MB
+const HTTP_DATA_LEEWAY = 5 * 10 ** 5; /// 500kb
```
mishig25, line 54:
```suggestion
const HTTP_CHUNK_SIZE = 2_000_000; /// 2MB
const HTTP_DATA_LEEWAY = 500_000; /// 500KB
```

up to you. for me, it is easier to see it this way

packages/gguf/src/gguf.ts
```diff
@@ -0,0 +1,275 @@
...
+		// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+		// @ts-ignore
+		this.buffer.resize((this.chunk + 1) * HTTP_CHUNK_SIZE);
+		new Uint8Array(this.buffer).set(buf, this.chunk * HTTP_CHUNK_SIZE);
+		this.chunk += 1;
+	}
+	/**
+	 * Check whether we need to fetch a new chunk
+	 */
+	async check(offset: number) {
```
mishig25, line 97:
```suggestion
	async fetchChunkIfNeeded(offset: number) {
```

nit. otherwise, `async check(offset: number)` sounds like a function that should return `bool`

packages/gguf/src/gguf.ts
```diff
@@ -0,0 +1,275 @@
...
+	version: Version;
+	tensor_count: bigint;
+	kv_count: bigint;
+} & Record<string, MetadataValue>;
+
+export interface GGUFTensorInfo {
+	name: string;
+	n_dims: number;
+	shape: bigint[];
+	type: GGMLQuantizationType;
```
mishig25, line 178:
```suggestion
	dtype: GGMLQuantizationType;
```

to be consistent with general ml lingo

packages/gguf/src/gguf.ts
```diff
@@ -0,0 +1,275 @@
...
+	}
+}
+
+function readString(view: DataView, offset: number): { value: string; newOffset: number } {
+	const length = view.getBigUint64(offset, true);
+	const value = new TextDecoder().decode(view.buffer.slice(offset + 8, offset + 8 + Number(length)));
+	return { value, newOffset: offset + 8 + Number(length) };
+}
+
+function readMetadataValue(
```
mishig25, line 123: `readVersionedSize`, `readString`, & `readMetadataValue` could be become methods of extended `DataView`

```ts
class GgufDataView extends DataView {
    readMetadataValue(){
        ...
    }
}
```

in which case, it would be `readMetadataValue(view, arrayType, arrayOffset)` VS `view.readMetadataValue(arrayType, arrayOffset)`

packages/gguf/src/gguf.ts
```diff
@@ -0,0 +1,275 @@
...
+	private buffer: ArrayBuffer;
+
+	readonly view: DataView;
+
+	constructor(public url: string) {
+		this.chunk = 0;
+		/// TODO(fix typing)
+		// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+		// @ts-ignore
+		this.buffer = new ArrayBuffer(0, { maxByteLength: 50 * 10 ** 6 });
```
mishig25, line 70: we should add `const MAX_BYTE_LENGTH` to these consts & use `MAX_BYTE_LENGTH` as `new ArrayBuffer(0, { maxByteLength: MAX_BYTE_LENGTH });`

https://github.com/huggingface/huggingface.js/blob/312aef0710b29428453013a10d33ac5e3698e357/packages/gguf/src/gguf.ts#L53-L54

julien-c:
Also tagging @Narsil for info

packages/gguf/src/gguf.ts
julien-c: i'd rather we pass an optional `fetch` like we do for safetensors:

https://github.com/huggingface/huggingface.js/blob/c26f6e0d9d26b0c0fe68ee8622d3a32433650d80/packages/hub/src/lib/parse-safetensors-metadata.ts#L63-L66

WDYT @coyotte508 @mishig25?

julien-c: will be more future-proof/flexible

julien-c: can you possibly make this change if you agree @mishig25?

packages/gguf/src/gguf.ts
julien-c: I tried doing it last friday but then it was failing my E2Es. 

I'll try again tomorrow

.github/workflows/agents-publish.yml
julien-c: ok with the upgrade of minimum Node version @coyotte508?

packages/gguf/src/gguf.ts
julien-c: 4eac1cde8fecb7abb31bf22620ee1e0b6655c896

packages/gguf/src/gguf.ts
julien-c: 9366d4ae27a0ff6e289b7f63280ff56d0c80a715

packages/gguf/src/gguf.ts
julien-c: 9366d4ae27a0ff6e289b7f63280ff56d0c80a715

packages/gguf/src/gguf.ts
julien-c: keeping them separate for now

julien-c:
@biw Yes, i liked the typing/validation in https://github.com/ahoylabs/gguf.js – neat idea (btw, opened this [small PR on gguf.js](https://github.com/ahoylabs/gguf.js/pull/6/files)). I'll keep this first PR minimal and we can revisit adding more types in the future.

packages/gguf/src/gguf.ts
```diff
@@ -0,0 +1,276 @@
...
+					headers: {
+						Range: `bytes=${range[0]}-${range[1]}`,
+					},
+				})
+			).arrayBuffer()
+		);
+		/// TODO(fix typing)
+		// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+		// @ts-ignore
+		this.buffer.resize((this.chunk + 1) * HTTP_CHUNK_SIZE);
```
julien-c, line 91: see typing discussion here: https://github.com/microsoft/TypeScript/issues/54636

packages/gguf/src/gguf.spec.ts
julien-c, line 1: ideally we would need a big-endian file to test too, to make sure our parsing is correct

from the spec:
> Must be `3` for version described in this spec, which introduces big-endian support.

julien-c:
also tagging @yagil 🔥

packages/gguf/src/gguf.ts
mishig25: yes, agree

packages/gguf/src/gguf.ts
mishig25: https://github.com/huggingface/huggingface.js/pull/543

packages/gguf/src/gguf.ts
mishig25: https://github.com/huggingface/huggingface.js/pull/543

packages/gguf/src/gguf.spec.ts
mishig25: testing big-endian fails
```ts
const { metadata, tensorInfos } = await gguf("https://huggingface.co/ggml-org/models/resolve/main/bert-bge-small/ggml-model-f16-big-endian.gguf");
```

> Error: not a valid gguf file: unsupported version "50331648"

ggerganov: The file seems to be converted correctly:

```bash
$ ▶ hexdump models/bert-bge-small/ggml-model-f16.gguf | head
0000000 4747 4655 0003 0000 00c5 0000 0000 0000     // this is the little-endian
0000010 0017 0000 0000 0000 0014 0000 0000 0000
0000020 6567 656e 6172 2e6c 7261 6863 7469 6365
0000030 7574 6572 0008 0000 0004 0000 0000 0000
0000040 6562 7472 000c 0000 0000 0000 6567 656e
0000050 6172 2e6c 616e 656d 0008 0000 0011 0000
0000060 0000 0000 6762 2d65 6d73 6c61 2d6c 6e65
0000070 762d 2e31 1035 0000 0000 0000 6200 7265
0000080 2e74 6c62 636f 5f6b 6f63 6e75 0474 0000
0000090 0c00 0000 1300 0000 0000 0000 6200 7265

$ ▶ hexdump models/bert-bge-small/ggml-model-f16-big-endian.gguf | head
0000000 4747 4655 0000 0300 0000 0000 0000 c500     // this is the converted big-endian
0000010 0000 0000 0000 1700 0000 0000 0000 1400
0000020 6567 656e 6172 2e6c 7261 6863 7469 6365
0000030 7574 6572 0000 0800 0000 0000 0000 0400
0000040 6562 7472 0000 0000 0000 0c00 6567 656e
0000050 6172 2e6c 616e 656d 0000 0800 0000 0000
0000060 0000 1100 6762 2d65 6d73 6c61 2d6c 6e65
0000070 762d 2e31 0035 0000 0000 0000 6210 7265
0000080 2e74 6c62 636f 5f6b 6f63 6e75 0074 0000
0000090 0004 0000 000c 0000 0000 0000 6213 7265
```

- The `"GGUF"` bytes are the same order for both LE and BE
- The next 4 bytes are the version - `3`
  This is what should be used to determine the endianness of the file. For more info, see this discussion: https://github.com/ggerganov/llama.cpp/issues/3957
- Next 8 bytes is the number of tensors: `c5 (hex) = 197 (dec)`
- etc.

Given the error `not a valid gguf file: unsupported version "50331648"`, it seems the reader that you are using does not check the version bytes to determine correctly the endianness

.github/workflows/agents-publish.yml
coyotte508: yes no problem

coyotte508:
JFI ok on my side, I'll fix tooling issues if any pop up

packages/gguf/src/gguf.ts
```diff
@@ -0,0 +1,325 @@
+type MetadataBaseValue = string | number | bigint | boolean;
+type MetadataValue = MetadataBaseValue | MetadataBaseValue[] | MetadataValue[]; /// recursive as arrays can be nested.
```
mishig25, line 2:
```suggestion
export type MetadataBaseValue = string | number | bigint | boolean;
export type MetadataValue = MetadataBaseValue | MetadataBaseValue[] | MetadataValue[]; /// recursive as arrays can be nested.
```

I'd need them to be exported for strongly typing the frontend metadata table

packages/gguf/src/gguf.spec.ts
```diff
@@ -0,0 +1,181 @@
+import { describe, expect, it } from "vitest";
+import { GGMLQuantizationType, gguf } from "./gguf";
+
+const URL_LLAMA = "https://huggingface.co/TheBloke/Llama-2-7B-Chat-GGUF/resolve/main/llama-2-7b-chat.Q2_K.gguf";
+const URL_MISTRAL_7B =
+	"https://huggingface.co/TheBloke/Mistral-7B-Instruct-v0.2-GGUF/resolve/main/mistral-7b-instruct-v0.2.Q5_K_M.gguf";
+const URL_GEMMA_2B = "https://huggingface.co/lmstudio-ai/gemma-2b-it-GGUF/resolve/main/gemma-2b-it-q4_k_m.gguf";
```
mishig25, line 7: nit: in subseq PR, we can probably move them under https://huggingface.co/huggingfacejs org so that the tests still work/pass if oweners of those repos make any change to those files or filenames

julien-c: or just replace main with the current revision?

mishig25 (approved):
lgtm ! 🚀

packages/gguf/src/gguf.ts
julien-c: sounds good, i let you do your final tweaks and merge whenever you want ok?

mishig25: yep

packages/gguf/src/gguf.spec.ts
coyotte508: There's also https://huggingface.co/datasets/huggingface/moon-tests (which we could move to a resource group :) )

mishig25: getting the current versions of models https://github.com/huggingface/huggingface.js/pull/540/commits/007c451bb94021d6a181b0523988cc9ac1b9b6fa

mishig25:
@coyotte508 please let me know if lgtm. I'm gonna merge soon

julien-c:
@biw i might add a little bit of typing and/or validation in the coming week – will ping here with a new PR cc @mishig25

PR #543, opened 2024-03-11
Gguf updates

mishig25:
1. [Use length rather than newOffset](https://github.com/huggingface/huggingface.js/pull/543/commits/fcab2c96be2e11c02fcdc7e3301c3c30164b75c7) (discussed [here](https://github.com/huggingface/huggingface.js/pull/540#discussion_r1519588604))
2. [custom fetch fn](https://github.com/huggingface/huggingface.js/pull/543/commits/18f93f37ce07f635dcd5cabb7156de4f7f1dc66f) (discussed [here](https://github.com/huggingface/huggingface.js/pull/540#discussion_r1519586431))

### DIFF
--- a/packages/gguf/src/gguf.ts
+++ b/packages/gguf/src/gguf.ts
@@ -63,7 +63,7 @@ class RangeView {
 
 	readonly view: DataView;
 
-	constructor(public url: string) {
+	constructor(public url: string, private _fetch: typeof fetch = fetch) {
 		this.chunk = 0;
 		/// TODO(fix typing)
 		// eslint-disable-next-line @typescript-eslint/ban-ts-comment
@@ -78,7 +78,7 @@ class RangeView {
 		const range = [this.chunk * HTTP_CHUNK_SIZE, (this.chunk + 1) * HTTP_CHUNK_SIZE - 1];
 		const buf = new Uint8Array(
 			await (
-				await fetch(this.url, {
+				await this._fetch(this.url, {
 					headers: {
 						Range: `bytes=${range[0]}-${range[1]}`,
 					},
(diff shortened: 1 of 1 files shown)


packages/gguf/src/gguf.ts
```diff
@@ -185,8 +188,8 @@ export interface GGUFParseOutput {
 	tensorInfos: GGUFTensorInfo[];
 }
 
-export async function gguf(url: string): Promise<GGUFParseOutput> {
-	const r = new RangeView(url);
+export async function gguf(url: string, _fetch: typeof fetch = fetch): Promise<GGUFParseOutput> {
```
julien-c, line 191:
```suggestion
export async function gguf(url: string, params: {
		/**
		 * Custom fetch function to use instead of the default one, for example to use a proxy or edit headers.
		 */
		fetch?: typeof fetch;
}): Promise<GGUFParseOutput> {
```

any way to pass it like this for consistency with safetensors?

packages/gguf/src/gguf.ts
```diff
@@ -185,8 +188,8 @@ export interface GGUFParseOutput {
...
+	const r = new RangeView(url, _fetch);
```
julien-c, line 192: same here

packages/gguf/src/gguf.ts
```diff
@@ -223,7 +226,7 @@ export async function gguf(url: string): Promise<GGUFParseOutput> {
 			throw new Error("Unsupported metadata type: " + valueType);
 		}
 
-		let valueResult: { value: MetadataValue; newOffset: number } | undefined;
+		let valueResult: ReturnType<typeof readMetadataValue> | undefined;
```
julien-c, line 229: 👍

julien-c (approved):
just one suggestion, other than that lgtm, thanks!

packages/gguf/src/gguf.ts
mishig25: https://github.com/huggingface/huggingface.js/pull/543/commits/682483a86e899d65a5dacd186e0a89bcdb91867c

PR #545, opened 2024-03-12
GGUF parser: support big-endian files

julien-c:
The important snippet is:

```ts
const [littleEndian, version] = (() => {
	/// https://github.com/ggerganov/llama.cpp/issues/3957
	/// Assume this code is always running on little-endian
	/// but wants to be able to parse both endianness
	const version = r.view.getUint32(4, true);
	if (version & 65535) {
		return [true, version];
	} else {
		return [false, r.view.getUint32(4, false)];
	}
})();
```

from https://github.com/ggerganov/llama.cpp/issues/3957 and thanks to @ggerganov [comment](https://github.com/huggingface/huggingface.js/pull/540/files#r1521103912)

### DIFF
--- a/packages/gguf/src/gguf.ts
+++ b/packages/gguf/src/gguf.ts
@@ -4,6 +4,12 @@ type MetadataValue = MetadataBaseValue | MetadataBaseValue[] | MetadataValue[]; 
 type Version = 1 | 2 | 3;
 const isVersion = (version: number): version is Version => version === 1 || version === 2 || version === 3;
 
+/**
+ * Must be `GGUF` at the byte level: `0x47` `0x47` `0x55` `0x46`.
+ * Your executor might do little-endian byte order, so it might be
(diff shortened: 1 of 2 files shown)


packages/gguf/src/gguf.ts
```diff
@@ -207,16 +218,36 @@ export async function gguf(
...
+		/// https://github.com/ggerganov/llama.cpp/issues/3957
+		/// Assume this code is always running on little-endian
+		/// but wants to be able to parse both endianness
+		const version = r.view.getUint32(4, true);
+		if (version & 65535) {
+			return [true, version];
+		} else {
+			return [false, r.view.getUint32(4, false)];
+		}
+	})();
```
julien-c, line 244: this is the important part

ggerganov (approved):

mishig25 (commented):
LGTM!

mishig25 (approved):
lgtm!

PR #546, opened 2024-03-13
[gguf] Fix fetch

mishig25:
Fix error when being used from svelte frontend:
```
Fetch TypeError: Failed to execute 'fetch' on 'Window': Illegal invocation
```


julien-c (commented):
lgtm

Issue #549, opened 2024-03-13
Streaming responses from scaled-to-zero Inference Endpoints return `undefined`


jinnovation:
Hey folks, what's the recommended way to deal with the initializing period of a scaled-to-zero Inference Endpoint when using [`.textGenerationStream()`](https://huggingface.co/docs/huggingface.js/en/inference/classes/HfInferenceEndpoint#textgenerationstream)?

For example, when using `.textGeneration()` directly, we get the 503 error directly that we can catch and maybe retry on.

However, when using `.textGenerationStream()`, fetching the next response chunk appears to succeed, except the `.value` at that chunk happens to be `undefined`. Reproducing code:

```typescript
import { HfInferenceEndpoint } from "@huggingface/inference";
import { experimental_buildLlama2Prompt } from "ai/prompts";

const hf = new HfInferenceEndpoint(
  "DEPLOYED_LLAMA2_ENDPOINT",
  "API_TOKEN",
);

const response = hf.textGenerationStream({
  inputs: experimental_buildLlama2Prompt([
    {
      role: "user",
      content: "hello",
    },
  ]),
});

response.next().then((res) => {
  console.log("print next chunk");
  console.log(res.value);
  console.log("done");
});

```

Running via terminal results in the following:
```
> npx ts-node hf-test.ts
print next chunk
undefined
done
```

My (possibly naive) expectation would be for the generator returned by `textGenerationStream()` to wait for the corresponding Inference Endpoint to fully initialize -- maybe with exponential backoff -- before allow the chunk to be returned via `.next()`. But I am a novice here so I could be missing something. 😆 

Is this expected behavior or a bug?

coyotte508:
@co42 @Narsil , does HF endpoints support `wait_for_model` like Inference API?

coyotte508:
It would be nice for API compatibiltity between inference API & inference endpoints!

coyotte508:
@jinnovation you will now get a 503 in streaming requests with most recent version `2.6.5`

With your current version, you can get a 503 if you set `retry_on_error` to `false`:

```ts
const response = hf.textGenerationStream({
  inputs: experimental_buildLlama2Prompt([
    {
      role: "user",
      content: "hello",
    },
  ]),
  {
    retry_on_error: false
  }
});
```

Soon the inference endpoint API will be updated, so that a call by default with `@huggingface/inference` will _wait_ until the model is loaded (you can disable this behavior with `retry_on_error: false` to handle the 503 yourself)

[...]

PR #554, opened 2024-03-14
Document use of `retry_on_error` for dedicated inference endpoints

jinnovation:
Shortly after #549, the inference endpoint backend
was updated to block by default on model loading.
This PR adds documentation explaining how to
circumvent that blocking so that the user, if
desired, can handle the 500 errors themselves.

### DIFF
--- a/packages/inference/README.md
+++ b/packages/inference/README.md
@@ -506,6 +506,21 @@ const gpt2 = hf.endpoint('https://xyz.eu-west-1.aws.endpoints.huggingface.cloud/
 const { generated_text } = await gpt2.textGeneration({inputs: 'The answer to the universe is'});
 ```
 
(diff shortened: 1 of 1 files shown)


jinnovation:
CC: @coyotte508

coyotte508:
Thank you!

Just a note, the backend hasn't been updated yet but should be in the coming days.

(And waiting for endpoints to be scaled actually needs https://github.com/huggingface/huggingface.js/pull/555, my bad)

coyotte508 (approved):

coyotte508:
Backend was updated, and version `2.6.6` on the lib should work fine with inference endpoints :)

jinnovation:
Amazing (: will give it a try soon-ish and report back.

PR #555, opened 2024-03-14
♻️ Always pass options through headers

coyotte508:
Thanks @Wauplin for pointing out :facepalm: 

We were only using the header for binary content, but it's better to use it always


radames:
Great catch, I'm surprised I've never noticed it using text-to-image, maybe because I was using different seeds

coyotte508:
> Great catch, I'm surprised I've never noticed it using text-to-image, maybe because I was using different seeds

I mean, it still worked, just it's better to always pass by header. It will be compatible with inference endpoints that way

PR #559, opened 2024-03-15
Add IFE task page and edit FE task page

merveenoyan:


merveenoyan:
I will lint once the content gets reviewed

packages/tasks/src/tasks/image-feature-extraction/about.md
```diff
@@ -0,0 +1,24 @@
+## Use Cases
+
+### Transfer Learning
+
+Models trained on a specific dataset can learn very shallow features about the data. For instance, a model trained on car classification dataset learns to recognize edges and curves on a very high level and car-specific features on a low level. This information can be transferred to a new model that is going to be trained on classifying trucks. This way of extracting features and transferring to a model is called feature extraction. One can pass their dataset through a feature extraction pipeline and feed the result to a classifier directly.
+
+### Similarity
+
+Features extracted from models contain semantically meaningful information about the world. These features can be used to detect the similarity between two images. Assume there are two images, a photo of a stray cat in a street setting, and a photo of a cat at home. These images both contain cats, and the features will contain the information that there's a cat in the image. Thus, comparing the features of a stray cat photo to the features of a domestic cat photo will result in higher similarity compared to any other image that doesn't contain any cats.
```
osanseviero, line 9:
```suggestion
Features extracted from models contain semantically meaningful information about the world. These features can be used to detect the similarity between two images. Assume there are two images: a photo of a stray cat in a street setting and a photo of a cat at home. These images both contain cats, and the features will contain the information that there's a cat in the image. Thus, comparing the features of a stray cat photo to the features of a domestic cat photo will result in higher similarity compared to any other image that doesn't contain any cats.
```

packages/tasks/src/tasks/image-feature-extraction/about.md
```diff
@@ -0,0 +1,24 @@
...
+
+## Inference
+
+```python
+import torch
+from transformers import pipeline
+
+DEVICE = torch.device('cuda' if torch.cuda.is_available() else 'cpu')
```
osanseviero, line 17: Let's aim to be consistent with how we set device across the task pages. Some task pages don't use `cuda`, some set `cuda` directly, some do this approach. Let's make them uniform

packages/tasks/src/tasks/image-feature-extraction/data.ts
```diff
@@ -0,0 +1,52 @@
+import type { TaskDataCustom } from "..";
+
+const taskData: TaskDataCustom = {
+	datasets: [
+		{
+			description:
+				"Wikipedia dataset containing cleaned articles of all languages. Can be used to train `feature-extraction` models.",
```
osanseviero, line 7: TODO?

packages/tasks/src/tasks/image-feature-extraction/about.md
```diff
@@ -0,0 +1,24 @@
+## Use Cases
+
+### Transfer Learning
+
+Models trained on a specific dataset can learn very shallow features about the data. For instance, a model trained on car classification dataset learns to recognize edges and curves on a very high level and car-specific features on a low level. This information can be transferred to a new model that is going to be trained on classifying trucks. This way of extracting features and transferring to a model is called feature extraction. One can pass their dataset through a feature extraction pipeline and feed the result to a classifier directly.
```
pcuenca, line 5:
```suggestion
Feature extraction is a process that transforms raw input data into numerical representations that try to preserve the information in the original dataset. Machine learning models trained on a dataset learn distinctive features about the data and can be used as feature extractors. For instance, a model trained on a car classification dataset may automatically learn to recognize edges and curves on a very high level, and car-specific features such as wheels or chassis shapes. The extracted features can be used to train downstream tasks, such as a truck classifier.
```

I personally find a bit confusing the identification of feature extraction with transfer learning.

packages/tasks/src/tasks/image-feature-extraction/about.md
pcuenca: I've observed this too. My preference would be to remove all device placement code in the task pages, for simplicity, and to focus on the task itself. However, it's also useful to learn these details. Perhaps we could link to some relevan transformers docs where appropriate? Example: https://huggingface.co/docs/transformers/pipeline_tutorial

packages/tasks/src/tasks/index.ts
```diff
@@ -187,6 +188,7 @@ export const TASKS_DATA: Record<PipelineType, TaskData | undefined> = {
 	"fill-mask": getData("fill-mask", fillMask),
 	"graph-ml": undefined,
 	"image-classification": getData("image-classification", imageClassification),
+	"image-feature-extraction": getData("image-feature-extraction", imageFeatureExtraction),
```
pcuenca, line 191: Duplicated in line 231.

packages/tasks/src/tasks/image-feature-extraction/about.md
merveenoyan: in the context of hugging face models it's mostly used for that no? a generic definition would confuse users on what they should do with hugging face models.

packages/tasks/src/tasks/image-feature-extraction/data.ts
merveenoyan: ah yes I wanted to find an image pretraining dataset

packages/tasks/src/tasks/image-feature-extraction/about.md
merveenoyan: Sure I'll open a PR

packages/tasks/src/tasks/image-feature-extraction/about.md
pcuenca: > the [definition](https://github.com/huggingface/huggingface.js/pull/559/files/afc9a1f7e8e9b7a92595c426847688b857f1dc17#diff-97656c9eaf51ebd19b884b88d76afb46453747a0c18abc9b6348573d5b622667R48) in the data.ts doesn't really identify with transfer learning

I was referring to the sentence you wrote "This way of extracting features and transferring to a model is called feature extraction", I don't think it's accurate or I'm not understanding it right. How about the following then?

```suggestion
Models trained on a specific dataset can learn features about the data. For instance, a model trained on a car classification dataset learns to recognize edges and curves on a very high level and car-specific features on a low level. This information can be transferred to a new model that is going to be trained on classifying trucks. This process of extracting features and transferring to another model is called transfer learning.
```

If you think it's right, then we can similarly adapt the paragraph in the general feature extraction task.

pcuenca: Any thoughts about this? It basically rephrases "this way of extracting features and transferring to a model is called feature extraction", which I find confusing.

osanseviero (approved):
Overall LGTM! Let's wait for Pedro's review to have a second pair of eyes

packages/tasks/src/tasks/feature-extraction/about.md
```diff
@@ -1,20 +1,13 @@
-## About the Task
-
-Feature extraction is the task of building features intended to be informative from a given dataset,
-facilitating the subsequent learning and generalization steps in various domains of machine learning.
-
 ## Use Cases
 
-Feature extraction can be used to do transfer learning in natural language processing, computer vision and audio models.
+Models trained on a specific dataset can learn very shallow features about the data. For instance, a model trained on English poetry dataset learns English grammar on a very high level. This information can be transferred to a new model that is going to be trained on tweets. This way of extracting features and transferring to a model is called feature extraction. One can pass their dataset through a feature extraction pipeline and feed the result to a classifier directly.
```
osanseviero, line 3: nit grammarly suggestions 

```suggestion
Models trained on a specific dataset can learn very shallow features about the data. For instance, a model trained on an English poetry dataset learns English grammar at a very high level. This information can be transferred to a new model that is going to be trained on tweets. This way of extracting features and transferring them to a model is called feature extraction. One can pass their dataset through a feature extraction pipeline and directly feed the result to a classifier.
```

merveenoyan:
test doesn't seem related to this branch, can I merge this?  @coyotte508

coyotte508:
yes!

PR #561, opened 2024-03-18
Add grok as lib for download stats

osanseviero:


SBrandeis (approved):

julien-c (commented):
maybe name it grok to be more future proof?

osanseviero:
I named it based in the GH repo https://github.com/xai-org/grok-1, i think being consistent with there makes sense, but no strong opinion

julien-c:
i think it's slightly better to name it grok but no strong opinion either

pcuenca (approved):
Very cool!

julien-c (approved):

Wauplin (approved):

JioCreators (approved):

PR #563, opened 2024-03-20
[Safetensors parser] Support optional file path

mishig25:
### Before

Safetensors parser was in assumption that safetensors files are strictly named `model.safetensors`, which is not the case always.

### This PR

Support optional file names so that other names like `llama3.safetensors` would still work

### DIFF
--- a/packages/hub/src/lib/parse-safetensors-metadata.ts
+++ b/packages/hub/src/lib/parse-safetensors-metadata.ts
@@ -8,8 +8,10 @@ import { downloadFile } from "./download-file";
 import { fileExists } from "./file-exists";
 import { promisesQueue } from "../utils/promisesQueue";
 
-const SINGLE_FILE = "model.safetensors";
-const INDEX_FILE = "model.safetensors.index.json";
+export const SINGLE_FILE = "model.safetensors";
+export const INDEX_FILE = "model.safetensors.index.json";
+export const RE_SINGLE_FILE = /.+\.safetensors$/;
+export const RE_INDEX_FILE = /.+\.index\.json$/;
 const PARALLEL_DOWNLOADS = 5;
 const MAX_HEADER_LENGTH = 25_000_000;
 
@@ -159,6 +161,7 @@ export async function parseSafetensorsMetadata(params: {
 	 *
 	 * @default false
 	 */
+	filePath?: string;
 	computeParametersCount: true;
 	hubUrl?: string;
 	credentials?: Credentials;
@@ -176,6 +179,7 @@ export async function parseSafetensorsMetadata(params: {
 	 *
 	 * @default false
 	 */
+	filePath?: string;
 	computeParametersCount?: boolean;
 	hubUrl?: string;
 	credentials?: Credentials;
@@ -187,6 +191,7 @@ export async function parseSafetensorsMetadata(params: {
 }): Promise<SafetensorsParseFromRepo>;
 export async function parseSafetensorsMetadata(params: {
 	repo: RepoDesignation;
+	filePath?: string;
 	computeParametersCount?: boolean;
 	hubUrl?: string;
 	credentials?: Credentials;
@@ -203,8 +208,8 @@ export async function parseSafetensorsMetadata(params: {
 		throw new TypeError("Only model repos should contain safetensors files.");
 	}
 
-	if (await fileExists({ ...params, path: SINGLE_FILE })) {
-		const header = await parseSingleFile(SINGLE_FILE, params);
+	if (RE_SINGLE_FILE.test(params.filePath ?? "") || (await fileExists({ ...params, path: SINGLE_FILE }))) {
+		const header = await parseSingleFile(params.filePath ?? SINGLE_FILE, params);
 		return {
 			sharded: false,
 			header,
@@ -212,8 +217,8 @@ export async function parseSafetensorsMetadata(params: {
 				parameterCount: computeNumOfParamsByDtypeSingleFile(header),
 			}),
 		};
-	} else if (await fileExists({ ...params, path: INDEX_FILE })) {
-		const { index, headers } = await parseShardedIndex(INDEX_FILE, params);
+	} else if (RE_INDEX_FILE.test(params.filePath ?? "") || (await fileExists({ ...params, path: INDEX_FILE }))) {
+		const { index, headers } = await parseShardedIndex(params.filePath ?? INDEX_FILE, params);
 		return {
 			sharded: true,
 			index,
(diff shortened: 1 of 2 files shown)


packages/hub/src/lib/parse-safetensors-metadata.ts
```diff
@@ -8,8 +8,10 @@ import { downloadFile } from "./download-file";
 import { fileExists } from "./file-exists";
 import { promisesQueue } from "../utils/promisesQueue";
 
-const SINGLE_FILE = "model.safetensors";
-const INDEX_FILE = "model.safetensors.index.json";
+export const SINGLE_FILE = "model.safetensors";
+export const INDEX_FILE = "model.safetensors.index.json";
+export const RE_SINGLE_FILE = /.+\.safetensors$/;
+export const RE_INDEX_FILE = /.+\.index\.json$/;
```
coyotte508, line 14:
```suggestion
export const RE_SINGLE_FILE = /\.safetensors$/;
export const RE_INDEX_FILE = /\.index\.json$/;
```

packages/hub/src/lib/parse-safetensors-metadata.ts
```diff
@@ -187,6 +191,7 @@ export async function parseSafetensorsMetadata(params: {
 }): Promise<SafetensorsParseFromRepo>;
 export async function parseSafetensorsMetadata(params: {
 	repo: RepoDesignation;
+	filePath?: string;
```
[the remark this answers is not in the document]
mishig25: handled in e966cb9ad8ab894b7ac2499133ae9b4af00f3d7f

packages/hub/src/lib/parse-safetensors-metadata.ts
```diff
@@ -8,8 +8,10 @@ import { downloadFile } from "./download-file";
 import { fileExists } from "./file-exists";
 import { promisesQueue } from "../utils/promisesQueue";
 
-const SINGLE_FILE = "model.safetensors";
-const INDEX_FILE = "model.safetensors.index.json";
+export const SINGLE_FILE = "model.safetensors";
+export const INDEX_FILE = "model.safetensors.index.json";
```
mishig25, line 12: exporting them so that we can use it in moon

coyotte508: let's rename them if we export them

```suggestion
export const SAFETENSORS_DEFAULT_FILE = "model.safetensors";
export const SAFETENSORS_DEFAULT_INDEX_FILE = "model.safetensors.index.json";
```

mishig25: wdyt b94af07eecc631b499d99dc87c68ce3f02090b33

mishig25:
cc: @apolinario for the upcoming moon support

packages/hub/src/lib/parse-safetensors-metadata.ts
```diff
@@ -154,6 +156,10 @@ async function parseShardedIndex(
 export async function parseSafetensorsMetadata(params: {
 	/** Only models are supported */
 	repo: RepoDesignation;
+	/**
+	 * Relative file path to safetensors file inside `repo`. Defaults to `SINGLE_FILE` or `INDEX_FILE` (whichever one exists).
```
julien-c, line 160:
```suggestion
	 * Relative file path to safetensors file inside `repo`. Defaults to `SAFETENSORS_FILE` or `SAFETENSORS_INDEX_FILE` (whichever one exists).
```

packages/hub/src/lib/parse-safetensors-metadata.ts
```diff
@@ -8,8 +8,10 @@ import { downloadFile } from "./download-file";
 import { fileExists } from "./file-exists";
 import { promisesQueue } from "../utils/promisesQueue";
 
-const SINGLE_FILE = "model.safetensors";
-const INDEX_FILE = "model.safetensors.index.json";
+export const SAFETENSORS_FILE = "model.safetensors";
+export const SAFETENSORS_INDEX_FILE = "model.safetensors.index.json";
+export const RE_SAFETENSORS_FILE = /\.safetensors$/;
+export const RE_SAFETENSORS_INDEX_FILE = /\.index\.json$/;
```
julien-c, line 14:
```suggestion
export const RE_SAFETENSORS_INDEX_FILE = /\.safetensors\.index\.json$/;
```

julien-c: also better to be a little bit more specific imo

packages/hub/src/lib/parse-safetensors-metadata.ts
```diff
@@ -8,8 +8,10 @@ import { downloadFile } from "./download-file";
 import { fileExists } from "./file-exists";
 import { promisesQueue } from "../utils/promisesQueue";
 
-const SINGLE_FILE = "model.safetensors";
-const INDEX_FILE = "model.safetensors.index.json";
+export const SAFETENSORS_FILE = "model.safetensors";
+export const SAFETENSORS_INDEX_FILE = "model.safetensors.index.json";
```
julien-c, line 12:
```suggestion
export const SAFETENSORS_INDEX_FILE = "model.safetensors.index.json";
/// We advise model/library authors to use the filenames above for convention,
/// but in some situations safetensors weights have different filenames.
```

julien-c (approved):
lgtm with a few suggestions!

julien-c:
yes valid question from @coyotte508 actually (read too fast initially)

IIRC @Narsil the `shard index file` contains relative filenames (there's no notion of directory), right? we should always assume the single files are in the same folder imo, correct?

(we can fix this in a later PR though i guess – and add a test from a real model)

mishig25:
> (we can fix this in a later PR though i guess – and add a test from a real model)

merging this PR

osanseviero:
Note that for diffusion, the situation is a bit tricky as we have non-diffusers `safetensors` at the top directory, while diffusers `safetensors` in folders

See https://huggingface.co/stabilityai/sdxl-turbo/tree/main
* Files people usually download manually https://huggingface.co/stabilityai/sdxl-turbo/blob/main/sd_xl_turbo_1.0.safetensors (e.g. for Auto1111)
* File usually accessed via `diffusers` https://huggingface.co/stabilityai/sdxl-turbo/blob/main/text_encoder/model.fp16.safetensors

osanseviero:
Yes, this PR is good, but we might need to make some distinction for diffusers in the UI between these two types of safetensors models.

mishig25:
> Yes, this PR is good, but we might need to make some distinction for diffusers in the UI between these two types of safetensors models.

I think that task is for the moon

osanseviero:
Yes, this indeed is for moon

julien-c:
> yes valid question from @coyotte508 actually (read too fast initially)
> 
> IIRC @Narsil the `shard index file` contains relative filenames (there's no notion of directory), right? we should always assume the single files are in the same folder imo, correct?

Let's open an issue so we don't forget this, ok?

julien-c:
@osanseviero for diffusers on the main model page i think we will still only parse a file `model.safetensors` so we won't display a model size for diffusers (too complex to calculate from different files / not worth it for now IMO)

Issue #564, opened 2024-03-20
[Widgets] Log in to compute


mishig25:
### Feature

Require user to be logged in for compute

| figma1    | figma2 |
| -------- | ------- |
|<img src="https://github.com/huggingface/huggingface.js/assets/11827707/be5e5d35-ea8d-4f3d-b797-5e7f18ddd758">  | <img src="https://github.com/huggingface/huggingface.js/assets/11827707/701bafe4-63da-48a9-b6fa-f7daef8d3543">    |

cc: @julien-c @gary149 @osanseviero @SBrandeis @beurkinger

mishig25:
> (not sure about the 1st mockup actually: in case there are widget examples or cached-API examples we still want to display them no?)

for widgets with examples: maybe we can have "Login to compute" on top of all "compute" buttons

> cached-API examples

for non-logged in users:
1. the widget is interactive/working
2. when they send request, only send the request asking API to return if the example was cached
    a. if the example was cached -> show the output
    b. if the example was not cached -> "Log in to compute" btn appears

osanseviero:
I like it! Maybe we need to make it extra clear it's free?

mishig25:
updated design 

![image](https://github.com/huggingface/huggingface.js/assets/11827707/dc3a701a-da3c-48dd-8e87-a5c5b6726283)

PR #567, opened 2024-03-20
[widgets] Require "Log In" to call/compute infrence

mishig25:
Closes https://github.com/huggingface/huggingface.js/issues/564

### Description

image

1. `InferenceWidget` component has prop `isLoggedIn`. 
2. Save that `isLoggedIn` prop into svelte store so that other child components can use/access this boolean
3. If `isLoggedIn === false` & user tries to call inference, show `Log in popover`

#### Scenario 1:
Canonical (when user is not logged in, `Log in popover` will be shown when user tries to trigger inference from components that can call inference. These components are: `WidgetSubmitBtn`, `WidgetFileInput`, `WidgetDropzone`)

https://github.com/huggingface/huggingface.js/assets/11827707/db4d1b8d-215c-4d93-93ba-575d9cd23d8b

#### Scenario 2:
Widget examples. Nothing changes in terms of widget examples. When you choose a widget example, there are two possibilities:
1. `possibility 1`: API/server had that example output cached OR widget example has [user-supplied output](https://huggingface.co/docs/hub/models-widgets#example-outputs), in which case the cached output would be shown
4. `possibility 2`: API/server did NOT have that example output cached, in which case only input fields would be populated, and when a user click `Compute`,  `Log in popover` will be shown

Note: ~~when testing on conv widget, there is a widget example bug #572 that should be solved in separate PR~~ (fixed with https://github.com/huggingface/huggingface.js/pull/575)

### Testing locally

```
cd huggingface.js/packages/widgets
pnpm run dev --open
```


packages/widgets/src/lib/components/InferenceWidget/InferenceWidget.svelte
```diff
@@ -28,6 +28,7 @@
 	import ZeroShotImageClassificationWidget from "./widgets/ZeroShotImageClassificationWidget/ZeroShotImageClassificationWidget.svelte";
 	import type { WidgetType } from "@huggingface/tasks";
 	import WidgetInfo from "./shared/WidgetInfo/WidgetInfo.svelte";
+	import { isLoggedIn as isLoggedInStore } from "./stores.js";
```
mishig25, line 31: there is a tooling issue that should be resolved in separate PR where js/ts files gets imported with extensions `.js` 
cc: @coyotte508

krampstudio (commented):
When you hit <kbd>Cmd/Ctrl + Enter</kbd> all the popover open.
We shouldn't have multiple widgets on the same page, but it should be better to avoid side effects. 
image

mishig25:
> We shouldn't have multiple widgets on the same page, but it should be better to avoid side effects.

fixed by #577 & rebased this PR

krampstudio:
> > We shouldn't have multiple widgets on the same page, but it should be better to avoid side effects.
> 
> fixed by #577 & rebased this PR

now the login popup doesn't appear on <kbd>Cmd+Enter</kbd>

packages/widgets/src/lib/components/LogInPopover/LogInPopover.svelte
```diff
@@ -0,0 +1,35 @@
...
+					? `?next=${encodeURIComponent(window.location.href)}`
+					: ''}"
+				class="bg-black text-white px-3 py-1 rounded-full dark:!bg-gray-100 dark:!text-black">Log In</a
+			>
+			<span class="text-gray-400">or</span>
+			<a
+				href="https://huggingface.co/join{typeof window !== 'undefined'
+					? `?next=${encodeURIComponent(window.location.href)}`
+					: ''}"
+				class="py-1 rounded-full underline decoration-gray-400 underline-offset-2">Create an account</a
```
gary149, line 31: Maybe we should add "free" for those kind of CTA 🤔 

```suggestion
				class="py-1 rounded-full underline decoration-gray-400 underline-offset-2">Create a free account</a
```

mishig25:
still need some more changes. Will ping here once done in a bit

mishig25:
ready for review @krampstudio @gary149 @SBrandeis 

as pre-requisite to this PR, submitted & merged #575 #576 #577 & rebased current branch with those changes.

updates:

#### update 1
addressed https://github.com/huggingface/huggingface.js/pull/567#discussion_r1537644217 fixed mobile view 


#### update 2
addressed https://github.com/huggingface/huggingface.js/pull/567#issuecomment-2021091279 chnages from #577 `cmd+enter` to run is correctly triggering the popover

#### update 3
addressed https://github.com/huggingface/huggingface.js/pull/567#discussion_r1537641771 & fixed/simplified the popover component & its props

#### update 3
show popover over the maximize btn as well for non-logged in users


#### update 4
updated calls to api-inference https://github.com/huggingface/huggingface.js/pull/567/commits/a986ff8b8f013cfa2c6daac60b83ee218adeb993. Specifically, when you try to call api-inference, the UI should block & dont send any calls to api-inference. However, when you choose an example, then a call can still be sent to api-inference. In this case, the call is being sent with headers: `don't load the model & use cached result if available`. (we need to send this call to api-inference because on the frontend there is no way to know if the example is cached or not). Api-inference needs to implement this:
if a request comes from a non-logged user:
1. if cached output is found, then return that output
2. otherwise, throw error saying "You need to log in"

The scenario described above for api-inference is non-blocker for this PR. Therefore, we should review & merge this PR

packages/widgets/src/lib/components/Popover/Popover.svelte
```diff
@@ -0,0 +1,167 @@
...
+				return () => {
+					anchorElement.removeEventListener("mouseover", debouncedShow);
+					anchorElement.removeEventListener("mouseleave", debouncedHide);
+				};
+			}
+		}
+	});
+
+	$: if (open && !waitForContent) {
+		updatePosition();
```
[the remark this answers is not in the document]
mishig25: update 1 from https://github.com/huggingface/huggingface.js/pull/567#issuecomment-2024937025

packages/widgets/src/lib/components/Popover/Popover.svelte
```diff
@@ -0,0 +1,167 @@
...
+	}
+
+	$: if (open) {
+		isActive = true;
```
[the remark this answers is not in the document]
mishig25: update 3 from update 1 from https://github.com/huggingface/huggingface.js/pull/567#issuecomment-2024937025

krampstudio (approved):

gary149 (approved):

mishig25:
approval from @SBrandeis and I merge

gary149:
The way it disappear is a bit weird. @krampstudio good usecase for the popover on click too?

https://github.com/huggingface/huggingface.js/assets/3841370/f541b2f3-15e4-4a82-8323-83d2c763107c

SBrandeis:
reviewing now

SBrandeis:
The popover is very far away from the button 
image

SBrandeis (commented):
I can still run examples even when not logged in - is that expected ?


https://github.com/huggingface/huggingface.js/assets/33657802/e486dadd-23dd-4143-bebe-a8f1d7ad267e

mishig25:
> I can still run examples even when not logged in - is that expected ?

Yes, it is the last paragraph described in https://github.com/huggingface/huggingface.js/pull/567#issuecomment-2024937025

mishig25:
> I can still run examples even when not logged in - is that expected ?

what's the setup for testing on hub. Did you install this branch on hub locally?

SBrandeis:
> what's the setup for testing on hub. Did you install this branch on hub locally?

yes: `pnpm i && pnpm run build` on `huggingface.js/packages/widgets`

Then `npm i --save @huggingface.js/widgets@<relative path to packages/widgets>` in the Hub repo (server & front)

packages/widgets/src/lib/components/InferenceWidget/InferenceWidget.svelte
```diff
@@ -85,6 +86,8 @@
 			  ? WIDGET_COMPONENTS[model.pipeline_tag as keyof typeof WIDGET_COMPONENTS]
 			  : undefined;
 
+	$isLoggedInStore = isLoggedIn;
```
SBrandeis, line 89: The widget is not picking this up in the tasks page (I'm logged in but the popover still shows)
(needs fixing in moon-landing probably)

image

packages/widgets/src/lib/utils/ViewUtils.ts
```diff
@@ -156,6 +156,21 @@ export function onCmdEnter(node: HTMLElement, opts: { disabled: boolean }): Acti
 	};
 }
 
+/**
+ * A debounce function that works in both browser and Nodejs.
+ * For pure Nodejs work, prefer the `Debouncer` class.
```
SBrandeis, line 161:
```suggestion
```

packages/widgets/src/lib/components/InferenceWidget/shared/helpers.ts
```diff
@@ -102,10 +104,10 @@ export async function callInferenceApi<T>(
 	if (waitForModel) {
 		headers.set("X-Wait-For-Model", "true");
 	}
-	if (useCache === false) {
+	if (useCache === false && get(isLoggedIn) === true) {
```
SBrandeis, line 107:
```suggestion
	if (useCache === false && get(isLoggedIn)) {
```

packages/widgets/src/lib/components/InferenceWidget/shared/helpers.ts
```diff
@@ -102,10 +104,10 @@ export async function callInferenceApi<T>(
...
 		headers.set("X-Use-Cache", "false");
 	}
-	if (isOnLoadCall) {
+	if (isOnLoadCall || get(isLoggedIn) === false) {
```
SBrandeis, line 110:
```suggestion
	if (isOnLoadCall || !get(isLoggedIn)) {
```

packages/widgets/src/lib/components/InferenceWidget/shared/helpers.ts
```diff
@@ -102,10 +104,10 @@ export async function callInferenceApi<T>(
...
 		headers.set("X-Load-Model", "0");
 	}
```
SBrandeis, line 112: You'll need to update the logic inside `ConversationalWidget` as well (it uses `@huggingface.js/inference` to  call the inference API)

packages/widgets/src/lib/components/InferenceWidget/shared/WidgetFileInput/WidgetFileInput.svelte
```diff
@@ -21,36 +29,45 @@
...
+	<LogInPopover bind:open={popOverOpen}>
+		<button
+			class={classNames}
+			on:click={(e) => {
+				if (!$isLoggedIn) {
+					popOverOpen = true;
+					e.preventDefault();
+					return;
+				}
+			}}
```
SBrandeis, line 41: (nit) maybe cleaner ? wdyt @krampstudio
```suggestion
			on:click|preventDefault={(e) => {
				if (!$isLoggedIn) {
					popOverOpen = true;
					return;
				}
			}}
```

packages/widgets/src/lib/components/InferenceWidget/shared/WidgetFooter/WidgetFooter.svelte
```diff
@@ -28,18 +29,26 @@
...
+			}}
+		>
+			<IconMaximize classNames="mr-1" />
+			{#if !isMaximized}
+				Maximize
+			{:else}
+				Minimize
+			{/if}
+		</button>
+	</LogInPopover>
```
SBrandeis, line 51: What's the reason behind this? (preventing logged out users from maximizing the widget)

SBrandeis (commented):
cool cool cool

packages/widgets/src/lib/components/InferenceWidget/shared/WidgetFooter/WidgetFooter.svelte
mishig25: yes

packages/widgets/src/lib/components/InferenceWidget/shared/WidgetFileInput/WidgetFileInput.svelte
mishig25: I think we want `preventDefault` only if `!$isLoggedIn`. Otherwise, we want the default behaviour

packages/widgets/src/lib/components/InferenceWidget/shared/WidgetFooter/WidgetFooter.svelte
SBrandeis: I'm not convinced we should do that, but OK (I am probably missing context here)

packages/widgets/src/lib/components/InferenceWidget/shared/helpers.ts
SBrandeis: let's do this change and merge? @mishig25

mishig25: handled in d91a5672e831d62fb9355b32e13b125d662ace3b

mishig25:
> The popover is very far away from the button

from https://github.com/huggingface/huggingface.js/pull/567#issuecomment-2027223903 is fixed by https://github.com/huggingface/huggingface.js/pull/567/commits/200b04cdb62f2bdc817c9478d7e30d5303afeb82

Screenshot 2024-03-29 at 4 09 59 PM

packages/widgets/src/lib/components/InferenceWidget/shared/helpers.ts
SBrandeis: I think this is where you want to do the change:
https://github.com/huggingface/huggingface.js/blob/3f8a3f3c0b53639e4804bce2ee0871f2a0cd7d6f/packages/widgets/src/lib/components/InferenceWidget/widgets/ConversationalWidget/ConversationalWidget.svelte#L154-L161

SBrandeis: no?

mishig25: yep thats better 7138b8fc78e80aa2a07ed1f01cdb290ea2804995

SBrandeis (approved):

Issue #572, opened 2024-03-21
[Conversation Widget] Bug on examples


mishig25:
Related to https://github.com/huggingface/huggingface.js/pull/510

after choosing the initial example, the examples stop calling inference. You can in the video below that initially choosing example `Text only` works, but stops working when you select other example & go back to selecting `Text only`  example again

https://github.com/huggingface/huggingface.js/assets/11827707/9b8a2260-82ca-4888-93dd-78e5b023cc8c

cc: @SBrandeis @xenova @Wauplin

PR #575, opened 2024-03-25
[Widget] Fix examples

mishig25:
Closes https://github.com/huggingface/huggingface.js/issues/572

https://github.com/huggingface/huggingface.js/assets/11827707/feef9bf5-f8fc-4748-b23c-270c150bdf5a

### DIFF
--- a/packages/widgets/src/lib/components/InferenceWidget/shared/WidgetExamples/WidgetExamples.svelte
+++ b/packages/widgets/src/lib/components/InferenceWidget/shared/WidgetExamples/WidgetExamples.svelte
@@ -49,6 +49,9 @@
 	}
 
 	function _applyWidgetExample(idx: number) {
+		if(!isOptionsVisible){
+			return;
+		}
 		hideOptions();
 		const sample = examples[idx];
 		title = sample.example_title as string;
@@ -56,6 +59,9 @@
 	}
 
 	function _previewInputSample(idx: number) {
+		if(!isOptionsVisible){
+			return;
+		}
 		const sample = examples[idx];
 		applyWidgetExample(sample, { isPreview: true });
 	}
--- a/packages/widgets/src/lib/components/InferenceWidget/shared/WidgetExamples/WidgetExamples.svelte
+++ b/packages/widgets/src/lib/components/InferenceWidget/shared/WidgetExamples/WidgetExamples.svelte
@@ -49,7 +49,7 @@
 	}
 
 	function _applyWidgetExample(idx: number) {
-		if(!isOptionsVisible){
+		if (!isOptionsVisible) {
 			return;
 		}
 		hideOptions();
@@ -59,7 +59,7 @@
 	}
 
 	function _previewInputSample(idx: number) {
-		if(!isOptionsVisible){
+		if (!isOptionsVisible) {
 			return;
 		}
 		const sample = examples[idx];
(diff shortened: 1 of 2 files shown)


packages/widgets/src/lib/components/InferenceWidget/shared/WidgetExamples/WidgetExamples.svelte
```diff
@@ -49,13 +49,19 @@
...
 		hideOptions();
 		const sample = examples[idx];
 		title = sample.example_title as string;
 		applyWidgetExample(sample);
 	}
 
 	function _previewInputSample(idx: number) {
+		if (!isOptionsVisible) {
+			return;
+		}
```
mishig25, line 64: needed those because of https://github.com/huggingface/huggingface.js/blob/13b95853a57718794b1bd21ccb522474a9a656b3/packages/widgets/src/lib/components/InferenceWidget/shared/WidgetExamples/WidgetExamples.svelte#L158

this `transition:slide`, when a user clicks & while options were being closed, https://github.com/huggingface/huggingface.js/blob/13b95853a57718794b1bd21ccb522474a9a656b3/packages/widgets/src/lib/components/InferenceWidget/shared/WidgetExamples/WidgetExamples.svelte#L165 was still being trigerred

packages/widgets/src/lib/components/InferenceWidget/widgets/ConversationalWidget/ConversationalWidget.svelte
```diff
@@ -214,24 +214,16 @@
...
-			if (opts.isPreview) {
-				return;
-			}
-			const exampleOutput = example.output;
-			await getOutput({ ...opts.inferenceOpts, exampleOutput });
-		} finally {
-			isLoading = false;
+		if (opts.isPreview) {
+			return;
 		}
```
mishig25, line 224: no need to manage `isLoading` in `aaplyWidgetExample` since it is already managed in `getOutput` function https://github.com/huggingface/huggingface.js/blob/13b95853a57718794b1bd21ccb522474a9a656b3/packages/widgets/src/lib/components/InferenceWidget/widgets/ConversationalWidget/ConversationalWidget.svelte#L197-L198

SBrandeis: So the initial issue was caused by `ìsLoading` not being properly reset to `false`, right?

Would be worth investigating how / why this happened

SBrandeis: I'm having a hard time connecting your changes to the initial issue (inference API not being called for examples after the first)

Can you explain why / how those changes fix it? 🙏

packages/widgets/src/lib/components/InferenceWidget/shared/WidgetExamples/WidgetExamples.svelte
SBrandeis: nice catch

packages/widgets/src/lib/components/InferenceWidget/widgets/ConversationalWidget/ConversationalWidget.svelte
mishig25: > I'm having a hard time connecting your changes to the initial issue (inference API not being called for examples after the first)

You are right. This is not related to the issue described in this PR. https://github.com/huggingface/huggingface.js/pull/575#discussion_r1537613029 solves the issue completely. 

Just added this change since I've noticed this redundant `isLoading ` management. Can put it in another PR as well

mishig25:
Merging it to conitnue https://github.com/huggingface/huggingface.js/pull/567

Issue #565, opened 2024-03-20
[Widget] Computation device bug


mishig25:
<img width="259" alt="image" src="https://github.com/huggingface/huggingface.js/assets/11827707/d0a5ac4f-65f9-4646-a978-13d63af24032">

instead of `cpu/gpu`, it is showing `[object Object]`

you can try at: https://huggingface.co/google/gemma-7b

cc: @SBrandeis  @Wauplin @xenova

julien-c:
is this a change from the backend @OlivierDehaene @Narsil?

PR #576, opened 2024-03-25
[Widget] fix computation device label

mishig25:
Closes https://github.com/huggingface/huggingface.js/issues/565

I guess there was a change in the shape of output from from api-inference status endpoints. For instance:
```
curl https://api-inference.huggingface.co/status/google/gemma-7b
```
would give:
```
{"loaded":false,"state":"Loadable","compute_type":{"gpu":{"gpu":"a10-g","count":4}}
```

Widget frontend/UI assumption was that `compute_type` field would only have string value of `cpu` or `gpu`. However, you can see un the curl that, it can be `{"gpu":{"gpu":"a10-g","count":4}`.

Therefore, this PR handles this possible shape change

Screenshot 2024-03-25 at 3 00 04 PM

### DIFF
--- a/packages/widgets/src/lib/components/InferenceWidget/shared/helpers.ts
+++ b/packages/widgets/src/lib/components/InferenceWidget/shared/helpers.ts
@@ -184,7 +184,7 @@ export async function getModelLoadInfo(
 		const compute_type =
 			typeof output.compute_type === "string"
 				? output.compute_type
-				: (Object.keys(output.compute_type)[0] as ComputeType);
+				: output.compute_type["gpu"] ? ("gpu" as const) : ("cpu" as const);
 		return { compute_type, state: output.state };
 	} else {
 		console.warn(response.status, output.error);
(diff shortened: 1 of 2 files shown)


packages/widgets/src/lib/components/InferenceWidget/shared/helpers.ts
```diff
@@ -173,11 +173,19 @@ export async function getModelLoadInfo(
...
+	} = await response.json();
 	if (response.ok && typeof output === "object" && output.loaded !== undefined) {
 		// eslint-disable-next-line @typescript-eslint/naming-convention
-		const { state, compute_type } = output;
-		return { compute_type, state };
+		const compute_type =
+			typeof output.compute_type === "string"
+				? output.compute_type
+				: (Object.keys(output.compute_type)[0] as ComputeType);
+		return { compute_type, state: output.state };
```
SBrandeis, line 188: lol the shape is a bit unpractical to work with

nit (feel free to ignore): we could favor `gpu` over `cpu` if present


```suggestion
		const compute_type =
			typeof output.compute_type === "string"
				? output.compute_type
				: output.compute_type["gpu"] ? ("gpu" as const) : ("cpu" as const);
		return { compute_type, state: output.state };
```

SBrandeis (approved):
Thank you for fixing this!

SBrandeis (approved):

PR #579, opened 2024-03-25
fix: update chat and generate specs

drbh:
This PR updates the `chat-completion` and `text-generation` schemas to align with TGI's api. 

idea: 
longer term it would be nice to add some functionality to TGI to automatically generate these schemas (we are using `utopia` to generate API docs and can likely extend it to help automate this and reduce likelihood of human error).

Please let me know if anything needs to be updated 🙏


packages/tasks/src/tasks/chat-completion/spec/input.json
```diff
@@ -17,12 +17,16 @@
...
+					"name": {
+						"type": "string",
+						"description": "The name of the message."
 					}
 				},
 				"required": ["role", "content"]
 			}
 		},
-		"frequency_penalty": {
+		"presence_penalty": {
```
Wauplin, line 29: For what I can see [here](https://huggingface.github.io/text-generation-inference/#/Text%20Generation%20Inference/chat_completions), it looks like both `frequency_penalty` and `presence_penalty` exist. What's the difference between those two? Should we add both to the specs?

packages/tasks/src/tasks/chat-completion/spec/input.json
```diff
@@ -17,12 +17,16 @@
 					"content": {
 						"type": "string",
 						"description": "The content of the message."
+					},
+					"name": {
+						"type": "string",
+						"description": "The name of the message."
```
Wauplin, line 23:
```suggestion
						"description": "The name of the message. Provides the model information to differentiate between participants of the same role."
```

packages/tasks/src/tasks/chat-completion/spec/input.json
```diff
@@ -47,17 +51,42 @@
...
+		"Tool": {
+			"title": "ChatCompletionTool",
+			"description": "The tool to use for generation.",
+			"properties": {
+				"type": {
+					"type": "string",
+					"description": "The type of tool to use."
+				},
+				"function": {
+					"type": "object",
```
Wauplin, line 85: Does `function` always look like `name: str, description: Optional[str], parameters: object`? (taken from [here](https://platform.openai.com/docs/api-reference/chat/create)) or is it different? If always the same, let's add it to the specs.

packages/tasks/src/tasks/chat-completion/spec/input.json
```diff
@@ -47,17 +51,42 @@
 			"type": "number",
 			"description": "The value used to modulate the logits distribution."
 		},
+		"tools": {
```
Wauplin, line 54: Should we also update `output.json` to include `output.choices[].message.tool_calls` or not yet supported by TGI?

Wauplin: Same comment for `output.choices[].delta.tool_calls` in `output_stream.json`?

packages/tasks/src/tasks/text-generation/spec/input.json
```diff
@@ -88,6 +93,33 @@
...
+						{
+							"type": "string"
+						},
+						{
+							"type": "object"
+						}
+					]
+				}
+			},
+			"required": ["rules", "start"]
```
Wauplin, line 122:
```suggestion
			"required": ["type", "value"]
```

Wauplin (commented):
Thanks for adding those @drbh! I've reviewed your changes and have some questions, mostly based on what I can read on [openai docs](https://platform.openai.com/docs/api-reference/chat/create). Feel free to ignore when it's not applicable to TGI.

Regarding auto-generating them, I'm all for it - especially if TGI is already typed and annotated. Most specs comes from transformers' pipelines that are not spec-ed so can't auto-magically create those.

packages/tasks/src/tasks/text-generation/spec/input.json
```diff
@@ -88,6 +93,33 @@
 					"description": "Watermarking with [A Watermark for Large Language Models](https://hf.co/papers/2301.10226)"
 				}
 			}
+		},
+		"GrammarType": {
+			"title": "GrammarType",
```
SBrandeis, line 98: (needed for type names to be prefixed by `TextGeneration`)
```suggestion
			"title": "TextGenerationGrammarType",
```

packages/tasks/src/tasks/text-generation/spec/input.json
```diff
@@ -88,6 +93,33 @@
...
+			"type": "string",
+			"description": "The type of grammar to use for generation",
+			"enum": ["json", "regex"]
+		},
+		"Grammar": {
+			"title": "Grammar",
```
SBrandeis, line 104:
```suggestion
			"title": "TextGenerationGrammar",
```

packages/tasks/src/tasks/chat-completion/spec/input.json
```diff
@@ -47,17 +51,42 @@
...
+			"type": "string",
+			"description": "The tool to use for generation."
+		},
 		"top_p": {
 			"type": "number",
 			"description": "If set to < 1, only the smallest set of most probable tokens with probabilities that add up to `top_p` or higher are kept for generation."
 		}
 	},
-	"required": ["messages"],
 	"definitions": {
```
SBrandeis, line 70: Unrelated to this PR but this section should be renamed `$defs` for consistency with other tasks

SBrandeis (commented):
Thank you for taking the time to do this @drbh !

julien-c (commented):
very cool PR, let's get it over the line soon-ish?

packages/tasks/src/tasks/chat-completion/spec/input.json
drbh: Hi @Wauplin thanks for all of the helpful comments. You're 100% right we should include both of those values.

details: https://platform.openai.com/docs/guides/text-generation/frequency-and-presence-penalties

***screen below for ref
frequency_presence_penalty_diff

drbh:
I believe that the failing CI is unrelated and all of the comments above have been addressed

Wauplin (commented):
Thanks @drbh for the changes and sorry for the late review :grimacing: 
Overall looks good to me. Would it be possible to add `tool_calls` to both the stream and non-stream outputs? (equivalent to [here](https://github.com/huggingface/text-generation-inference/blob/106d8ee81899e6461648e1efd56f5a873ce7294d/clients/python/text_generation/types.py#L39) and [here](https://github.com/huggingface/text-generation-inference/blob/106d8ee81899e6461648e1efd56f5a873ce7294d/clients/python/text_generation/types.py#L77) in TGI client). Once we have this, we should be good to go. Thanks in advance!

Wauplin:
Superseded by https://github.com/huggingface/huggingface.js/pull/629.

PR #585, opened 2024-03-26
Fix Types in Build

### LINKED ISSUE #584: Missing Type From the Inference Package
In the types of the `@huggingface/inference` package, the `textGeneration` function is declared as follows

```ts
export function textGeneration(
    args: BaseArgs & TextGenerationInput,
    options?: Options
): Promise<TextGenerationOutput>;
```

But the `TextGenerationOutput` type is missing from the file, so the return type is inferred as `any` by TypeScript.

In the [source code](https://github.com/huggingface/huggingface.js/blob/b89956d3449e6cfd4f78a3d1350fb94e478a3bb7/packages/inference/src/tasks/nlp/textGeneration.ts#L1), the `TextGenerationOutput` type is imported from [this file](https://github.com/huggingface/huggingface.js/blob/b89956d3449e6cfd4f78a3d1350fb94e478a3bb7/packages/tasks/src/tasks/text-generation/inference.ts#L102) in a separate package, which does correctly have the type, so this is likely an issue with the build step.

abdel-17:
Fixes #584

### DIFF
--- a/packages/inference/src/tasks/nlp/textGeneration.ts
+++ b/packages/inference/src/tasks/nlp/textGeneration.ts
@@ -1,7 +1,209 @@
-import type { TextGenerationInput, TextGenerationOutput } from "@huggingface/tasks/src/tasks/text-generation/inference";
 import { InferenceOutputError } from "../../lib/InferenceOutputError";
 import type { BaseArgs, Options } from "../../types";
 import { request } from "../custom/request";
+
+/**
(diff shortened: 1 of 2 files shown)


coyotte508 (approved):
Thanks

Ok for now, later let's try to do as in https://github.com/huggingface/huggingface.js/issues/584#issuecomment-2022180989

cc @SBrandeis

PR #587, opened 2024-03-28
[gguf] add `RE_GGUF_FILE` regex

mishig25:
Add gguf equavalent of https://github.com/huggingface/huggingface.js/blob/2bfaa65fe895fc00e65879999225d238f43ef9b0/packages/hub/src/lib/parse-safetensors-metadata.ts#L15-L16


julien-c (commented):
lgtm

Issue #588, opened 2024-03-28
Safetensors sharded model inspector does not work in subdirs


madgetr:
![image](https://github.com/huggingface/huggingface.js/assets/16808224/f9f72094-67bb-4c97-a52d-52d37ceb9709)
It seems that the model inspector for sharded safetensors does not work when the model is in a sub directory.

coyotte508:
cc @mishig25 @julien-c

coyotte508:
We have an internal PR open for the hub that should fix it @madgetr (cc @mishig25)

mishig25:
Yep it should be fixed by https://github.com/huggingface/huggingface.js/pull/563

the hf.co needs to use latest hf.js

coyotte508:
hmm thinking about it the issue is sharded models, and the hub PR is for single-file models I think 

This issue is related to https://github.com/huggingface/huggingface.js/issues/592 I think

julien-c:
@madgetr can you provide a link to the model (if it's a public one) you screenshoted that has a `safetensors sharded file inside a subdir`? So we can make sure it works:)

madgetr:
https://huggingface.co/Alignment-Lab-AI/ALAI-gemma-7b/tree/main/7b/1

mishig25:
thanks a lot for notifying the issue again. The fix is deployed. You can see a working example at: https://huggingface.co/Alignment-Lab-AI/ALAI-gemma-7b/tree/main/7b/1?show_tensors=7b%2F1%2Fmodel.safetensors.index.json

![Screenshot 2024-04-02 at 12 34 23 PM](https://github.com/huggingface/huggingface.js/assets/11827707/8355f070-19f7-40bd-9dfb-578040f67c02)

PR #594, opened 2024-03-29
[safetensors] sharded with file path

mishig25:
Fixes https://github.com/huggingface/huggingface.js/issues/588

cc: @madgetr

### DIFF
--- a/packages/hub/src/lib/parse-safetensors-metadata.spec.ts
+++ b/packages/hub/src/lib/parse-safetensors-metadata.spec.ts
@@ -83,4 +83,25 @@ describe("parseSafetensorsMetadata", () => {
 		assert.deepStrictEqual(parse.parameterCount, { F32: 859_520_964 });
 		assert.deepStrictEqual(sum(Object.values(parse.parameterCount)), 859_520_964);
 	});
+
+	it("fetch info for sharded (with the default conventional filename) with file path", async () => {
+		const parse = await parseSafetensorsMetadata({
+			repo: "Alignment-Lab-AI/ALAI-gemma-7b",
+			computeParametersCount: true,
+			path: "7b/1/model.safetensors.index.json",
+		});
+
+		assert(parse.sharded);
+
+		assert.strictEqual(Object.keys(parse.headers).length, 4);
+
+		assert.deepStrictEqual(parse.headers["model-00004-of-00004.safetensors"]["model.layers.24.mlp.up_proj.weight"], {
+			dtype: "BF16",
+			shape: [24576, 3072],
+			data_offsets: [301996032, 452990976],
+		});
+
+		assert.deepStrictEqual(parse.parameterCount, { BF16: 8_537_680_896 });
+		assert.deepStrictEqual(sum(Object.values(parse.parameterCount)), 8_537_680_896);
+	});
 });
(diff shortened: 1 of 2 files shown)


madgetr (commented):
Thanks for tackling this, I was just about to comment confirming that it is indeed broken!

packages/hub/src/lib/parse-safetensors-metadata.spec.ts
```diff
@@ -83,4 +83,25 @@ describe("parseSafetensorsMetadata", () => {
 		assert.deepStrictEqual(parse.parameterCount, { F32: 859_520_964 });
 		assert.deepStrictEqual(sum(Object.values(parse.parameterCount)), 859_520_964);
 	});
+
+	it("fetch info for sharded (with the default conventional filename) with file path", async () => {
+		const parse = await parseSafetensorsMetadata({
```
madgetr, line 88: I noticed that the test is not commit aware, if this model is deleted then this test will fail.

julien-c: yep we can pass a `revision` (and we'd recreate this test if the repo is ever completely deleted)

packages/hub/src/lib/parse-safetensors-metadata.ts
```diff
@@ -139,12 +140,13 @@ async function parseShardedIndex(
 		throw new SafetensorParseError(`Failed to parse file ${path}: not a valid JSON.`);
 	}
 
+	const pathPrefix = path.match(RE_FILE_DIR)?.[1] ?? "";
```
julien-c, line 143: maybe clearer intent to do something like (untested code):
```ts
const parentFolder = path.substr(0, path.lastIndexOf("/"))
```
?

packages/hub/src/lib/parse-safetensors-metadata.spec.ts
mishig25: handled in 06ab3c0f33420ddad25974f26d3e15d4521106f6

packages/hub/src/lib/parse-safetensors-metadata.ts
mishig25: handled in 75ce40a190cdc910d9ca667d0650d1d1860152bc

madgetr (approved):
+1

julien-c (approved):
lgtm 

(CI failure is unrelated i think?)

packages/hub/src/lib/parse-safetensors-metadata.ts
```diff
@@ -139,12 +139,13 @@ async function parseShardedIndex(
 		throw new SafetensorParseError(`Failed to parse file ${path}: not a valid JSON.`);
 	}
 
+	const pathPrefix = path.substr(0, path.lastIndexOf("/") + 1);
```
coyotte508, line 142: `substr` is deprecated for strings: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr

(`.slice` is my goto usually)

mishig25: https://github.com/huggingface/huggingface.js/pull/598

julien-c: it's still in the LLMs though =)

PR #600, opened 2024-04-03
[safetensors metadata] 8x faster parsing for sharded

mishig25:
https://huggingface.co/databricks/dbrx-instruct?show_tensors=model.safetensors.index.json is being quite slow because it needs to parse headers of [61 files](https://huggingface.co/databricks/dbrx-instruct/tree/main).

This PR increases concurrent requests from 5 to 40 for 8x throughput. The tests pass without giving any error (in regards to rate-limit)

### DIFF
--- a/packages/hub/src/lib/parse-safetensors-metadata.ts
+++ b/packages/hub/src/lib/parse-safetensors-metadata.ts
@@ -15,7 +15,7 @@ export const SAFETENSORS_INDEX_FILE = "model.safetensors.index.json";
 export const RE_SAFETENSORS_FILE = /\.safetensors$/;
 export const RE_SAFETENSORS_INDEX_FILE = /\.safetensors\.index\.json$/;
 export const RE_SAFETENSORS_SHARD_FILE = /\d{5}-of-\d{5}\.safetensors$/;
-const PARALLEL_DOWNLOADS = 5;
+const PARALLEL_DOWNLOADS = 40;
 const MAX_HEADER_LENGTH = 25_000_000;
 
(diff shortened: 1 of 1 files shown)


packages/hub/src/lib/parse-safetensors-metadata.ts
```diff
@@ -15,7 +15,7 @@ export const SAFETENSORS_INDEX_FILE = "model.safetensors.index.json";
 export const RE_SAFETENSORS_FILE = /\.safetensors$/;
 export const RE_SAFETENSORS_INDEX_FILE = /\.safetensors\.index\.json$/;
 export const RE_SAFETENSORS_SHARD_FILE = /\d{5}-of-\d{5}\.safetensors$/;
-const PARALLEL_DOWNLOADS = 5;
```
mishig25, line 18: was there any reasons why it was 5? this PR bumps it up to 40. Would we hit any rate-limit error? Also, all the requests are [http range requets](https://developer.mozilla.org/en-US/docs/Web/HTTP/Range_requests) for the header to get metadata

cc: @coyotte508 @SBrandeis

coyotte508 (approved):

julien-c (commented):
not opposed to it, but i hope it doesn't bring too many side effects

(i dont think the slowness on really large models is a huge deal, personally)

mishig25:
decreased the number from 40 to 20 https://github.com/huggingface/huggingface.js/pull/600/commits/3de884b110d811f2e02c2477d37f435cf1fea432 (just in case, being conservative) and merged this PR

PR #601, opened 2024-04-03
[gguf] RE_GGUF_SHARD_FILE

mishig25:
I need on the frontend to detect which gguf file is sharded or not

cc: @ggerganov @madgetr

### DIFF
--- a/packages/gguf/src/gguf.spec.ts
+++ b/packages/gguf/src/gguf.spec.ts
@@ -222,7 +222,7 @@ describe("gguf", () => {
 	});
 
 	it("should detect sharded gguf filename", async () => {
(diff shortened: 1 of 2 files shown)


madgetr:
> I need on the frontend to detect which gguf file is sharded or not
> 
> cc: @ggerganov @madgetr

Should these files be read as a signle model? If so, which file is the authority? gguf doesnt seem to use an index.json file.

julien-c:
the first file of a sharded weights contains the complete metadata i.e. tokenizer etc. (cc @phymbert)

madgetr:
> the first file of a sharded weights contains the complete metadata i.e. tokenizer etc. (cc @phymbert)

seems that the tensor shapes are scattered across the shards tho

PR #605, opened 2024-04-04
[safetensors] better RE_SAFETENSORS_SHARD_FILE

mishig25:
Follow up to https://github.com/huggingface/huggingface.js/pull/593

Two updates:
1. Extract shard numbers like in https://github.com/huggingface/huggingface.js/pull/601
2. Explicitly capture `-` or `_` character present at the beginning of shard substring

### DIFF
--- a/packages/hub/src/lib/parse-safetensors-metadata.ts
+++ b/packages/hub/src/lib/parse-safetensors-metadata.ts
@@ -14,7 +14,7 @@ export const SAFETENSORS_INDEX_FILE = "model.safetensors.index.json";
 /// but in some situations safetensors weights have different filenames.
 export const RE_SAFETENSORS_FILE = /\.safetensors$/;
 export const RE_SAFETENSORS_INDEX_FILE = /\.safetensors\.index\.json$/;
-export const RE_SAFETENSORS_SHARD_FILE = /\d{5}-of-\d{5}\.safetensors$/;
+export const RE_SAFETENSORS_SHARD_FILE = /[-_]?(\d{5})-of-(\d{5})\.safetensors$/;
 const PARALLEL_DOWNLOADS = 5;
 const MAX_HEADER_LENGTH = 25_000_000;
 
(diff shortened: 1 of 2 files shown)


packages/hub/src/lib/parse-safetensors-metadata.ts
```diff
@@ -14,7 +14,7 @@ export const SAFETENSORS_INDEX_FILE = "model.safetensors.index.json";
 /// but in some situations safetensors weights have different filenames.
 export const RE_SAFETENSORS_FILE = /\.safetensors$/;
 export const RE_SAFETENSORS_INDEX_FILE = /\.safetensors\.index\.json$/;
-export const RE_SAFETENSORS_SHARD_FILE = /\d{5}-of-\d{5}\.safetensors$/;
+export const RE_SAFETENSORS_SHARD_FILE = /[-_]?(\d{5})-of-(\d{5})\.safetensors$/;
```
coyotte508, line 17: I don't get `[-_]?`, since it's optional it doesn't do anything?

Maybe remove the `?`

mishig25: > I don't get [-_]?, since it's optional it doesn't do anything?

Screenshot 2024-04-05 at 10 34 38 src [here](https://huggingface.co/google/gemma-7b/blob/7aeedade2bfdf69adddb754cff0461e74541e436/model-00001-of-00004.safetensors)

on example above, I'd like to think that `-` is part of `-xxxxx-of-xxxxx.safetensors` rather than `model-`. The reason of this decision is that, if in the future, we wanna show common name of those tensors files it should be `model` rather than `model-` & with this regex we can achieve that by `str.slice(0, -str.match(RE_SAFETENSORS_SHARD_FILE).length)`

> Maybe remove the ?

`-` or `_` are optional. So we need to have `?`. For example there can be `model00002-of-00072.safetensors`

coyotte508: hm ok but that's a tenuous dependency between hub.js and the hub. It should be made an explicit function `extractShardTensorsData(filename)` that returns everything (the root, the numbers, ...) in a typed JSON.

Otherwise assumptions made in one codebase about another codebase can easily be broken, especially if there's no comment / warning that some things are there for a particular reason

julien-c (commented):
i think this one can be closed in favor of #622

packages/hub/src/lib/parse-safetensors-metadata.ts
julien-c: i don't think `[_-]` can really be optional, i've always seen either of them

mishig25:
Superceded by https://github.com/huggingface/huggingface.js/pull/622

PR #607, opened 2024-04-05
[widgets] LogIn Popover update

mishig25:
### Description

Two updates:
1. Close popover only on outside click. Otherwise, keep it open. (Currently on `main`, when mouse leaves the element, the popover disappears, & this PR changes this behaviour). Described in https://github.com/huggingface/huggingface.js/pull/567#issuecomment-2027145369

https://github.com/huggingface/huggingface.js/assets/11827707/1f750fb6-0929-48dd-b9a3-ff8fefa2d7da

2. Small screen fix

| ui bug    | this PR fix |
| -------- | ------- |
| Screenshot 2024-04-05 at 11 54 10  | Screenshot 2024-04-05 at 12 26 18    |

### DIFF
--- a/packages/widgets/src/lib/components/Popover/Popover.svelte
+++ b/packages/widgets/src/lib/components/Popover/Popover.svelte
@@ -1,7 +1,12 @@
+<!-- 
+	In https://github.com/huggingface/huggingface.js/pull/567, this component was changed significantly from the original one in moon/Popover.svelte.
+	Therefore, if/when widgets code goes back to moon, make sure not to delete this Popover.svelte right away before reconciling the differences.
+ -->
+
 <script lang="ts">
 	import { onMount, tick, createEventDispatcher } from "svelte";
 	import { fade } from "svelte/transition";
-	import { debounce, portalToBody } from "../../utils/ViewUtils.js";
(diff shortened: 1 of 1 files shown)


packages/widgets/src/lib/components/Popover/Popover.svelte
```diff
@@ -1,7 +1,12 @@
+<!-- 
+	In https://github.com/huggingface/huggingface.js/pull/567, this component was changed significantly from the original one in moon/Popover.svelte.
```
mishig25, line 2: btw, I tried using the in moon/Popover.svelte. However, wasn't able to make it work out of the box (i.e. there were issues on alignments & the popup was looking in unexpected behaviours) without changing the Popover.svelte codebase

krampstudio: Indeed, I quickly tried and there's a difference on how the anchor  and opening is handled. Here the anchor is outside the popover. 
I suggest to merge this PR as it is and refactor it when integrating back to moon

krampstudio (approved):

SBrandeis (approved):
lgtm thx

PR #622, opened 2024-04-11
[safetensors] add named groups to RE_SAFETENSORS_SHARD_FILE regex

mishig25:
add named groups to RE_SAFETENSORS_SHARD_FILE regex

safetensors equivalent of https://github.com/huggingface/huggingface.js/pull/621

### DIFF
--- a/packages/hub/src/lib/parse-safetensors-metadata.ts
+++ b/packages/hub/src/lib/parse-safetensors-metadata.ts
@@ -14,7 +14,7 @@ export const SAFETENSORS_INDEX_FILE = "model.safetensors.index.json";
 /// but in some situations safetensors weights have different filenames.
 export const RE_SAFETENSORS_FILE = /\.safetensors$/;
 export const RE_SAFETENSORS_INDEX_FILE = /\.safetensors\.index\.json$/;
-export const RE_SAFETENSORS_SHARD_FILE = /\d{5}-of-\d{5}\.safetensors$/;
+export const RE_SAFETENSORS_SHARD_FILE = /(?<shard>\d{5})-of-(?<total>\d{5})\.safetensors$/;
 const PARALLEL_DOWNLOADS = 20;
 const MAX_HEADER_LENGTH = 25_000_000;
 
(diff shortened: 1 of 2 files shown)


packages/hub/src/lib/parse-safetensors-metadata.ts
```diff
@@ -14,7 +14,7 @@ export const SAFETENSORS_INDEX_FILE = "model.safetensors.index.json";
 /// but in some situations safetensors weights have different filenames.
 export const RE_SAFETENSORS_FILE = /\.safetensors$/;
 export const RE_SAFETENSORS_INDEX_FILE = /\.safetensors\.index\.json$/;
-export const RE_SAFETENSORS_SHARD_FILE = /\d{5}-of-\d{5}\.safetensors$/;
+export const RE_SAFETENSORS_SHARD_FILE = /(?<shard>\d{5})-of-(?<total>\d{5})\.safetensors$/;
```
coyotte508, line 17: Might as well add the "prefix"? If you're planning on extracting it

julien-c: in that case add `[_-]` between the captured prefix and the shard number

(most models on the hub use a dash at this position (it's what's generated by transformers), but not `bigscience/bloom`)

julien-c (approved):

packages/hub/src/lib/parse-safetensors-metadata.ts
julien-c, line 17:
```suggestion
export const RE_SAFETENSORS_SHARD_FILE = /^(?<prefix>.*)[_-](?<shard>\d{5})-of-(?<total>\d{5})\.safetensors$/;
```

coyotte508:
```suggestion
export const RE_SAFETENSORS_SHARD_FILE = /^(?<prefix>.*?)[_-]?(?<shard>\d{5})-of-(?<total>\d{5})\.safetensors$/;
```

This is more accurate I think - it makes the `[-_]` optional but still tries to leave it out of `prefix`

mishig25: great idea !

julien-c: but that was my point, the `-_` is *not* optional

julien-c: we do want to enforce some level of convention

PR #623, opened 2024-04-11
Reorder quant-descriptions.ts

JorgeCepeda:
I ordered it by number and type as I saw fit

### DIFF
--- a/packages/gguf/src/quant-descriptions.ts
+++ b/packages/gguf/src/quant-descriptions.ts
@@ -9,13 +9,21 @@ export const GGUF_QUANT_DESCRIPTIONS: Record<GGMLQuantizationType, { txt: string
 		txt: "16-bit standard IEEE 754 half-precision floating-point number.",
 		src_url: "https://en.wikipedia.org/wiki/Half-precision_floating-point_format",
 	},
-	[GGMLQuantizationType.Q4_0]: {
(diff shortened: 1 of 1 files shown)


mishig25:
Thanks so much. Indeed, the ordering is cleaner now

julien-c:
i'm a quantization noob, but would an additional column of average `bits-per-weight` make sense? (And then the columns could be sortable with some JS maybe)

mishig25:
>  would an additional column of average bits-per-weight make sense? 

Totally. It would make sense to add column `bits-per-weight`

In this line
https://github.com/huggingface/huggingface.js/blob/f0e3eff1c301379cf9f9ac7183f32a3572a520af/packages/gguf/src/quant-descriptions.ts#L3-L4

turn `{ txt: string; src_url?: string }` to `{ txt: string; bpw: number, src_url?: string }`

and then publish those changes as a new column to 

https://github.com/huggingface/hub-docs/blob/20a6276348f347b3640a18fadcc142d5a90d20d6/docs/hub/gguf.md#L62-L64

julien-c:
hmm for the enum we should keep the order of the llama.cpp code no? and the enum is int-based anyways

JorgeCepeda:
To be honest I don't know about typescript so feel free to tackle it just in case I were to mess it up

PR #627, opened 2024-04-15
[safetensors] Fix RE_SAFETENSORS_SHARD_FILE

mishig25:
follow up to https://github.com/huggingface/huggingface.js/pull/622#discussion_r1562398324

> but that was my point, the -_ is not optional
> we do want to enforce some level of convention


coyotte508 (approved):
why not a typed function `fn(filename)` that returns all the data in a typed manner instead of the regex :see_no_evil: .

Currently in the next version of HF.js we can change the regex to something else, removing or renaming the groups, and on moon's side we'd have no clue. No compiler warning or anything.

PR #631, opened 2024-04-17
[gguf & st] parse shard filenames in typed function

mishig25:
follow up to https://github.com/huggingface/huggingface.js/pull/627#pullrequestreview-2000893065

> why not a typed function fn(filename) that returns all the data in a typed manner instead of the regex 🙈 .
> Currently in the next version of HF.js we can change the regex to something else, removing or renaming the groups, and on moon's side we'd have no clue. No compiler warning or anything.


coyotte508 (approved):
thanks

julien-c:
in most cases it's useful to wait a little more than 10 mins before merging 🥲

(LGTM)

PR #634, opened 2024-04-18
Fix fileDownloadInfo fetching entirely the file when there is no cred…

Kakulukian:
…entials param

### DIFF
--- a/packages/hub/src/lib/file-download-info.ts
+++ b/packages/hub/src/lib/file-download-info.ts
@@ -50,8 +50,8 @@ export async function fileDownloadInfo(params: {
 		method: "GET",
 		headers: {
 			...(params.credentials && {
-						Authorization: `Bearer ${params.credentials.accessToken}`,
-				  }),
+				Authorization: `Bearer ${params.credentials.accessToken}`,
+			}),
 			Range: "bytes=0-0",
 		},
 	});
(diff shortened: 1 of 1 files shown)


coyotte508 (approved):
Thank you

packages/hub/src/lib/file-download-info.ts
```diff
@@ -48,12 +48,14 @@ export async function fileDownloadInfo(params: {
...
-					Authorization: `Bearer ${params.credentials.accessToken}`,
-					Range: "bytes=0-0",
-			  }
-			: {},
+		headers: {
+			...(params.credentials
+				? {
+						Authorization: `Bearer ${params.credentials.accessToken}`,
+				  }
+				: undefined),
```
coyotte508, line 56:
```suggestion
			...(params.credentials && {
						Authorization: `Bearer ${params.credentials.accessToken}`,
				  }),
```

In this codebase you can do this if you want :p

Kakulukian: I like that :D

mishig25 (approved):
lgtm !

Kakulukian:
The tests are failing because we tell to retrieve one byte so Content-length will return 1 (follow the standard) so we may need to use head to get the file size

coyotte508 (approved):
nice!

PR #637, opened 2024-04-25
[gguf] get optional `paramCount`

mishig25:
Optionally get `paramCount` from parsing the gguf file so that we can use this information in hub-backend

### DIFF
--- a/packages/gguf/src/gguf.ts
+++ b/packages/gguf/src/gguf.ts
@@ -318,8 +318,8 @@ export async function gguf(
 			const PARALLEL_DOWNLOADS = 20;
 			const shardsParamCount = await promisesQueue(
 				urls.map((shardUrl) => async () => {
-					const { tensorInfos: shardTesnorInfos } = await gguf(shardUrl);
-					const shardParamCount = shardTesnorInfos
+					const { tensorInfos: shardTensorInfos } = await gguf(shardUrl);
+					const shardParamCount = shardTensorInfos
 						.map(({ shape }) => shape.reduce((acc, val) => acc * Number(val), 1))
 						.reduce((acc, val) => acc + val, 0);
 					return shardParamCount;
--- a/packages/gguf/src/gguf.ts
+++ b/packages/gguf/src/gguf.ts
@@ -15,7 +15,7 @@ export interface GgufShardFileInfo {
 	total: string;
 }
 
-export function parseGgufShardFile(filename: string): GgufShardFileInfo | null {
+export function parseGgufShardFilename(filename: string): GgufShardFileInfo | null {
 	const match = RE_GGUF_SHARD_FILE.exec(filename);
 	if (match && match.groups) {
 		return {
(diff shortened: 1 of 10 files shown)


packages/gguf/src/gguf.ts
```diff
@@ -293,5 +295,41 @@ export async function gguf(
...
+				if (shardIdx !== shard) {
+					urls.push(`${prefix}-${shardIdx.toString().padStart(5, "0")}-of-${total.toString().padStart(5, "0")}.gguf`);
+				}
+			}
+
+			const PARALLEL_DOWNLOADS = 20;
+			const shardsParamCount = await promisesQueue(
+				urls.map((shardUrl) => async () => {
+					const { tensorInfos: shardTesnorInfos } = await gguf(shardUrl);
+					const shardParamCount = shardTesnorInfos
```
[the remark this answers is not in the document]
mishig25: handled in 6be0dc120973a5942847cd0ee90bae1da468b01b

ngxson:
I additionally reviewed this PR. It's LGTM overall.

One small suggestion is that maybe we should abstract/generalize the support for shared gguf (while using the correct metadata) in the future. If you're OK with this idea, I'll have a look on it.

packages/gguf/src/gguf.ts
```diff
@@ -183,6 +184,7 @@ export async function gguf(
 		 * Custom fetch function to use instead of the default one, for example to use a proxy or edit headers.
 		 */
 		fetch?: typeof fetch;
+		calculateParamCount?: boolean;
```
julien-c, line 187: why not the same parameter name as for safetensors (`computeParametersCount`)?

packages/gguf/src/types.ts
```diff
@@ -138,4 +138,5 @@ export interface GGUFTensorInfo {
 export interface GGUFParseOutput {
 	metadata: GGUFMetadata;
 	tensorInfos: GGUFTensorInfo[];
+	paramCount?: number;
```
julien-c, line 141: why not the same prop name as for safetensors (`parameterCount`)? Besides i don't think this should be in this type (which is strictly what is in GGUF), but just in the function return type

packages/gguf/src/gguf.ts
```diff
@@ -293,5 +295,41 @@ export async function gguf(
 		});
 	}
 
-	return { metadata, tensorInfos };
+	let paramCount: number | undefined = undefined;
+
+	if (params?.calculateParamCount) {
+		paramCount = tensorInfos
+			.map(({ shape }) => shape.reduce((acc, val) => acc * Number(val), 1))
+			.reduce((acc, val) => acc + val, 0);
```
julien-c, line 303: i agree with this part, but shouldn't we return a `Record<Dtype, number>` like in safetensors? (no strong opinion, I'm ok with only returning a number, that's fine)

packages/gguf/src/gguf.ts
```diff
@@ -293,5 +295,41 @@ export async function gguf(
...
+
+		const ggufShardFileInfo = parseGgufShardFile(url);
+		if (ggufShardFileInfo) {
```
julien-c, line 306:
```suggestion
		if (ggufShardFileInfo && params?.fetchAllShards) {
```

IMO this behavior (fetching other shards magically) is quite unintuitive, so I would at least add a special flag for it.

Or maybe better API-wise to split this into another method

julien-c (changes requested):

julien-c (approved):

PR #638, opened 2024-04-26
gguf: Fix ArrayBuffer.resize not supported on Firefox

ngxson:
`ArrayBuffer.resize()` is not supported on Firefox. This causes the error while visualizing gguf:

image

This PR introduces a small patch that uses a polyfill in case `ArrayBuffer.resize()` cannot be used.

Tested on Firefox and Chrome.

### DIFF
--- a/packages/gguf/src/gguf.ts
+++ b/packages/gguf/src/gguf.ts
@@ -52,7 +52,7 @@ class RangeView {
 	private chunk: number;
 	private buffer: ArrayBuffer;
 	private dataView: DataView;
-	
+
 	get view(): DataView {
 		return this.dataView;
 	}
@@ -91,7 +91,7 @@ class RangeView {
 		this.chunk += 1;
 	}
 	/**
-	 * Append new data into the buffer 
+	 * Append new data into the buffer
 	 */
 	appendBuffer(buf: Uint8Array) {
 		/// TODO(fix typing)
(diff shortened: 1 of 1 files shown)


julien-c:
(though i wish they would support it soon?)

ngxson:
According to [caniuse page](https://caniuse.com/?search=ArrayBuffer.resize), the support is experimental on version 124 (current version) of firefox (can be enabled via a flag). So I think it's expected to be stable on 125. Unfortunate I couldn't find the issue tracker anywhere so I can't confirm that information.

In any way, I feel like this API is relative new on all other browsers, so it's maybe better to keep this polyfill for a while.

julien-c:
let's merge @mishig25 and set a reminder to remove (in a couple of weeks?) when Firefox support is fully out?

mishig25:
Merged it. Some tests were failing citing a token issue (likely because the PR comes from a fork)

>  set a reminder to remove (in a couple of weeks?) when Firefox support is fully out?

https://github.com/huggingface/huggingface.js/issues/642

mishig25:
@ngxson thanks a lot for the contribution! I've tweeted about it [here](https://twitter.com/mishig25/status/1784945423104245763)

ngxson:
> @ngxson thanks a lot for the contribution! I've tweeted about it [here](https://twitter.com/mishig25/status/1784945423104245763)

Thanks! Great to see that it works on Firefox now :smile: 

Btw my twitter is also @ngxson

PR #640, opened 2024-04-27
gguf: Add types for LLM architectures

### LINKED ISSUE #566: [gguf types] Add missing types & make existing types stronger
https://github.com/huggingface/huggingface.js/pull/562 adds initial typing to @huggingface/gguf package.

However, not all architectures are added in #562. (see [here](https://github.com/ggerganov/ggml/blob/master/docs/gguf.md#required) & [here](https://github.com/huggingface/huggingface.js/pull/562#discussion_r1531981246)). For example, one can add [mamba](https://github.com/ggerganov/ggml/blob/master/docs/gguf.md#mamba) & type mamba specific fields (i.e. `mamba.ssm.inner_size`. See RWKV-specific typing for inspiration https://github.com/huggingface/huggingface.js/blob/81e6ce19d119978ce827c3f166d601c8f24fe454/packages/gguf/src/types.ts#L117

ngxson:
Resolve #566

This PR introduces a generator script that pulls `llama.cpp` file, then use regex to extract needed information.

It also addresses some smaller issues:
- Some places (for example `Attention`, `Rope`, `MOE`,...) use type intersection while it should be union
- `TokenizerModel` should be limited to 4 models, see `llm_load_vocab()`: https://github.com/ggerganov/llama.cpp/blob/928e0b7013c862cf10701957b3d654aa70f11bd8/llama.cpp#L4198

While it's fun to have the script & these type definitions, I think it would be nice to have some clearer directions on how to use them in the future.

Demo:

image


packages/gguf/scripts/generate-llm.ts
```diff
@@ -0,0 +1,179 @@
+/**
+ * Script for generating llm.ts
+ * The source data is taken from llama.cpp
+ */
+
+import { writeFileSync } from "node:fs";
+
+const SOURCE_CPP_URL = "https://raw.githubusercontent.com/ggerganov/llama.cpp/master/llama.cpp";
```
julien-c, line 8: i was basing myself on the `gguf-py/gguf/constants.py` file (i.e. from the Python parser), but I guess the cpp code is at least as good a source of truth?

julien-c (commented):
very cool work @ngxson

I've launched the CI run, let me know if any help is needed, LGTM otherwise

packages/gguf/scripts/generate-llm.ts
```diff
@@ -0,0 +1,179 @@
...
+};
+
+async function main() {
+	const res = await fetch(SOURCE_CPP_URL);
+	const cppSource = await res.text();
+
+	/////////////////////////////////////
+	// extract list of all architectures
+	const archList: Arch[] = [];
+	const matchedArchList = cppSource.match(/LLM_ARCH_NAMES = ([^;]+)/)!![1].split("\n");
```
julien-c, line 67:
```suggestion
	const matchedArchList = cppSource.match(/LLM_ARCH_NAMES = ([^;]+)/)?.[1].split("\n") ?? [];
```

julien-c: (and similar linting nits)

mishig25:
@ngxson locally you can check the errors since the CI uses same commands:
```
pnpm format
pnpm lint
```

packages/gguf/package.json
```diff
@@ -32,6 +32,7 @@
 		"format:check": "prettier --check .",
 		"prepublishOnly": "pnpm run build",
 		"build": "tsup src/index.ts --format cjs,esm --clean --dts",
```
mishig25, line 34:
```suggestion
		"build": "pnpm build:llm && tsup src/index.ts --format cjs,esm --clean --dts",
```

`build:llm` should be called here before build right? (cc: @coyotte508)

packages/gguf/src/transformer-llm.ts
```diff
@@ -0,0 +1,175 @@
+import { ModelBase } from "./types";
```
mishig25, line 1: since it is a generated file, should we add it into the .gitignore ?

packages/gguf/scripts/generate-llm.ts
```diff
@@ -0,0 +1,179 @@
...
+	for (const line of matchedArchList) {
+		const matched = line.match(/(LLM_ARCH_[A-Z0-9_]+),\s+"([^"]+)"/);
```
mishig25, line 69:
```suggestion
		const matched = line.match(/(LLM_ARCH_[A-Z0-9_]+),\s+"(.+?)"/);
```

I think you can use the [greedy/lazy regex quantifiers](https://javascript.info/regexp-greedy-and-lazy). Let me know if this change works. If so, uou can replace all instances of `"([^"]+)"` with `"(.+?)"`

packages/gguf/scripts/generate-llm.ts
```diff
@@ -0,0 +1,179 @@
...
+	for (const line of cppSource.split("\n")) {
+		// check if current line is function llm_load_hparams()
+		if (line.startsWith("static void llm_load_hparams")) {
+			insideLoadHParamsFn = true;
+		}
+		if (!insideLoadHParamsFn) {
+			continue;
+		}
+		// check if current line has LLM_ARCH_*
+		const cppConst = (line.match(/case (LLM_ARCH_[A-Z0-9_]+)/) || [])[1];
```
mishig25, line 126: `LLM_ARCH_[A-Z0-9_]+` part is also used on line 100 https://github.com/ngxson/huggingface.js/blob/5313453c75b3f72fd1381762aa6450d33f77749e/packages/gguf/scripts/generate-llm.ts#L100

therefore, we should declare this regex as const `cont RE_XYZ = /LLM_ARCH_[A-Z0-9_]+)/` & re-use this re in this line like `new Regexp(/case ${RE_XYZ}/)`

packages/gguf/scripts/generate-llm.ts
```diff
@@ -0,0 +1,179 @@
...
+	/////////////////////////////////////
+	// extract list of all architectures
+	const archList: Arch[] = [];
+	const matchedArchList = cppSource.match(/LLM_ARCH_NAMES = ([^;]+)/)!![1].split("\n");
+	for (const line of matchedArchList) {
+		const matched = line.match(/(LLM_ARCH_[A-Z0-9_]+),\s+"([^"]+)"/);
+		if (matched && !matched[0].match(/unknown/)) {
+			archList.push({
+				cppConst: matched[1],
+				name: matched[2],
```
mishig25, line 73: to make the regexes even more readable, you can use [named capturing group](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Regular_expressions/Named_capturing_group) & use like `matched.groups.name rather than matched[2]`

packages/gguf/package.json
julien-c: or maybe run it once a week from a GitHub Action or something?

coyotte508: It fetches a remote file so yes - calling it with a cron job is better than at every install

ngxson: I prefer to have it run manually, because this way we can verify the generated script before pushing. But if we can run a Github Action workflow and create new PR if there're changes, it's also a good idea.

packages/gguf/scripts/generate-llm.ts
ngxson: I'm not very familiar with `gguf-py`, but seem like only in cpp code that you can find the list of metadata used by model (in python library, there's only the list of tensors, not list of kv - those can be found in `convert*.py` scripts)

packages/gguf/scripts/generate-llm.ts
ngxson: Actually I used these unsafe casting because I'd rather to have the script failed, instead of having it generates bad output. But I'm OK with this proposal, I should have added checks and throw errors. I'll fix this :+1:

packages/gguf/scripts/generate-llm.ts
julien-c: yes that makes sense!

packages/gguf/scripts/generate-llm.ts
```diff
@@ -0,0 +1,179 @@
...
+			"};",
+		]).flat(3),
+		"",
+		`export type TransformerLLM = ${archList.map(a => a.tsName).join(' | ')};`
+	].join("\n");
+
+	writeFileSync(DEST_FILE_PATH, content);
+};
+
+function snakeToPascal(str: string) {
```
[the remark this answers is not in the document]
ngxson: This function is only used by the generator script, which should not be part of the distributed (transpiled) source code. Do you think it's still useful to move it into utils?

packages/gguf/scripts/generate-llm.ts
ngxson: Thanks for pointing me in the correct direction. I replaced all instances of `"([^"]+)"` with `"(.+?)"`. However, the same technique does not work with `([^;]+)` ==> `(.+?);` (matches `ABC;` for example). Please let me know if you have any idea. Thanks.

packages/gguf/scripts/generate-llm.ts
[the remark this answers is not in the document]
mishig25: good for now

ngxson:
@julien-c @mishig25 Regarding how to actually use these types, I'm thinking about having a list of functions like this:

```ts
getArchLlama(metadata: GGUFMetadata): ArchLlama?
getArchFalcon(metadata: GGUFMetadata): ArchFalcon?
...
```

Which returns `undefined` if the model is not the one being checked.

The idea is to prevent user from doing something like `const arch = model as ArchLlama`, but I'm not sure if my proposal is the right way to do.

Do you have any other ideas? Thanks.

packages/gguf/scripts/generate-llm.ts
```diff
@@ -0,0 +1,194 @@
...
+		if (line === "}") {
+			break;
+		}
+	}
+
+	/////////////////////////////////////
+	// write result to file
+	const content = [
+		'import type { ModelBase } from "./types";',
+		"// This file is auto-generated by generate-llm.ts",
```
mishig25, line 160: why not put this comment line on top of the file?

packages/gguf/scripts/generate-llm.ts
```diff
@@ -0,0 +1,194 @@
...
+	& Attention<TArchitecture>
+	& Rope<TArchitecture>;
+`;
+
+const KV_TYPE = {
+	LLM_KV_ATTENTION_LAYERNORM_RMS_EPS: "number",
+	LLM_KV_ATTENTION_LAYERNORM_EPS: "number",
+	LLM_KV_ATTENTION_CAUSAL: "boolean",
+	LLM_KV_TOKENIZER_TOKEN_TYPE_COUNT: "number",
+	LLM_KV_POOLING_TYPE: "number", // TODO: this is an enum
```
mishig25, line 42: is this todo being handled in this PR or follow-up PR ?

packages/gguf/scripts/generate-llm.ts
```diff
@@ -0,0 +1,194 @@
...
+async function main() {
+	const res = await fetch(SOURCE_CPP_URL);
+	const cppSource = await res.text();
+
+	/////////////////////////////////////
+	// extract list of all architectures
+	const archList: Arch[] = [];
+	const RE_ARCH_NAME = /LLM_ARCH_[A-Z0-9_]+/;
+	const matchedArchList = cppSource.match(/LLM_ARCH_NAMES = (?<names>[^;]+)/)?.groups?.names.split("\n") || [];
+	if (matchedArchList.length === 0) {
```
mishig25, line 69:
```suggestion
	const matchedArchList = cppSource.match(/LLM_ARCH_NAMES = (?<names>[^;]+)/)?.groups?.names.split("\n");
	if (!matchedArchList?.length) {
```

packages/gguf/scripts/generate-llm.ts
ngxson: I usually place import on the top then the comment. Just personal preference though, but you're right, it should be on top of the file.

packages/gguf/scripts/generate-llm.ts
```diff
@@ -0,0 +1,194 @@
...
+		throw new Error("LLM_ARCH_NAMES is empty");
+	}
+	for (const line of matchedArchList) {
+		const matched = line.match(/(?<cppConst>LLM_ARCH_[A-Z0-9_]+),\s+"(?<name>.+?)"/);
+		if (matched && !matched.groups?.name.match(/unknown/)) {
+			archList.push({
+				cppConst: matched.groups?.cppConst || "",
+				name: matched.groups?.name || "",
+				tsName: snakeToPascal(matched[1].replace("LLM_", "")),
```
mishig25, line 78:
```suggestion
				tsName: snakeToPascal(matched.groups?.cppConst.replace("LLM_", "")),
```

packages/gguf/scripts/generate-llm.ts
```diff
@@ -0,0 +1,194 @@
...
+	// for example: LLM_KV_ATTENTION_LAYERNORM_RMS_EPS ==> "%s.attention.layer_norm_rms_epsilon"
+	const constToKVName: { [cppConst: string]: string } = {};
+	const matchedKVList = cppSource.match(/LLM_KV_NAMES = (?<names>[^;]+)/)?.groups?.names.split("\n") || [];
+	if (matchedKVList.length === 0) {
+		throw new Error("LLM_KV_NAMES is empty");
+	}
+	for (const line of matchedKVList) {
+		const matched = line.match(/(LLM_KV_[A-Z0-9_]+)[,\s]+"([^"]+)"/);
+		if (matched) {
+			constToKVName[matched[1]] = matched[2];
```
mishig25, line 96: as in other regexs, let's use named groups here as well

packages/gguf/scripts/generate-llm.ts
```diff
@@ -0,0 +1,194 @@
...
+		}
+	}
+
+	/////////////////////////////////////
+	// extract list of tensor names based on architecture
+	// TODO: unused for now
+	const matchedTensorList = cppSource.match(/LLM_TENSOR_NAMES = (?<names>[^;]+)/)?.groups?.names.split("\n") || [];
+	if (matchedTensorList.length === 0) {
```
mishig25, line 104: similar change is needed as in https://github.com/huggingface/huggingface.js/pull/640#discussion_r1583785883

packages/gguf/scripts/generate-llm.ts
```diff
@@ -0,0 +1,194 @@
...
+		throw new Error("LLM_TENSOR_NAMES is empty");
+	}
+	let currCppConst = "";
+	for (const line of matchedTensorList) {
+		// check if current line has LLM_ARCH_*
+		const cppConst = (line.match(RE_ARCH_NAME) || [])[0];
```
mishig25, line 110:
```suggestion
		const cppConst = line.match(RE_ARCH_NAME)?.[0];
```

packages/gguf/scripts/generate-llm.ts
```diff
@@ -0,0 +1,194 @@
...
+		// check if current line is function llm_load_hparams()
+		if (line.startsWith("static void llm_load_hparams")) {
+			insideLoadHParamsFn = true;
+		}
+		if (!insideLoadHParamsFn) {
+			continue;
+		}
+		// check if current line has LLM_ARCH_*
+		const RE_CASE = new RegExp(`case (${RE_ARCH_NAME.source})`);
+		const cppConst = (line.match(RE_CASE) || [])[1];
```
mishig25, line 137:
```suggestion
		const cppConst = line.match(RE_CASE)?.[1];
```

packages/gguf/scripts/generate-llm.ts
```diff
@@ -0,0 +1,194 @@
...
+		if (cppConst) {
+			currCppConst = cppConst;
+			continue;
+		}
+		// check if current line has get_key(...)
+		const keyConst = (line.match(/LLM_KV_[A-Z0-9_]+/) || [])[0];
```
mishig25, line 143:
```suggestion
		const keyConst = line.match(/LLM_KV_[A-Z0-9_]+/)?.[0];
```

packages/gguf/scripts/generate-llm.ts
```diff
@@ -0,0 +1,194 @@
...
+		"// This file is auto-generated by generate-llm.ts",
+		"",
+		"export const LLM_ARCHITECTURES = [",
+		...archList.map((a) => `\t${JSON.stringify(a.name)},`),
+		"] as const;",
+		"type LLMArchitecture = (typeof LLM_ARCHITECTURES)[number];",
+		DEST_COMMON_SOURCE,
+		...archList.map((a) => {
+			let code = `export type ${a.tsName} = TransformerLLMBase<${JSON.stringify(a.name)}>`;
+			if (a.hparams.length > 0) {
```
mishig25, line 169:
```suggestion
			if (a.hparams.length) {
```

packages/gguf/scripts/generate-llm.ts
```diff
@@ -0,0 +1,194 @@
...
+	const matchedArchList = cppSource.match(/LLM_ARCH_NAMES = (?<names>[^;]+)/)?.groups?.names.split("\n") || [];
+	if (matchedArchList.length === 0) {
+		throw new Error("LLM_ARCH_NAMES is empty");
+	}
+	for (const line of matchedArchList) {
+		const matched = line.match(/(?<cppConst>LLM_ARCH_[A-Z0-9_]+),\s+"(?<name>.+?)"/);
+		if (matched && !matched.groups?.name.match(/unknown/)) {
+			archList.push({
+				cppConst: matched.groups?.cppConst || "",
+				name: matched.groups?.name || "",
```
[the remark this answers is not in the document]
ngxson: Good point! There should be no more `||` or `??` in my code: https://github.com/huggingface/huggingface.js/pull/640/commits/e9f076626ed532199b2ed509f15128816747218a

packages/gguf/src/types.ts
```diff
@@ -1,3 +1,7 @@
+import type { TransformerLLM } from "./transformer-llm";
+import { LLM_ARCHITECTURES } from "./transformer-llm";
+export * from "./transformer-llm";
```
mishig25, line 3:
```suggestion
```

this import is not needed

packages/gguf/scripts/generate-llm.ts
mishig25: handled in 6962b5143a70631f5a2736b53ba1ddbdc8d609f0

packages/gguf/scripts/generate-llm.ts
mishig25: handled in 6962b5143a70631f5a2736b53ba1ddbdc8d609f0

packages/gguf/scripts/generate-llm.ts
mishig25: handled in 6962b5143a70631f5a2736b53ba1ddbdc8d609f0

packages/gguf/scripts/generate-llm.ts
mishig25: handled in 6962b5143a70631f5a2736b53ba1ddbdc8d609f0

packages/gguf/scripts/generate-llm.ts
mishig25: handled in 6962b5143a70631f5a2736b53ba1ddbdc8d609f0

packages/gguf/scripts/generate-llm.ts
mishig25: handled in 6962b5143a70631f5a2736b53ba1ddbdc8d609f0

packages/gguf/scripts/generate-llm.ts
mishig25: handled in https://github.com/huggingface/huggingface.js/commit/6962b5143a70631f5a2736b53ba1ddbdc8d609f0

packages/gguf/scripts/generate-llm.ts
mishig25: handled in https://github.com/huggingface/huggingface.js/commit/6962b5143a70631f5a2736b53ba1ddbdc8d609f0

packages/gguf/scripts/generate-llm.ts
mishig25: handled in e9f076626ed532199b2ed509f15128816747218a

packages/gguf/scripts/generate-llm.ts
mishig25: handled in 6962b5143a70631f5a2736b53ba1ddbdc8d609f0

packages/gguf/scripts/generate-llm.ts
mishig25: `([^;]+)` is fine as it is

mishig25:
[the proposal](https://github.com/huggingface/huggingface.js/pull/640#issuecomment-2083666339) sounds good to me. 

> but I'm not sure if my proposal is the right way to do

You can start with a draft PR and we can iterate if needed

mishig25 (approved):
If you can write a small section on [readme](https://github.com/ngxson/huggingface.js/blob/e9f076626ed532199b2ed509f15128816747218a/packages/gguf/README.md) about this change, that would be great.

I've left one last nit here https://github.com/huggingface/huggingface.js/pull/640#discussion_r1584348110

LGTM! Unless @julien-c comments in the next few hours, I will merge this PR soon

packages/gguf/src/types.ts
ngxson: Removed with https://github.com/huggingface/huggingface.js/pull/640/commits/09b70cda0654d48005b783bae92369f81a11783c

ngxson:
Thanks for the careful review. For now it's not very clear how to use these types, so I don't think I can add anything to readme right now. Maybe with the next PR (the proposal that you mentioned), I will add an example to readme.

packages/gguf/src/types.ts
```diff
@@ -44,19 +47,7 @@ export enum GGUFValueType {
...
-	"gptj",
-	"gpt2",
-	"bloom",
-	"falcon",
-	"gemma",
-	"rwkv",
-	"whisper",
-] as const;
-
+const ARCHITECTURES = [...LLM_ARCHITECTURES, "rwkv", "whisper"];
```
julien-c, line 50:
```suggestion
const ARCHITECTURES = [...LLM_ARCHITECTURES, "rwkv", "whisper"] as const;
```

I think you need to add this here otherwise the type inference is `string[]` – unless we're not using the same version of typescript?

julien-c (approved):
re. your proposal, please note that I am not super good with TS type inference, but IMO the ideal way we would use this is to enable like this:

```ts
if (metadata["general.architecture"] === "grok") {
	const x = metadata; // type = TransformerLLMBase<"grok">
	const foo = x["grok.attention.alibi_bias_max"]; // type = number
}
```

do you think it's possible at all?

packages/gguf/src/types.ts
```diff
@@ -66,39 +57,18 @@ interface General {
...
-	| { [K in `${TArchitecture}.rope.freq_base`]: number }
-	| { [K in `${TArchitecture}.rope.scale`]: number }
-	| { [K in `${TArchitecture}.rope.scale_linear`]: number };
-
-type ModelBase<
+export type ModelBase<
 	TArchitecture extends
 		| Architecture
 		| `encoder.${Extract<Architecture, "whisper">}`
 		| `decoder.${Extract<Architecture, "whisper">}`,
```
coyotte508, line 64: @mishig25 is this on purpose? 

`TArchitecture` will always extend `"encoder.whisper" | "decoder.whisper"`

packages/gguf/src/types.ts
```diff
@@ -66,39 +57,18 @@ interface General {
...
-	| { [K in `${TArchitecture}.context_length`]: number }
-	| { [K in `${TArchitecture}.embedding_length`]: number }
-	| { [K in `${TArchitecture}.block_count`]: number };
-
-type MOE<TArchitecture extends Architecture> =
-	| { [K in `${TArchitecture}.expert_count`]: number }
-	| { [K in `${TArchitecture}.expert_used_count`]: number };
+> = { [K in `${TArchitecture}.layer_count`]: number } & { [K in `${TArchitecture}.feed_forward_length`]: number } & {
+	[K in `${TArchitecture}.context_length`]: number;
+} & { [K in `${TArchitecture}.embedding_length`]: number } & { [K in `${TArchitecture}.block_count`]: number };
```
coyotte508, line 67: It's maybe more readable to have:

```suggestion
Record<
	| `${TArchitecture}.layer_count`
	| `${TArchitecture}.feed_forward_length`
	| `${TArchitecture}.context_length`
	| `${TArchitecture}.embedding_length`
	| `${TArchitecture}.block_count`,
	number
>
```

coyotte508:
> image
> 
> checked it manually. As of right now, ts server is just giving generic `MetadatValue` rather than specific type (e.g. `number`)
> 
> cc: @coyotte508 @ngxson do you know if above request is possible [#640 (comment)](https://github.com/huggingface/huggingface.js/pull/640#issuecomment-2085961578) ?

You can do this:

```ts
const x: TransformerLLM = {} as any;

if ("grok.layer_count" in x) {
	// below is typed correctly
	console.log(x["grok.attention.layer_norm_epsilon"])
}
```



> maybe by doing some generics stuff in
> 
> ```ts
> export type GGUFMetadata<K extends Architecture = Architecture> = {
> 	version: Version;
> 	tensor_count: bigint;
> 	kv_count: bigint;
> } & Partial<General<K>> &
> 	Partial<ModelBase<K>>;
> ```
> 
> maybe too complex though

Yes it's possible probably

You would need a mapped type:

```ts
type Archs = {
  "grok": ArchGrok,
  ...
} satisfies Record<Architecture, unknown>
```

And then call `Archs[T]` somewhere in you `Model<T>` or `GGufMetaData<T>` type.

Since we have code generation it's also possible to generate every possible `GGufMetadata` and do `GGufMetadata = GgufMetadaGrok | ...`. And have less complex generics code (but more verbose)

coyotte508 (approved):
LGTM, although would appreciate https://github.com/huggingface/huggingface.js/pull/640#discussion_r1585321087 :)

packages/gguf/src/types.ts
ngxson: Actually it's being formatted by `pnpm run format`. Can we maybe ignore specific lines from the formatter?

packages/gguf/src/types.ts
ngxson: I couldn't find any whisper gguf model on the internet, so maybe we should temporary remove support for whisper and we'll see later?

More info: there're some [ggml models uploaded by ggerganov](https://huggingface.co/ggerganov/whisper.cpp/tree/main), but they uses ggml format, not gguf.

packages/gguf/src/types.ts
julien-c: maybe we merge this one and improve in later PRs?

ngxson: Yeah I think we can merge now. I'm not having much time for doing code atm, but I'll open another pull to resolve this and the also the proposal related to usage of these types (hopefully I'll have time this weekend).

packages/gguf/src/types.ts
mishig25: the whisper spec of gguf is here https://github.com/ggerganov/ggml/blob/master/docs/gguf.md#whisper

mishig25: > @mishig25 is this on purpose?

yes

mishig25:
Let's merge ?

mishig25:
@ngxson thanks a lot for the contribution ! 
looking forward to your future contrib 🔥 https://github.com/huggingface/huggingface.js/pull/640#discussion_r1587394771

Issue #625, opened 2024-04-12
[Widget] "Model not loaded yet" error on page load


SBrandeis:
Ctx: https://huggingface.slack.com/archives/C02EMARJ65P/p1712930061968009

cc @OlivierDehaene @mishig25

PR #641, opened 2024-04-29
[widget] Suppress errors on onMount example load calls

mishig25:
Closes https://github.com/huggingface/huggingface.js/issues/625

### Description

The widget behaviour is: if a widget has an example(s), then onMount, pick a random example & run inference on it. However, it was causing not a nice UX like https://github.com/huggingface/huggingface.js/issues/625 (onMount inference call was failing with "Model is not ready" & since it is a onMount call, user was being welcomed with a page that has error). To fix this UI/UX issue, error is being suppressed on onMount inference calls (isOnLoadCall flagged)


coyotte508 (approved):

gary149 (approved):
we'll test in prod :)

julien-c (approved):

SBrandeis (approved):

mishig25:
working on one more related fix/PR and will merge botj

PR #645, opened 2024-05-01
Add chat completion method

radames:
Supersede #581. Thanks to @Wauplin, I can import the types from "@huggingface/tasks"
I've followed the pattern for `textGeneration` and `textGenerationStream`.


radames:
One question @coyotte508 , do we want to abstract the `/v1/chat/completions` endpoint if the user uses `chatCompletion` with our hosted models, let's say `mistralai/Mistral-7B-Instruct-v0.2`, shall this also work?

```
const stream = hf.chatCompletionStream({
				model: "mistralai/Mistral-7B-Instruct-v0.2",
				messages: [{ role: "user", content: "Complete the equation 1+1= ,just the answer" }],
				max_tokens: 500,
				return_full_text: false,
				temperature: 0.1,
				seed: 0,
			});
```

packages/inference/src/tasks/nlp/chatCompletion.ts
```diff
@@ -0,0 +1,31 @@
+import { InferenceOutputError } from "../../lib/InferenceOutputError";
+import type { BaseArgs, Options } from "../../types";
+import { request } from "../custom/request";
+import type { ChatCompletionInput, ChatCompletionOutput } from "@huggingface/tasks";
```
coyotte508, line 4: This will cause issues when importing the lib - see https://github.com/huggingface/huggingface.js/issues/584

We either need to solve https://github.com/huggingface/huggingface.js/issues/584 not import the types from another package

radames: yes, maybe we could copy the types for now, since this method is need soon? 
we only need `ChatCompletionInput, ChatCompletionStreamOutput, ChatCompletionOutput` wdyt?

Wauplin:
> The only issue is now all models are served/compatible with /v1/chat/completions we might ended up on the same

@radames not quite exactly. TGI-served models expose a `/v1/chat/completions` route which is fully compatible with the chat completion API. `transformers`-served models on the other hand don't have a `/v1/chat/completions` route but they still accept to receive a list of messages as input in their `text-generation` pipeline. In this case messages are rendered server-side which makes figure from https://github.com/huggingface/huggingface_hub/pull/2094#issue-2173830914 obsolete. I made the change in https://github.com/huggingface/huggingface_hub/pull/2258 if you are interested to look at it. In particular you'll see that the transformers response is simply a text string since it is still the `text-generation` pipeline. To return a proper `ChatCompletionOutput` object, you'll have to populate fields manually (in my case a lot of fields are set to `"dummy"`). Hope this will help, please let me know if you have some remaining questions.

Wauplin:
Oh and btw, TGI-served models have a `/info` route exposed (see https://api-inference.huggingface.co/models/meta-llama/Meta-Llama-3-70B-Instruct/info). With this endpoint you can be 100% sure whether a model is TGI or transformers backed. In the current `huggingface_hub` implementation we rely on errors to "guess/infer" if a model is not TGI-served but don't do this :grimacing: I'll open a PR to fix this on `hfh`-side as well.

packages/inference/src/tasks/nlp/chatCompletion.ts
coyotte508: I'll look at it today

coyotte508: https://github.com/huggingface/huggingface.js/pull/651

packages/inference/test/HfInference.spec.ts
```diff
@@ -645,11 +639,89 @@ describe.concurrent(
...
+				}
+			}
+			expect(out).toContain("2");
+		});
+		it("custom mistral - OpenAI Specs", async () => {
+			const MISTRAL_KEY = env.MISTRAL_KEY;
+			if (!MISTRAL_KEY) {
+				console.warn("Skipping test because MISTRAL_KEY is not set");
+				return;
+			}
```
coyotte508, line 691: @radames can you remove those conditional skips?

You can do `VCR_MODE=cache` in the env before running the test to record the result in `tapes.json`.

radames: done

radames:
> > The only issue is now all models are served/compatible with /v1/chat/completions we might ended up on the same
> 
> @radames not quite exactly. TGI-served models expose a `/v1/chat/completions` route which is fully compatible with the chat completion API. `transformers`-served models on the other hand don't have a `/v1/chat/completions` route but they 

thanks @Wauplin, yes I miswrote my sentence, and I wanted to say that not all models are served with `mixedbread-ai/mxbai-embed-large-v1` exactly your point! 

and I think we should copy this logic to complete the model url, in case ones provide only a model id 
```python 
           model_url = self._resolve_url(model)
            if not model_url.endswith("/chat/completions"):
                model_url += "/v1/chat/completions"
```

packages/inference/src/tasks/nlp/chatCompletionStream.ts
```diff
@@ -0,0 +1,17 @@
...
+ * Use to continue text from a prompt. Same as `textGeneration` but returns generator that can be read one token at a time
+ */
+export async function* chatCompletionStream(
+	args: BaseArgs & ChatCompletionInput,
+	options?: Options
+): AsyncGenerator<ChatCompletionStreamOutput> {
+	yield* streamingRequest<ChatCompletionStreamOutput>(args, {
+		...options,
+		taskHint: "text-generation",
+		chatCompletion: true,
```
radames, line 15: @coyotte508 I've added this hint here so we can fix the url (i.e. add /v1/chat/completions) in case it's only modelId when making the streamingRequest

packages/inference/src/tasks/custom/request.ts
```diff
@@ -11,9 +11,17 @@ export async function request<T>(
 		task?: string | InferenceTask;
 		/** To load default model if needed */
 		taskHint?: InferenceTask;
+		/** Is chat completion compatible */
+		chatCompletion?: boolean;
```
radames, line 15: same here

packages/inference/test/HfInference.spec.ts
```diff
@@ -645,11 +639,114 @@ describe.concurrent(
...
+			});
+			if (res.choices && res.choices.length > 0) {
+				const completion = res.choices[0].message?.content;
+				expect(completion).toContain("to two");
+			}
+		});
+
+		it("chatCompletionStream modelId - OpenAI Specs", async () => {
+			const stream = hf.chatCompletionStream({
+				model: "mistralai/Mistral-7B-Instruct-v0.2",
```
radames, line 666: two new tests that have to work, we're assuming that if you're using `chatCompletion` it has to be `/v1/chat/completions` ,

radames (commented):
I've add `chatCompletion` hint, feel free to change if there is a better way to send that to the request call

packages/inference/test/HfInference.spec.ts
```diff
@@ -645,11 +639,141 @@ describe.concurrent(
...
+						model: "openai-community/gpt2",
+						messages: [{ role: "user", content: "Complete the equation 1+1= ,just the answer" }],
+						max_tokens: 500,
+						return_full_text: false,
+						temperature: 0.1,
+						seed: 0,
+					})
+					.next()
+			).rejects.toThrowError("Check if the model inference is compatible with chat completion.");
+		});
```
radames, line 707: if using modelid without support for chat completion,

coyotte508 (approved):
Ok for me, thanks @radames !

Regarding chat completion itself, don't have much domain knowledge, cc @Wauplin I guess!

radames:
Thanks @coyotte508, @Wauplin got some inspiration from your Python implementation. Much simpler for now, but please let me know what you think.

julien-c (commented):
this should be close to being ready to merge no?

Can you update an inference example in the main README and the docs @radames as this is expected to be one of the most used/useful method in AI inference nowadays?

radames:
great point @julien-c , I'll add docs , and I agree with you this method will be quite used, I have a few projects waiting for this to replace openai client. The PR is good to go, I just wanted @Wauplin to have a final look, but we could merge and deploy, so we can test it in the wild?

coyotte508:
Yes for sure @radames please update the README in the root and in `packages/inference` and then we cna merge

radames:
@coyotte508 Done, feel free to check and merge thanks! I'll push the demo list soon after the package is published 🚀

coyotte508:
Can you update the tests in `HfInference.spec.ts` to use the `chatCompletionStream` method instead of `streamingRequest`?

(also check my README changes)

It should be good afterwards

packages/inference/README.md
```diff
@@ -63,6 +61,89 @@ This will enable tree-shaking by your bundler.
...
+
+[Demo](https://huggingface.co/spaces/huggingfacejs/streaming-chat-completion)
+
+```typescript
+// Non-streaming API
+const out = await hf.chatCompletion({
+  model: "mistralai/Mistral-7B-Instruct-v0.2",
+  messages: [{ role: "user", content: "Complete the this sentence with words one plus one is equal " }],
+  max_tokens: 500,
+  return_full_text: false,
```
coyotte508, line 98: Is this parameter needed/useful for chatCompletion? It's not in the typescript type

radames: you're right, it's not part of chat completion specs, only text generation

radames:
> Can you update the tests in `HfInference.spec.ts` to use the `chatCompletionStream` method instead of `streamingRequest`?
> 
> (also check my README changes)
> 
> It should be good afterwards

@coyotte508 for the external providers, it has to be `streamingRequest` since they don't support receiving extra arguments that comes with the `chatCompletionStream` , which sends `options` params, that are compatible with our serveless API and TGI 

image

packages/inference/README.md
radames: removed it all, not needed

README.md
```diff
@@ -119,6 +119,30 @@ const HF_TOKEN = "hf_...";
...
+});
+
+// Streaming chat completion API
+for await (const chunk of openai.chatCompletionStream({
+  model: "gpt-3.5-turbo",
+  messages: [{ role: "user", content: "Complete the this sentence with words one plus one is equal " }],
+  max_tokens: 100
+})) {
+  console.log(chunk.choices[0].delta.content);
+}
```
julien-c, line 144: do we really need/want to support openai as a provider? in any case i dont think we want to document it in the main README

packages/inference/README.md
```diff
@@ -63,6 +61,85 @@ This will enable tree-shaking by your bundler.
...
+    { role: "user", content: "Complete the equation 1+1= ,just the answer" },
+  ],
+  max_tokens: 500,
+  temperature: 0.1,
+  seed: 0,
+})) {
+  if (chunk.choices && chunk.choices.length > 0) {
+    out += chunk.choices[0].delta.content;
+  }
+}
```
julien-c, line 136: (here is fine i guess)

packages/inference/README.md
```diff
@@ -497,13 +553,52 @@ for await (const output of hf.streamingRequest({
 }
 ```
 
+You can use any OpenAI Chat Completion API-compatible provider with the `chatCompletion` method.
```
julien-c, line 556:
```suggestion
You can use any Chat Completion API-compatible provider with the `chatCompletion` method.
```

README.md
radames: great point, the main support is TGI and Serveless, but it's possible to use other providers, this could be very nice if your app, switch between local, cloud. But I agree it doesn't need to be on the main Readme, there is an snippet on the other doc.

radames:
@coyotte508 here is the issue, when we use predefined tasks, such as `textGeneration`, which is compatible with our serveless API, https://huggingface.co/docs/api-inference/en/detailed_parameters, `option` parameter is valid. That's why on my example I use the custom request, since it won't send anything `options` 
 
https://github.com/huggingface/huggingface.js/blob/b66fcf345e03e72754aa20d0d4bb123b0d3d2f44/packages/inference/src/lib/makeRequestOptions.ts#L114-L117

radames:
Great point. We might not need otherOptions. It seems like we can do it via headers, and we're already doing it.

coyotte508:
> Great point. We might not need otherOptions. It seems like we can do it via headers, and we're already doing it.

@radames can you run tests again with `VCR_MODE=cache` to record the new `tapes.json`?

Hopefully it should be good afterwards

radames:
thanks @coyotte508 done! I ran the tests only for the new api `VCR_MODE=cache pnpm exec vitest run  -t "OpenAI Specs"`

julien-c (commented):
yay! this is close to being merged? i want to play with it:)

radames:
I think is just the final ok from @coyotte508 it's ready to be merged and released. I'm also waiting for this, to build a couple of things.

coyotte508:
Thanks @radames !

PR #649, opened 2024-05-03
[inference] Flipped usage of wait_for_model flag

mishig25:
I think `wait_for_model` usage is flipped incorrectly in https://github.com/huggingface/huggingface.js/pull/148. Check [files tab](https://github.com/huggingface/huggingface.js/pull/649/files)

What prompted this error:
from @gary149 

> 503 but no error appears in the UI?
https://huggingface.co/microsoft/Phi-3-mini-128k-instruct?text=hi

Explanation: although widget sends `wait_for_model: false`, inference.js does NOT throw error and tries to load the model and make a call to this model once it is loaded

### DIFF
--- a/packages/inference/src/tasks/custom/request.ts
+++ b/packages/inference/src/tasks/custom/request.ts
@@ -16,7 +16,7 @@ export async function request<T>(
 	const { url, info } = await makeRequestOptions(args, options);
 	const response = await (options?.fetch ?? fetch)(url, info);
 
-	if (options?.retry_on_error !== false && response.status === 503 && !options?.wait_for_model) {
+	if (options?.retry_on_error !== false && response.status === 503 && options?.wait_for_model) {
 		return request(args, {
 			...options,
 			wait_for_model: true,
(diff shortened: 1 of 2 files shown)


Wauplin (approved):
Looks good at first glance

mishig25:
will wait for others to confirm

packages/inference/src/tasks/custom/request.ts
```diff
@@ -16,7 +16,7 @@ export async function request<T>(
 	const { url, info } = await makeRequestOptions(args, options);
 	const response = await (options?.fetch ?? fetch)(url, info);
 
-	if (options?.retry_on_error !== false && response.status === 503 && !options?.wait_for_model) {
+	if (options?.retry_on_error !== false && response.status === 503 && options?.wait_for_model) {
```
coyotte508, line 19:
```suggestion
	if (options?.retry_on_error !== false && response.status === 503 && options?.wait_for_model === undefined) {
```

same for streaming

current behavior is wanted (to retry with `wait_for_model` if we get a 503)

coyotte508:
you should pass `retry_on_error: false` instead of `wait_for_model: false` in widgets, I think that would solve it

mishig25:
@coyotte508  submitted #650, should I close this PR #649 then?

coyotte508:
yes you can close :)

mishig25:
superceded by https://github.com/huggingface/huggingface.js/pull/650

PR #652, opened 2024-05-03
Gguf tweaks

julien-c:
cc @ngxson too (second commit is optional, up to you)

### DIFF
--- a/packages/gguf/src/gguf.ts
+++ b/packages/gguf/src/gguf.ts
@@ -3,7 +3,7 @@ import { GGUFValueType } from "./types";
 import { promisesQueue } from "./utils/promisesQueue";
 
 export type { MetadataBaseValue, MetadataValue, Version, GGUFMetadata, GGUFTensorInfo, GGUFParseOutput } from "./types";
-export { GGUFValueType, GGMLQuantizationType, Architecture } from "./types";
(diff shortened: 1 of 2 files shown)


packages/gguf/src/gguf.ts
```diff
@@ -3,7 +3,7 @@ import { GGUFValueType } from "./types";
 import { promisesQueue } from "./utils/promisesQueue";
 
 export type { MetadataBaseValue, MetadataValue, Version, GGUFMetadata, GGUFTensorInfo, GGUFParseOutput } from "./types";
-export { GGUFValueType, GGMLQuantizationType } from "./types";
+export { GGUFValueType, GGMLQuantizationType, GGUFArchitecture } from "./types";
```
ngxson, line 6: I'm wondering if it should be named `GGMLArchitecture`, because gguf is more like a container, so it does not contains the actual architecture. It's a bit like `GGMLQuantizationType` is defined at tensor level (ggml) and rather than at gguf level.

(FYI, before gguf, llama.cpp used `*.ggml` format to distribute tensor data)

julien-c: ok i'll revert that second commit then

ngxson (approved):
LGTM. Thank you!

PR #653, opened 2024-05-03
[gguf] Ability to specify some fetch headers

julien-c:
cc @ngxson too

### DIFF
--- a/packages/gguf/src/gguf.ts
+++ b/packages/gguf/src/gguf.ts
@@ -64,6 +64,7 @@ class RangeView {
 			 * Custom fetch function to use instead of the default one, for example to use a proxy or edit headers.
 			 */
 			fetch?: typeof fetch;
+			fetchHeaders?: Record<string, string>;
 		}
 	) {
 		this.chunk = 0;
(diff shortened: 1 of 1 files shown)


coyotte508:
hmmm there is already a custom `fetch` function, is there a strong need?

packages/gguf/src/gguf.ts
```diff
@@ -64,6 +64,7 @@ class RangeView {
 			 * Custom fetch function to use instead of the default one, for example to use a proxy or edit headers.
 			 */
 			fetch?: typeof fetch;
+			fetchHeaders?: Record<string, string>;
```
coyotte508, line 67:
```suggestion
			additionalFetchHeaders?: Record<string, string>;
```

julien-c:
> hmmm there is already a custom `fetch` function, is there a strong need?

Well, for safetensors we have a `credentials` param

packages/gguf/src/gguf.ts
julien-c: [2f9339c](https://github.com/huggingface/huggingface.js/pull/653/commits/2f9339c5809223dffb391685117a08e7a014e713)

coyotte508 (approved):

ngxson:
Thanks for tagging me. I have no comment about this PR. But just for my curiosity, is this functionality needed to access gated model? (with a token passed via header, as I remember?)

PR #655, opened 2024-05-04
gguf: better type usage

ngxson:
Follow up #640

Ref comments:
- https://github.com/huggingface/huggingface.js/pull/640#pullrequestreview-2031992487 by @julien-c suggests using a check `metadata["general.architecture"] === ...` to select the correct type
- https://github.com/huggingface/huggingface.js/pull/640#issuecomment-2086561504 by @coyotte508 suggests using less generic but more verbose code

The type system introduce in this PR allows type-checking at both compile time & runtime:

```ts
const model: GGUFMetadata<GGUFType.STRICT> = null as any;

if (model["general.architecture"] === "whisper") {
	model["encoder.whisper.block_count"] = 0;
	// @ts-expect-error because it must be a number
	model["encoder.whisper.block_count"] = "abc";
}

if (model["tokenizer.ggml.model"] === undefined) {
	// @ts-expect-error because it's undefined
	model["tokenizer.ggml.eos_token_id"] = 1;
}
if (model["tokenizer.ggml.model"] === "gpt2") {
	// @ts-expect-error because it must be a number
	model["tokenizer.ggml.eos_token_id"] = undefined;
	model["tokenizer.ggml.eos_token_id"] = 1;
}

if (model["general.architecture"] === "mamba") {
	model["mamba.ssm.conv_kernel"] = 0;
	// @ts-expect-error because it must be a number
	model["mamba.ssm.conv_kernel"] = "abc";
}
if (model["general.architecture"] === "llama") {
	// @ts-expect-error llama does not have ssm.* keys
	model["mamba.ssm.conv_kernel"] = 0;
}
```

Type checks can be disable with `GGUFMetadata<GGUFType.NON_STRICT>`


ngxson:
Also in the process, I noticed some details not being mentioned in gguf specs:
- For generic fields, only `general.architecture` and `general.name` are required ==> this is because in cpp code, only `LLM_KV_GENERAL_NAME` and `LLM_KV_GENERAL_ARCHITECTURE` are being read
- MOE is optional, ref:  
  https://github.com/ggerganov/llama.cpp/blob/842500144ee02c8b0a46d2cc43880f8d80998fa5/llama.cpp#L3785
  (explain: third param of `get_key` is set to `false`, indicating that it is not required)
- whisper.cpp indeed does not use gguf, ref code:  
 https://github.com/ggerganov/whisper.cpp/blob/58210d6a7634ea1e42e0a2dab611f4a0518731dc/whisper.cpp#L1258  
 So unfortunately we don't have any way to test whisper gguf. Regarding fields being mentioned in the documentation, I found a related discussion here: https://github.com/ggerganov/ggml/pull/302#issuecomment-1616747195

packages/gguf/src/types.ts
```diff
@@ -75,21 +86,40 @@ interface Tokenizer {
...
+		"whisper.encoder.attention.head_count": number;
+		"whisper.decoder.attention.head_count": number;
+	};
+
+/// Types for parse output
+
+export type GGUFStrictType = true;
+export type GGUFNonStrictType = false;
+
+export type GGUFMetadata<T extends GGUFStrictType | GGUFNonStrictType = GGUFStrictType> = {
```
mishig25, line 116: small nit. Like in `TArchitecture`, we can use more descriptive generic name other than just `T`. Maybe `TGgufType` ?

mishig25:
Very nice PR! And much needed improvement to the typing of gguf, especially type inference.
Left some nits. Otherwise, we are close to merge 🚀

packages/gguf/src/types.ts
```diff
@@ -75,21 +86,40 @@ interface Tokenizer {
...
+	ModelBase<"decoder.whisper"> & {
+		"whisper.encoder.mels_count": number;
+		"whisper.encoder.attention.head_count": number;
+		"whisper.decoder.attention.head_count": number;
+	};
+
+/// Types for parse output
+
+export type GGUFStrictType = true;
+export type GGUFNonStrictType = false;
```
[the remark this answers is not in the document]
ngxson: Implemented in https://github.com/huggingface/huggingface.js/pull/655/commits/c2afbdc080980c81ad37a933dfcea840dde9d3bc

packages/gguf/src/types.ts
ngxson: Implemented in https://github.com/huggingface/huggingface.js/pull/655/commits/c2afbdc080980c81ad37a933dfcea840dde9d3bc

packages/gguf/src/types.ts
```diff
@@ -75,21 +86,43 @@ interface Tokenizer {
...
+		"whisper.encoder.mels_count": number;
+		"whisper.encoder.attention.head_count": number;
+		"whisper.decoder.attention.head_count": number;
+	};
+
+/// Types for parse output
+
+export enum GGUFType {
+	strict,
+	nonStrict,
```
mishig25, line 115:
```suggestion
	STRICT,
	NON_STRICT,
```

stylitic nit. we usually prefer the enum values to be uppercase

ngxson: Thanks, that's done with https://github.com/huggingface/huggingface.js/pull/655/commits/5f547dd882ad07dad612225f6b8193845fb318a9

mishig25:
Question regarding `GGUFType.NON_STRICT`. Why do we need  `GGUFType.NON_STRICT`? Shouldn't we want `GGUFType.STRICT` to be always the case?

ngxson:
I believe that `GGUFType.NON_STRICT` will be useful in case user want to use their own gguf structure. In fact, gguf can also be used for storing [control vectors](https://github.com/ggerganov/llama.cpp/pull/5970), [lora weights](https://github.com/ggerganov/llama.cpp/blob/master/convert-lora-to-ggml.py), and many other use cases outside of llama.cpp

mishig25 (approved):
LGTM !

mishig25:
@coyotte508 could you give a quick look and I will merge 🚀

julien-c (approved):

packages/gguf/src/gguf.ts
```diff
@@ -308,6 +308,9 @@ export async function gguf(
 			}
 		}
 		offset += valueResult.length;
+		/// TODO(fix typing)
+		// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+		// @ts-ignore
```
coyotte508, line 313: Let's use `// @ts-expect-error` instead of `// @ts-ignore` in general

(no need for eslint-disable this way)

Here I think you can change `const metadata: GGUFMetadata` to `const metadata: GGUFMetadata<GGUFType.NON_STRICT>` to remove the error (not sure if it's the best fix)

packages/gguf/src/types.ts
```diff
@@ -75,21 +86,43 @@ interface Tokenizer {
...
+	};
+
+/// Types for parse output
+
+export enum GGUFType {
+	STRICT,
+	NON_STRICT,
+}
+
+export type GGUFMetadata<TGGUFType extends GGUFType = GGUFType.STRICT> = {
```
coyotte508, line 118: We should probably switch to something like https://github.com/sindresorhus/type-fest/blob/main/source/except.d.ts

```ts
interface GGUFMetadataOptions {
	/**
   * ...
   * 
	 * @default true
   */
  strict: boolean;
}


export GGUFMetadata<Options extends GGUFMetadataOptions = { strict: true}> {
  ...
```

coyotte508 (commented):
Looks good!

I'm wondering about flat paths vs nested metadata (maybe it could be another option in the future :thinking:) but LGTM with the 1/2 changes requested

packages/gguf/src/types.ts
ngxson: Good idea, I've never thought about this before. Implemented in https://github.com/huggingface/huggingface.js/pull/655/commits/31bac8bbba365e13c517c48452fdee1d5843abd3

packages/gguf/src/gguf.ts
ngxson: Fixed in https://github.com/huggingface/huggingface.js/pull/655/commits/31bac8bbba365e13c517c48452fdee1d5843abd3

ngxson:
> I'm wondering about flat paths vs nested metadata (maybe it could be another option in the future 🤔)

@coyotte508 I thought about using nested object, but that seems to be quite complicated atm, since part of the code/naming here is copied directly from cpp file (so it save us some headaches)

But I'm also agree that having nested object is a reasonable feature, since gguf keys are "prefixed" anyway. Maybe we will re-visit this idea later on.

julien-c:
alright let's merge!

re. nested properties, that's what they did in https://github.com/ahoylabs/gguf.js (credited as one of the inspirations for this package) but I'm not sure it's worth it / I find it cleaner to just expose the raw data from the GGUF file.

PR #656, opened 2024-05-05
gguf: Add ability to load local file

ngxson:
Being able to load a local gguf file can be useful when we want to debug a gguf file.

**Without this PR**, this ability could be done by using [file-fetch](https://www.npmjs.com/package/file-fetch). However, that won't work with big models, since the whole file is loaded into RAM.

This PR add a new `RangeViewLocalFile` internal class that extends `RangeView`. It redirects calls to `fetchChunk()` to `fs.createReadStream` with the appropriate byte range. This allows the library to read specific chunk from a local file.

For security reason, this ability is locked under `localFile: boolean` param. By default, it is disabled (i.e. when this library is run on hub backend, this param is disabled if unspecified)

- [x] Add test case to `gguf.spec.ts`
- [x] Being able to build with target=browser (only build, but will throw error on browser if being used)


packages/gguf/src/gguf.ts
```diff
@@ -56,6 +56,9 @@ class RangeView {
 	get view(): DataView {
 		return this.dataView;
 	}
+	get currentChunk(): number {
```
mishig25, line 59: looks like this method is only used inside `RangeViewLocalFile`. Therefore, you can move this method to  `RangeViewLocalFile` only

packages/gguf/src/gguf.ts
```diff
@@ -234,9 +271,10 @@ export async function gguf(
 		fetch?: typeof fetch;
 		additionalFetchHeaders?: Record<string, string>;
 		computeParametersCount?: boolean;
+		localFile?: boolean;
 	}
 ): Promise<GGUFParseOutput & { parameterCount?: number }> {
-	const r = new RangeView(url, params);
+	const r = params?.localFile ? new RangeViewLocalFile(url, params) : new RangeView(url, params);
```
mishig25, line 277: maybe we should rename the `url` variable here to `uri` since it can refer to local file

ngxson: Resolved with https://github.com/huggingface/huggingface.js/pull/656/commits/e68fc5075acd75da9a7fb0e143e63faee9f82ff7

packages/gguf/src/gguf.ts
ngxson: Yeah I forgot that I can use `protected` keyword in ts. I was doing this way because `private chunk` can't be read by subclass.

It is fixed with https://github.com/huggingface/huggingface.js/pull/656/commits/e68fc5075acd75da9a7fb0e143e63faee9f82ff7

coyotte508:
Note that we already have `FileBlob`, that you could reuse: https://github.com/huggingface/huggingface.js/blob/4fdb7cb8fdcbcde3210150e684f3e31d17cd5bab/packages/hub/src/utils/FileBlob.ts

```ts
const blob  = await Fileblob.create("./local-file");

const chunk1 = await blob.slice(start, end).then(slice => slice.arrayBuffer());
const chunk2 = ...
```

Note that there are also `WebBlob` and even a more generic [`createBlob`](https://github.com/huggingface/huggingface.js/blob/203fd4ca95e2fa86f8481250ea9ca116e05d8dc8/packages/hub/src/utils/createBlob.ts) that takes any `URL` - that are used by `@huggingface/hub` internally.

It could be copied over or shared. Or a small improvement: no need for streams to get a chunk you can do it like this:

```ts
const file = await open(this.path, "r");

try {
	const chunk = await file.read(buffer, 0, end-start, start);
} finally {
	await file.close();
}
```

And soon with typescript 5.5 beta:

```ts
await using file = await open(this.path, "r");

const chunk = await file.read(buffer, 0, end-start, start);
```

---

anyway, just passing notes, will leave it up to you folks :)

packages/gguf/src/gguf.ts
```diff
@@ -128,6 +128,38 @@ class RangeView {
...
+			const stream = require("node:fs").createReadStream(this.uri, {
+				start: this.chunk * HTTP_CHUNK_SIZE,
+				end: (this.chunk + 1) * HTTP_CHUNK_SIZE,
+			});
+			stream.on("error", reject);
+			stream.on("data", (chunk: Buffer) => (buffer = Buffer.concat([buffer, chunk])));
+			stream.on("end", () => {
+				this.appendBuffer(buffer);
+				resolve();
+			});
```
julien-c, line 158: hmm no need for streams probably as @coyotte508 says

julien-c (commented):
i'm not opposed to this (an alternative would be to document how to simply start a local webserver from where your files are)

ngxson:
Thanks @coyotte508 for the suggestion. I ended up copying `FileBlob` into the gguf library, because:
- Moving it to shared ==> I'm not sure how it will effect `FileBlob.spec.ts`
- Via adding `"@huggingface/hub": "workspace:^"` to `gguf/package.json` ==> It doesn't compile, because `FileBlob.ts` is not exported

Otherwise, for now, it works fine on both node & browser. Feel free to tell me if you have other ideas.

packages/gguf/tsconfig.json
```diff
@@ -13,6 +13,6 @@
 		"noImplicitOverride": true,
 		"outDir": "./dist"
 	},
-	"include": ["src"],
+	"include": ["src", "../shared/src"],
```
coyotte508, line 16:
```suggestion
	"include": ["src"],
```

packages/gguf/src/gguf.ts
```diff
@@ -1,6 +1,7 @@
 import type { MetadataValue, Version, GGUFMetadata, GGUFTensorInfo, GGUFParseOutput } from "./types";
 import { GGUFValueType } from "./types";
 import { promisesQueue } from "./utils/promisesQueue";
+import { isFrontend } from "../../shared";
```
coyotte508, line 4: Sorry better copy that file too, I'm deleted `shared` in another PR

packages/gguf/tsconfig.json
ngxson: Resolved in https://github.com/huggingface/huggingface.js/pull/656/commits/1caedf34106013a2e1f7527d6a1db890da6eb2d6

packages/gguf/src/gguf.ts
ngxson: Resolved in https://github.com/huggingface/huggingface.js/pull/656/commits/1caedf34106013a2e1f7527d6a1db890da6eb2d6

packages/gguf/src/gguf.ts
```diff
@@ -205,38 +227,41 @@ function readMetadataValue(
...
-	url: string,
+	uri: string,
 	params?: {
 		/**
 		 * Custom fetch function to use instead of the default one, for example to use a proxy or edit headers.
 		 */
 		fetch?: typeof fetch;
 		additionalFetchHeaders?: Record<string, string>;
 		computeParametersCount?: boolean;
+		localFile?: boolean;
```
coyotte508, line 261:
```suggestion
		/** By default this function will not look at the local file system for security reasons */
		allowLocalFile?: boolean;
```

Also, when detecting a relative url in a backend context, we should probably throw an explicit error if this is not set ("Set allowLocalFile to ..."). So that the user can know why it failed, since the user may not know about the option

ngxson: I implemented it in https://github.com/huggingface/huggingface.js/pull/656/commits/a0e174e581f9572da8ebc15e9c9c32c34ed05b24 , please have a look

coyotte508 (approved):

coyotte508:
regarding conflicts: I think you can remove `type-fest` from the dev dependencies

julien-c (approved):

mishig25:
@ngxson merged it 🚀 

Would you mind adding a section on how to load locally in the readme https://github.com/huggingface/huggingface.js/blob/main/packages/gguf/README.md#usage ?

Thanks a lot !

PR #657, opened 2024-05-06
[widgets] Conversation widget render markdown

mishig25:
### Description

Make conversational widget render markdown. Pretty much cop paste of [chat-ui implementation here](https://github.com/huggingface/chat-ui/blob/4805941429bb3dccbf95944b469c733cfe81f708/src/lib/components/chat/ChatMessage.svelte#L221) with a difference:
1. rendering code in  `<div class="font-mono whitespace-pre-wrap">` rather than specific CodeBlock.svelte (that has copy-code feature)
2. No support for latex in this PR. Should I add it?

| main    | this PR |
| -------- | ------- |
|  |    |

### testing locally
`pnpm install` to get `marked` (new dep; fyi, we also use marked for chat-ui & moon)


packages/widgets/src/lib/components/InferenceWidget/shared/WidgetOuputConvoBubble/WidgetOutputConvoBubble.svelte
```diff
@@ -1,14 +1,64 @@
...
+
+		const stopTokens = ["<|endoftext|>"];
+		if (typeof specialTokensMap?.eos_token === "string") {
+			stopTokens.push(specialTokensMap.eos_token);
+		}
+		for (const stop of stopTokens) {
+			if (ret.endsWith(stop)) {
+				ret = ret.slice(0, -stop.length).trim();
+			}
+		}
```
mishig25, line 37: do I need these lines considering that we already gave https://github.com/huggingface/huggingface.js/blob/59d641b6a7ccdfdf86596f69e4845ef2680e3834/packages/widgets/src/lib/components/InferenceWidget/widgets/ConversationalWidget/ConversationalWidget.svelte#L140-L144

cc: @xenova

xenova: I think it is, because this just controls stripping of the final special token (output), right? (and is different to the rendered chat template / input).

osanseviero (approved):
Very nice :fire: conceptually LGTM

mishig25:
merging it

PR #658, opened 2024-05-06
♻️ Make "go to source definition" work

coyotte508:


julien-c (approved):
lgtm in principle

packages/widgets/vite.config.ts
```diff
@@ -25,6 +25,7 @@ export default defineConfig(({ mode }) => {
 					emitCss: false,
 					compilerOptions: {
 						hydratable: true,
+						// @ts-expect-error to check cc @krampstudio
```
coyotte508, line 28: cc @krampstudio , apparently the correct type is `Omit<ViteOptions, "generate">`, for now bypassed the error (not sure the error is relevant or not)

krampstudio: indeed the type in vite-plugin-svelte omits it, and the [doc](https://github.com/sveltejs/vite-plugin-svelte/blob/main/docs/config.md#compileroptions) says it's non-configurable.
But we use it in [lib mode](https://vitejs.dev/guide/build#library-mode) and the type might not be accurate since the options are passed down to the svelte compiler

packages/tasks/package.json
```diff
@@ -23,8 +23,8 @@
 		"lint:check": "eslint --ext .cjs,.ts .",
 		"format": "prettier --write .",
 		"format:check": "prettier --check .",
-		"prepublishOnly": "pnpm run build",
-		"build": "tsup src/index.ts --format cjs,esm --clean --dts && pnpm run inference-codegen",
+		"prepublishOnly": "pnpm run inference-codegen && git diff --name-only --exit-code src && pnpm run build",
+		"build": "tsup src/index.ts --format cjs,esm --clean && tsc --emitDeclarationOnly --declaration",
```
coyotte508, line 27: @Wauplin @SBrandeis I changed the codegen to automatically run in `@huggingface/tasks` only during prepublish (publish will error if not up to date)

We would do the same for tgi codegen

It avoids doing it during every `pnpm install` but should be done manually before publishing

julien-c:
is the `Adding a package` section in `CONTRIBUTING.md` still up to date with this change?

coyotte508:
> is the `Adding a package` section in `CONTRIBUTING.md` still up to date with this change?

good point, updated

PR #659, opened 2024-05-07
Local apps

julien-c:


packages/tasks/src/local-apps.ts
```diff
@@ -0,0 +1,119 @@
...
+		docsUrl: "https://lmstudio.ai",
+		mainTask: "text-generation",
+		displayOnModelPage: isGgufModel,
+		deeplink: (model) => new URL(`lmstudio://open_from_hf?model=${model.id}`),
+	},
+	jan: {
+		prettyLabel: "Jan",
+		docsUrl: "https://jan.ai",
+		mainTask: "text-generation",
+		displayOnModelPage: isGgufModel,
```
Vaibhavs10, line 86: AFAIR Jan works for both GGUFs as well as TRT-LLM too, would be good to add support for that if possible too! 🙏

packages/tasks/src/local-apps.ts
```diff
@@ -0,0 +1,119 @@
...
+			snippet: (model: ModelData) => string;
+	  }
+);
+
+function isGgufModel(model: ModelData) {
+	return model.tags.includes("gguf");
+}
+
+const snippetLlamacpp = (model: ModelData): string => {
+	return `llama-cpp \
```
Vaibhavs10, line 50: This doesn't work atm. :/

```
./main
```

Vaibhavs10 (commented):
Love itt! Question: could this also apply to audio i.e. whisper for example?

packages/tasks/src/local-apps.ts
julien-c: arg :(

care to suggest the replacement?

pcuenca (approved):
Fantastic! This should encourage other app creators to follow through!

packages/tasks/src/local-apps.ts
ngxson: Probably users can take the pre-built binary from here: https://github.com/ggerganov/llama.cpp/releases

Then use either `main` or `server`. Personally I recommend `server`, since `main` does not have support for chat templates.

The command will be `./server -m model.gguf` (remember to add `.exe` for windows). Then, a basic frontend then can be access via `http://localhost:8080`

ngxson: Also a small note: we also need to specify `--hf-file` along with `--hf-repo`, because gguf files does not have a fixed file name. For example:

```
./server --hf-repo ggml-org/models --hf-file tinyllamas/stories260K.gguf -m ./model.gguf
```

Though, that seems a bit complicated IMO, so I prefer to manually download the file and use `-m` to load it.

packages/tasks/src/local-apps.ts
```diff
@@ -0,0 +1,119 @@
...
+	 */
+	prettyLabel: string;
+	/**
+	 * Link to get more info about a local app (website etc)
+	 */
+	docsUrl: string;
+	/**
+	 * main category of app
+	 */
+	mainTask: PipelineType;
```
osanseviero, line 19: Should we have a list here? I don't think we want a 1:1 mapping of local app to pipeline, e.g. if lmstudio supports tti later on, we should just add one to the list. Similar to https://github.com/huggingface/huggingface.js/blob/main/packages/tasks/src/library-to-tasks.ts#L29

packages/tasks/src/local-apps.ts
```diff
@@ -0,0 +1,119 @@
...
+		displayOnModelPage: (model) => model.tags.includes("textual_inversion"),
+		deeplink: (model) => new URL(`drawthings://open_from_hf?model=${model.id}`),
+	},
+	diffusionbee: {
+		prettyLabel: "DiffusionBee",
+		docsUrl: "https://diffusionbee.com",
+		mainTask: "text-to-image",
+		macOSOnly: true,
+		comingSoon: true,
+		displayOnModelPage: (model) => model.library_name === "diffusers" && model.pipeline_tag === "text-to-image",
```
osanseviero, line 114: very cool

osanseviero (commented):
Nice!

packages/tasks/src/local-apps.ts
julien-c: this is only to categorize in the listing UI, though. i.E. it doesn't control on which page it will be displayed

packages/tasks/README.md
```diff
@@ -1,8 +1,18 @@
 # Tasks
 
-This package contains data used for https://huggingface.co/tasks.
+This package contains the definition files (written in Typescript) for the huggingface.co hub's:
+
+- **pipeline types** a.k.a. **task types** (used to determine which widget to display on the model page, and which inference API to run)
+- **default widget inputs** (when they aren't provided in the model card)
+- definitions and UI elements for **model libraries** (and soon for **dataset libraries**).
```
mishig25, line 7:
```suggestion
- **pipeline types** (a.k.a. **task types**) - used to determine which widget to display on the model page, and which inference API to run.
- **default widget inputs** - when they aren't provided in the model card.
- definitions and UI elements for **model libraries** (and soon for **dataset libraries**).
```

packages/tasks/README.md
```diff
@@ -1,8 +1,18 @@
...
 
-## Philosophy behind Tasks
+Please add to any of those definitions by opening a PR. Thanks 🔥
```
mishig25, line 9:
```suggestion
Please add any missing ones to these definitions by opening a PR. Thanks 🔥
```

packages/tasks/src/local-apps.ts
```diff
@@ -0,0 +1,119 @@
...
+import type { PipelineType } from "./pipelines";
+
+/**
+ * Elements configurable by a local app.
+ */
+export type LocalApp = {
+	/**
+	 * Name that appears in buttons
+	 */
+	prettyLabel: string;
```
mishig25, line 11: just making a comment that previously we ended up mixing snake_case & camelCase like in https://github.com/ngxson/huggingface.js/blob/5f547dd882ad07dad612225f6b8193845fb318a9/packages/tasks/src/model-data.ts#L125-L143

packages/tasks/src/local-apps.ts
```diff
@@ -0,0 +1,119 @@
...
+);
+
+function isGgufModel(model: ModelData) {
+	return model.tags.includes("gguf");
+}
+
+const snippetLlamacpp = (model: ModelData): string => {
+	return `llama-cpp \
+	--hf-repo ${model.id} \
+	-m file.gguf \
```
mishig25, line 52:
```suggestion
	-m {FILE_NAME}.gguf \
```

packages/tasks/src/local-apps.ts
julien-c: marking this as resolved

packages/tasks/src/local-apps.ts
julien-c: we don't have any file name here so i think it's better to write file.gguf

packages/tasks/src/local-apps.ts
julien-c:
```suggestion
	return `./main \
```

packages/tasks/src/local-apps.ts
julien-c: let's do it in a subsequent PR if it works for you @Vaibhavs10

julien-c (commented):
this v1 is ready to merge (cc @krampstudio), shall i merge and we can iterate further in other PRs?

packages/tasks/src/local-apps.ts
Vaibhavs10: Yes! also chatting with Jan folks on this.

krampstudio:
> this v1 is ready to merge (cc @krampstudio), shall i merge and we can iterate further in other PRs?

yes it looks a good starting point. 

btw, I think we should also re-export  into https://github.com/huggingface/huggingface.js/blob/main/packages/tasks/src/index.ts (cc @coyotte508)

```ts
export { LOCAL_APPS } from "./local-apps"
export type { LocalApp, LocalAppKey } from "./local-apps";
```

Vaibhavs10 (approved):
Quite excited by this! LFG! 🚀

packages/tasks/src/local-apps.ts
Vaibhavs10: Noting this here that this would be interesting to solve for, specially when it comes to multi-quants repo i.e. a repo with more than one quant.

julien-c:
good point @krampstudio, want to push this directly to this branch? (and merge it whenever you want)

PR #660, opened 2024-05-10
[gguf] Fx type in "should parse a local file"

mishig25:
Follow up to https://github.com/huggingface/huggingface.js/pull/656

cc: @ngxson


coyotte508 (approved):

ngxson:
Thank you! LGTM.

I was quite surprise that it isn't got catch by `pnpm run lint:check`. Do you have any idea why? (Or maybe because my branch is not up-to-date, so there's lint config that I didn't know about)

mishig25:
> I was quite surprise that it isn't got catch by pnpm run lint:check

indeed, ` pnpm run lint:check` didn't catch it. But the tests on huggingface.js were failing. Maybe @coyotte508 knows why

PR #663, opened 2024-05-12
gguf: update README

ngxson:
Follow up #655 and https://github.com/huggingface/huggingface.js/pull/656#issuecomment-2104265961

Added some examples on how to use local file + strictly typed

### DIFF
--- a/packages/gguf/README.md
+++ b/packages/gguf/README.md
@@ -18,6 +18,8 @@ npm install @huggingface/gguf
 
 ## Usage
 
+### Basic usage
+
 ```ts
 import { GGMLQuantizationType, gguf } from "@huggingface/gguf";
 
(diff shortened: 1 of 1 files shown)


packages/gguf/README.md
```diff
@@ -56,6 +58,44 @@ console.log(tensorInfos);
...
+
+By default, known fields in `metadata` are typed. This includes various fields found in [llama.cpp](https://github.com/ggerganov/llama.cpp), [whisper.cpp](https://github.com/ggerganov/whisper.cpp) and [ggml](https://github.com/ggerganov/ggml).
+
+```ts
+const { metadata, tensorInfos } = await gguf(URL_MODEL);
+
+// Type check for model architecture at runtime
+if (metadata["general.architecture"] === "llama") {
+
+  // "llama.attention.head_count" is a valid key for llama architecture
```
julien-c, line 81:
```suggestion
  // "llama.attention.head_count" is a valid key for llama architecture, this is typed as a number
```

julien-c (approved):
love it!

BTW, if you want (not sure it makes sense for this particular change), there's also a doc page inside the Hub docs, here: https://huggingface.co/docs/hub/gguf

There's also a hosted doc for huggingface.js [here](https://huggingface.co/docs/huggingface.js) in case this is useful in the future.

packages/gguf/README.md
```diff
@@ -56,6 +58,44 @@ console.log(tensorInfos);
...
+  console.log(model["llama.attention.head_count"]);
+
+  // "mamba.ssm.conv_kernel" is an invalid key, because it requires model architecture to be mamba
+  console.log(model["mamba.ssm.conv_kernel"]); // error
+}
+```
+
+### Disable strictly typed
+
+Because GGUF format can be used to store tensors, we can technically use it for other usages. For example, storing [control vectors](https://github.com/ggerganov/llama.cpp/pull/5970), [lora weights](https://github.com/ggerganov/llama.cpp/pull/2632),...
```
mishig25, line 91:
```suggestion
Because GGUF format can be used to store tensors, we can technically use it for other usages. For example, storing [control vectors](https://github.com/ggerganov/llama.cpp/pull/5970), [lora weights](https://github.com/ggerganov/llama.cpp/pull/2632), etc.
```

mishig25 (approved):
lgtm! Looks great

ngxson:
> BTW, if you want (not sure it makes sense for this particular change), there's also a doc page inside the Hub docs, here: https://huggingface.co/docs/hub/gguf
> 
> There's also a hosted doc for huggingface.js [here](https://huggingface.co/docs/huggingface.js) in case this is useful in the future.

Thanks for the info. I will make a PR to the hosted docs if needed.

mishig25:
Awesome !

PR #664, opened 2024-05-13
[widgets] use `chatCompletionStream`

mishig25:
Thanks to https://github.com/huggingface/huggingface.js/pull/645, it is now possible to use `chatCompletionStream` in Conversational Widget

image

### DIFF
--- a/packages/widgets/src/lib/components/InferenceWidget/widgets/ConversationalWidget/ConversationalWidget.svelte
+++ b/packages/widgets/src/lib/components/InferenceWidget/widgets/ConversationalWidget/ConversationalWidget.svelte
@@ -2,21 +2,20 @@
 	import { onMount, tick } from "svelte";
 	import type { WidgetProps, ExampleRunOpts, InferenceRunOpts } from "../../shared/types.js";
 	import type { Options } from "@huggingface/inference";
-	import { Template } from "@huggingface/jinja";
 	import type {
 		SpecialTokensMap,
 		TokenizerConfig,
 		WidgetExampleTextInput,
-		TextGenerationInput,
+		ChatCompletionInput,
 		WidgetExampleOutputText,
 		WidgetExampleChatInput,
 		WidgetExample,
 		AddedToken,
+		ChatCompletionInputMessage,
 	} from "@huggingface/tasks";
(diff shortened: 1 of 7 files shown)


mishig25:
ready for review !

radames:
nice @mishig25 LGTM! one curious question, how are you differentiating the models supporting chatCompletion API versus simple text generation?  this https://huggingface.co/bigscience/bloom-560m from https://huggingface.co/HuggingFaceH4/zephyr-7b-beta? Because the `chatCompletion` would throw an error in case the backend doesn't support the API,

packages/widgets/src/lib/components/InferenceWidget/widgets/ConversationalWidget/ConversationalWidget.svelte
```diff
@@ -125,33 +117,18 @@
...
-			error = `An error occurred while rendering the chat template: "${(e as Error).message}"`;
-			return;
-		}
+		const previousMessages = [...messages];
 
-		const input: TextGenerationInput & Required<Pick<TextGenerationInput, "parameters">> = {
-			inputs: chatText,
-			parameters: {
-				return_full_text: false,
-			},
```
julien-c, line 154: given that's recent code from @Wauplin and @SBrandeis, i would like a review from them on this

julien-c:
@radames i think the conversational widget would not be displayed on pure text-generation models like https://huggingface.co/bigscience/bloom-560m

radames:
> @radames i think the conversational widget would not be displayed on pure text-generation models like https://huggingface.co/bigscience/bloom-560m

yes that's right, I'm just curious how on differentiate these models since they're all `pipeline_tag: text-generation` ?

julien-c:
@radames there's a notion of a `WidgetType` that can be conversational:

https://github.com/huggingface/huggingface.js/blob/d1c0cb8604db4f504fa33b233fc12b9a501abc17/packages/tasks/src/pipelines.ts#L666

If a model repo is `text-generation` AND it has a tag of `conversational`, it displays that widget

cc @osanseviero who implemed

SBrandeis (approved):
Nice factorization!

PR #667, opened 2024-05-14
Local apps links 

krampstudio:
Deeplinks (and Faraday rebranding)

 - [x] LM Studio: deeplink works well with 0.2.23
 - [x] Jan: deeplink works well with the nightly
 - [ ] Faraday/Backyard: the weblink doesn't work (yet) 
 - [ ] Draw things: coming soon
 - [ ] Diffusionbee: coming soon

### DIFF
--- a/packages/tasks/src/local-apps.ts
+++ b/packages/tasks/src/local-apps.ts
@@ -99,10 +99,11 @@ export const LOCAL_APPS = {
 		docsUrl: "https://drawthings.ai",
 		mainTask: "text-to-image",
 		macOSOnly: true,
+		comingSoon: true,
 		/**
 		 * random function, will need to refine the actual conditions:
 		 */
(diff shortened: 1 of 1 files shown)


packages/tasks/src/local-apps.ts
```diff
@@ -34,6 +34,12 @@ export type LocalApp = {
 			 */
 			deeplink: (model: ModelData) => URL;
 	  }
+	| {
+			/**
+			 * If the app supports opening through their website
+			 */
+			weblink?: (mode: ModelData) => URL;
+	  }
```
julien-c, line 42: it could be the same no? ie. i was thinking of using `deeplink()` for it (conceptually speaking it's the same)

krampstudio: yes I can be the same

julien-c (approved):

julien-c (approved):

PR #669, opened 2024-05-14
Add download stats for big_vision

osanseviero:


packages/tasks/src/model-libraries.ts
```diff
@@ -102,6 +102,15 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		snippets: snippets.bertopic,
 		filter: true,
 	},
+	big_vision: {
```
julien-c, line 105: dash or underscore?

julien-c: yep sounds good, but note that if they ever publish to PyPI, it will become `big-vision` :)

osanseviero: Yes, although as a research repo I think they might never publish there hehe

PR #670, opened 2024-05-14
Tasks: multiline snippet for llama.cpp

krampstudio:


coyotte508:
Can you explain the issue? Intuitively I expect antislashes at the end of lines

packages/tasks/src/local-apps.ts
```diff
@@ -47,10 +47,11 @@ function isGgufModel(model: ModelData) {
...
 const snippetLlamacpp = (model: ModelData): string => {
-	return `./main \
-	--hf-repo ${model.id} \
-	-m file.gguf \
-	-p "I believe the meaning of life is " -n 128`;
+	return `./main 
+	--hf-repo ${model.id} 
+	-m file.gguf 
+	-p "I believe the meaning of life is " 
+	-n 128`;
```
julien-c, line 54: ah yes the problem is that this will be displayed on newlines but won't be copy/pasteable in your terminal

I think maybe you need to escape them with two backslashes `\\`?

packages/tasks/src/local-apps.ts
```diff
@@ -47,10 +47,11 @@ function isGgufModel(model: ModelData) {
...
 const snippetLlamacpp = (model: ModelData): string => {
-	return `./main \
-	--hf-repo ${model.id} \
-	-m file.gguf \
-	-p "I believe the meaning of life is " -n 128`;
+	return `./main \\
+	--hf-repo "${model.id}" \\
+	-m file.gguf \\
+	-p "I believe the meaning of life is " \\
+	-n 128`;
```
coyotte508, line 54: nit:

```suggestion
	return [".main", `--hf-repo "${model.id}"`, "-m file.gguf",  '-p "I believe the meaning of life is "', "-n 128"].join(' \\\n');
```

100% optional

coyotte508 (approved):

packages/tasks/src/local-apps.ts
krampstudio: keeping like it is for consistency, ie 
https://github.com/huggingface/huggingface.js/blob/d1c0cb8604db4f504fa33b233fc12b9a501abc17/packages/tasks/src/snippets/curl.ts#L5-L11

PR #672, opened 2024-05-15
[tasks] deeplink for drawthings make it conditional on model repo type

julien-c:


packages/tasks/src/local-apps.ts
```diff
@@ -105,7 +105,13 @@ export const LOCAL_APPS = {
 		 * random function, will need to refine the actual conditions:
 		 */
 		displayOnModelPage: (model) => model.library_name === "diffusers" && model.pipeline_tag === "text-to-image",
```
apolinario, line 107:
```suggestion
		displayOnModelPage: (model) => model.library_name === "diffusers" && (model.pipeline_tag === "text-to-image" || model.tags.includes("lora")),
```
As per https://huggingface.slack.com/archives/C03RVEGLNBW/p1715780731909589?thread_ts=1713958091.775499&cid=C03RVEGLNBW

julien-c:
ok, ready for review!

julien-c:
cc @liuliu too in case you can review

packages/tasks/src/local-apps.ts
```diff
@@ -100,12 +100,14 @@ export const LOCAL_APPS = {
...
-		comingSoon: true,
-		/**
-		 * random function, will need to refine the actual conditions:
-		 */
-		displayOnModelPage: (model) => model.library_name === "diffusers" && model.pipeline_tag === "text-to-image",
-		deeplink: (model) => new URL(`drawthings://open_from_hf?model=${model.id}`),
+		displayOnModelPage: (model) => model.library_name === "diffusers" && (model.pipeline_tag === "text-to-image" || model.tags.includes("lora")),
+		deeplink: (model) => {
+			if (model.tags.includes("lora")) {
+				return new URL(`https://drawthings.ai/import/diffusers/pipeline.load_lora_weights?repo_id=${model.id}`);
```
liuliu, line 106: It seems this only works in Safari (Universal Link). I will include scheme based link support in the next update and also publish our fallback page soon (design ready, need to code up the HTML page).

The scheme based link would look like: `drawthings:///import/diffusers/pipeline.from_pretrained?repo_id=${model.id}` (similarly for LoRA weights). (note that it is /// because the host name is empty).

Our fallback page available at https://drawthings.ai/import/diffusers/pipeline.load_lora_weights will look roughly like this:
Screenshot 2024-05-15 at 4 14 32 PM

julien-c: very cool!

should we merge that PR i.e. we send users to your website and from there, you attempt to open the drawthings:/// url yourselves, correct?

Or do you want us to open the custom protocol ourselves (and fallback to the https:// page)?

julien-c: will merge this one so we can already test in production, but feel free to open a new PR or suggest changes

julien-c:
will merge this one so we can already test in production, but feel free to open a new PR or suggest changes

PR #677, opened 2024-05-16
:adhesive_bandage: local-apps : remove extra space from llama.cpp snippet

ggerganov:
The extra space at the end of the `--prompt` input can mess up the tokenization, so it is better to remove it.

To illustrate, here is how `"I believe the meaning of life is "` tokenizes:

```
main: prompt: 'I believe the meaning of life is '
main: number of tokens in prompt = 9
     2 -> '<bos>'
   590 -> ' I'
  4564 -> ' believe'
   573 -> ' the'
  6996 -> ' meaning'
   576 -> ' of'
  1913 -> ' life'
   603 -> ' is'
235248 -> ' '
```

There is an extra whitespace token 235248 at the end due to the extra space.

### DIFF
--- a/packages/tasks/src/local-apps.ts
+++ b/packages/tasks/src/local-apps.ts
@@ -50,7 +50,7 @@ const snippetLlamacpp = (model: ModelData): string => {
 	return `./main \\
 	--hf-repo "${model.id}" \\
 	-m file.gguf \\
-	-p "I believe the meaning of life is " \\
+	-p "I believe the meaning of life is" \\
 	-n 128`;
 };
 


pcuenca:
Interesting, I sometimes use it on purpose to signal the end of the previous word.

ggerganov:
I think the correct way to force a new word immediately after the prompt is through sampling restrictions (e.g. grammar) - do not sample a token that doesn't start with a space. Adding the space explicitly is almost never a good idea since you get this unexpected token which the model didn't "see" much during training

pcuenca:
Yes, makes sense, thank you!

pcuenca (approved):

coyotte508 (approved):
will leave it up to someone else to merge

Vaibhavs10 (approved):
Yay! Thanks for fixing this Georgi! 

Note: the failing test is not related to this PR.

mishig25 (approved):

Issue #221, opened 2023-07-14
Remove dependency on hash-wasm


coyotte508:
For example https://gist.github.com/coyotte508/3cfa9b011e819bcb4ba9286798c7d33c is the code of a worker, using `hash-wasm`'s internals, but in a standalone file.

It will improve bundled size too, but the main draw is having 0 dependencies.

PR #678, opened 2024-05-16
Hub: Move hash-wasm into vendor

ngxson:
For now, I have no intent to merge it. Just want to share the idea.

This is a small attempt to completely move `hash-wasm` into `vendor` code directory.

The PR consists of 3 parts:
- Add `scripts/pull-vendor.ts` to download and save the vendor code into the repo
- Import `vendor/hash-wasm` into non-web worker version
- Import `vendor/hash-wasm` into web worker version

Feel free to leave some feedback if you want.


coyotte508:
Thanks! you can add the `vendor/hash-wasm` folder to `.eslintignore` to pass the lint CI.

coyotte508:
Related issue: https://github.com/huggingface/huggingface.js/issues/221

ngxson:
@coyotte508 Thanks for the direction. I didn't notice that there is already an issue for this.

However, I'm still a bit skeptical with this approach:
- We duplicate the `hash-wasm` code, because it is bundled twice (once as module code, once as web worker code)
- There's still a compiled wasm blob which is not compiled by us

I'm replacing it with another PR that resolves these 2 problems.

PR #679, opened 2024-05-16
gguf: Add type definitions for `split.*` metadata + sanity check

ngxson:
This PR add correct type definitions for GGUF `split.*`, and use them in `ggufAllShards` for sanity checks.

The sanity checks does not really serves real-life purpose, since it is very unlikely that users mess up with file name. For now, it mostly to demonstrate how to use these metadata keys.

This PR also modify output format of `ggufAllShards`. Because metadata is saved in the first split only, we can simply discard metadata of other splits.

For more details, please have a look on `gguf-split` source code:

https://github.com/ggerganov/llama.cpp/blob/24ecb58168dce81646c2ed425690a106591c8c6d/examples/gguf-split/gguf-split.cpp#L223-L225

### DIFF
--- a/packages/gguf/src/gguf.ts
+++ b/packages/gguf/src/gguf.ts
@@ -400,8 +400,9 @@ export async function ggufAllShards(
 		 */
 		fetch?: typeof fetch;
 		additionalFetchHeaders?: Record<string, string>;
+		allowLocalFile?: boolean;
 	}
-): Promise<{ shards: GGUFParseOutput[]; parameterCount: number }> {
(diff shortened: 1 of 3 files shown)


julien-c (commented):
maybe of interest to @phymbert too

julien-c:
and btw related a bit to https://github.com/ggerganov/ggml/pull/826

coyotte508:
what's the status of this? (btw feel free to merge PRs in `huggingface/gguf`)

PR #683, opened 2024-05-17
Tasks: local apps: multiple snippets and build instructions for llama.cpp

krampstudio:
cc @Vaibhavs10

### DIFF
--- a/packages/tasks/src/local-apps.ts
+++ b/packages/tasks/src/local-apps.ts
@@ -38,7 +38,7 @@ export type LocalApp = {
 			/**
 			 * And if not (mostly llama.cpp), snippet to copy/paste in your terminal
 			 */
-			snippet: (model: ModelData) => string;
+			snippet: (model: ModelData) => string | string[];
 	  }
 );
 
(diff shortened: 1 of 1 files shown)


packages/tasks/src/local-apps.ts
```diff
@@ -38,20 +38,29 @@ export type LocalApp = {
...
 	return model.tags.includes("gguf");
 }
 
-const snippetLlamacpp = (model: ModelData): string => {
-	return `./main \\
+const snippetLlamacpp = (model: ModelData): string[] => {
+	return [
+		`
+## Install and build llama.cpp with curl support
+git clone https://github.com/ggerganov/llama.cpp.git
```
gary149, line 53: could add "&&" between each instructions? not sure it's a good idea :eyes:

julien-c (approved):

packages/tasks/src/local-apps.ts
krampstudio: no strong opinion on my side

packages/tasks/src/local-apps.ts
```diff
@@ -38,20 +38,29 @@ export type LocalApp = {
...
-	return `./main \\
+const snippetLlamacpp = (model: ModelData): string[] => {
+	return [
+		`
+## Install and build llama.cpp with curl support
+git clone https://github.com/ggerganov/llama.cpp.git 
+cd llama.cpp
+LLAMA_CURL=1 make
+`,
+		`## Load and run the model
```
gary149, line 57: _Download then run the model_ instead 👀

PR #687, opened 2024-05-20
Add ollama

boxabirds:
Very popular local LLM inference server.

### DIFF
--- a/packages/tasks/src/local-apps.ts
+++ b/packages/tasks/src/local-apps.ts
@@ -1,5 +1,6 @@
 import type { ModelData } from "./model-data";
 import type { PipelineType } from "./pipelines";
+import { parseGGUFQuantLabel } from "@huggingface/gguf";
 
 export interface LocalAppSnippet {
 	/**
@@ -126,10 +127,15 @@ const snippetLlamacpp = (model: ModelData, filepath?: string): LocalAppSnippet[]
 	];
 };
 
-const snippetOllama = (model: ModelData, ollamatag?: string): string => {
-	return `ollama run hf.co/${model.id}${ollamatag ?? "{{OLLAMA_TAG}}"}`;
+const snippetOllama = (model: ModelData, filepath?: string): string => {
+	if (filepath) {
+		const quantLabel = parseGGUFQuantLabel(filepath);
+		const ollamatag = quantLabel ? `:${quantLabel}` : "";
+		return `ollama run hf.co/${model.id}${ollamatag}`;
+	}
+	return `ollama run hf.co/${model.id}{{OLLAMA_TAG}}`;
 };
-``
+
 const snippetLocalAI = (model: ModelData, filepath?: string): LocalAppSnippet[] => {
 	const command = (binary: string) =>
 		["# Load and run the model:", `${binary} huggingface://${model.id}/${filepath ?? "{{GGUF_FILE}}"}`].join("\n");
--- a/packages/tasks/src/local-apps.ts
+++ b/packages/tasks/src/local-apps.ts
@@ -126,6 +126,10 @@ const snippetLlamacpp = (model: ModelData, filepath?: string): LocalAppSnippet[]
 	];
 };
 
+const snippetOllama = (model: ModelData, ollamatag?: string): string => {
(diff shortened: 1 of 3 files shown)


boxabirds:
Right so the deeplink doesn’t work but the other fields should be ok.

osanseviero:
The only goal of this button is to work with deeplinks. The button serves no other purpose

Vaibhavs10 (commented):
Hi @boxabirds - Just a heads-up we are thinking about this a bit more in our current sprint. We might have some exciting updates soon! 🤗

julien-c:
i thinik we could revive this PR but switch to a copy-pasteable command (like llama.cpp) rather than a deeplink

julien-c:
@boxabirds ok for you if @mishig25 from our team takes over your PR and completes it?

packages/tasks/src/local-apps.ts
```diff
@@ -389,6 +399,13 @@ export const LOCAL_APPS = {
 		displayOnModelPage: (model) => model.library_name === "diffusers" && model.pipeline_tag === "text-to-image",
 		deeplink: (model) => new URL(`https://models.invoke.ai/huggingface/${model.id}`),
 	},
+	ollama: {
+		prettyLabel: "Ollama",
+		docsUrl: "https://ollama.com",
```
ngxson, line 404: Probably have dedicated docs on the hub for this, but we can make a follow-up PR later on

ngxson (approved):

packages/tasks/src/local-apps.ts
```diff
@@ -143,6 +144,15 @@ const snippetNodeLlamaCppCli = (model: ModelData, filepath?: string): LocalAppSn
 	];
 };
 
+const snippetOllama = (model: ModelData, filepath?: string): string => {
+	if (filepath) {
+		const quantLabel = parseGGUFQuantLabel(filepath);
+		const ollamatag = quantLabel ? `:${quantLabel}` : "";
+		return `ollama run hf.co/${model.id}${ollamatag}`;
+	}
+	return `ollama run hf.co/${model.id}{{OLLAMA_TAG}}`;
```
julien-c, line 153: maybe document it line 55 (where you document `GGUF_FILE`) (can be in a later PR if you prefer)

julien-c (approved):

packages/tasks/src/local-apps.ts
```diff
@@ -389,6 +399,13 @@ export const LOCAL_APPS = {
 		displayOnModelPage: (model) => model.library_name === "diffusers" && model.pipeline_tag === "text-to-image",
 		deeplink: (model) => new URL(`https://models.invoke.ai/huggingface/${model.id}`),
 	},
+	ollama: {
+		prettyLabel: "Ollama",
+		docsUrl: "https://ollama.com",
+		mainTask: "text-generation",
+		displayOnModelPage: (model) => model.tags.includes("conversational") && isLlamaCppGgufModel(model),
```
julien-c, line 406: rm conversational no?

mishig25: handled in cbf44d6165284d14d74a4bb644278032176967ef as ollama also supports embedding & next-token prediction models (besides the conversational ones)

packages/tasks/src/local-apps.ts
mishig25: handled in c2691b00d212a2f5b811ab9136fc040724d63043

mishig25:
@coyotte508 could you help me with correctly importing hfjs/gguf inside hfjs/tasks? 
I did https://github.com/huggingface/huggingface.js/pull/687/commits/acae2361b16331952db23ed6134904dd10b87aa1 & https://github.com/huggingface/huggingface.js/pull/687/commits/b34e76b6d36bb1af9dfafcbea128110298972a59 with command 
```
cd packages/tasks
pnpm add @huggingface/gguf@workspace:^
```

However, the tests are failing & `huggingface/gguf` is not visible inside `huggingface/tasks`

julien-c:
> (but not thrilled to add a dep to `@huggingface/tasks`)

im guessing we could merge tasks and gguf...

PR #693, opened 2024-05-21
Adds vLLM as Option for Local App

EliMCosta:
Adds vLLM as option for "Local apps" in Hugginface


packages/tasks/src/local-apps.ts
```diff
@@ -63,6 +63,41 @@ LLAMA_CURL=1 make
...
+		`
+## Deploy with docker (needs Docker installed) a gated model (please, request access in Hugginface's model repo):
+docker run --runtime nvidia --gpus all \
+    --name my_vllm_container \
+    -v ~/.cache/huggingface:/root/.cache/huggingface \
+    --env "HUGGING_FACE_HUB_TOKEN=<secret>" \
+    -p 8000:8000 \
+    --ipc=host \
+    vllm/vllm-openai:latest \
+    --model mistralai/Mistral-7B-Instruct-v0.1
```
julien-c, line 77:
```suggestion
    --model ${model.id}
```

packages/tasks/src/local-apps.ts
```diff
@@ -63,6 +63,41 @@ LLAMA_CURL=1 make
...
+`,
+`
+## Load and run the model
+docker exec -it my_vllm_container bash -c "python -m vllm.entrypoints.openai.api_server --model mistralai/Mistral-7B-Instruct-v0.1 --dtype auto --api-key token-abc123"
```
julien-c, line 81:
```suggestion
docker exec -it my_vllm_container bash -c "python -m vllm.entrypoints.openai.api_server --model ${model.id} --dtype auto --api-key token-abc123"
```

packages/tasks/src/local-apps.ts
```diff
@@ -63,6 +63,41 @@ LLAMA_CURL=1 make
...
+`,
+`
+## Call the server using the official OpenAI Python client library, or any other HTTP client
+from openai import OpenAI
+client = OpenAI(
+    base_url="http://localhost:8000/v1",
+    api_key="token-abc123",
+)
+completion = client.chat.completions.create(
+  model="mistralai/Mistral-7B-Instruct-v0.1",
```
julien-c, line 91: same here

packages/tasks/src/local-apps.ts
```diff
@@ -82,6 +117,13 @@ export const LOCAL_APPS = {
 		displayOnModelPage: isGgufModel,
 		snippet: snippetLlamacpp,
 	},
+	"vllm": {
+		prettyLabel: "vLLM",
+		docsUrl: "https://docs.vllm.ai",
+		mainTask: "text-generation",
+		displayOnModelPage: isGptqModel && isAwqModel,
```
julien-c, line 124: how would you define those methods?

EliMCosta: > how would you define those methods?

In fact, the suggested vLLM method deploys the non-quantized version from the Hugginface repository. All examples of type "text-generation" in the code are GGUF. Any suggestion?

simon-mo: Thank you for the PR! Concretely we support a set of `architectures` which is readable from the model data https://github.com/huggingface/huggingface.js/blob/04a0eb4a01f22f61e9a0b87897c448b914ab1833/packages/tasks/src/model-data.ts#L40

https://github.com/vllm-project/vllm/blob/757b62c49560baa6f294310a53032348a0d95939/vllm/model_executor/models/__init__.py#L13-L63

And for quantization method we can read in `config.quantization_config.quant_method` which we support awq, gptq, aqlm, and marlin

https://huggingface.co/TheBloke/zephyr-7B-alpha-AWQ/blob/main/config.json#L28

julien-c: awesome @simon-mo, super clear!

julien-c: i've pushed `2123430` on this PR to type `config.quantization_config.quant_method` which we will parse & pass from the Hub

EliMCosta: > i've pushed [2123430](https://github.com/huggingface/huggingface.js/commit/2123430a737a92938d41d22b5fc9229b384a12bc) on this PR to type `config.quantization_config.quant_method` which we now parse & pass from the Hub

I made some changes, I need your help to review

krampstudio:
@EliMCosta do you have the vLLM icon as a square SVG?

EliMCosta:
> @EliMCosta do you have the vLLM icon as a square SVG?

vllm-logo

packages/tasks/src/local-apps.ts
```diff
@@ -42,8 +42,29 @@ export type LocalApp = {
 	  }
 );
 
-function isGgufModel(model: ModelData) {
-	return model.tags.includes("gguf");
+function isGgufModel(model: ModelData): boolean {
+    return model.config?.quantization_config?.quant_method === "gguf";
```
krampstudio, line 46: in case some repo doesn't have a quantization config file
```suggestion
    return model.config?.quantization_config?.quant_method === "gguf" || model.tags.includes("gguf");
```

julien-c: does vLLM support all GGUF files @simon-mo?

krampstudio: @julien-c it's not used by the [vLLM integration](https://github.com/huggingface/huggingface.js/pull/693/files#diff-98bf76e04efd0b5b0e0ea40e3d8d8ade7b129f461e2f7ff45ff3ff6e567c9302R142) but it's used by other apps and was modified in the meantime (see https://github.com/huggingface/huggingface.js/pull/693/commits/43de0e9d71804d330f6a4c17c3ac0366df93435d)

julien-c: ah ok!

packages/tasks/src/local-apps.ts
```diff
@@ -42,8 +42,29 @@ export type LocalApp = {
...
+    return model.config?.quantization_config?.quant_method === "aqlm";
+}
+
+function isMarlinModel(model: ModelData): boolean {
+    return model.config?.quantization_config?.quant_method === "marlin";
+}
+
+function isFullModel(model: ModelData): boolean {
+    // Assuming a full model is identified by not having a quant_method
+    return !model.config?.quantization_config?.quant_method;
```
krampstudio, line 67: `isFullModel` creates a lot of false positive. 

Instead we can maybe check against the [supported architectures](https://docs.vllm.ai/en/stable/models/supported_models.html) as suggested by @simon-mo 

Something like this 
```
const VLLM_SUPPORTED_ARCHS = [
    "AquilaForCausalLM", "ArcticForCausalLM", "BaiChuanForCausalLM", "BloomForCausalLM", ...
];
model.config?.architectures?.some((arch) => VLLM_SUPPORTED_ARCHS.includes(arch)
``` 

cc @julien-c

mgoin: You can query the vLLM package for this list once you have it installed:
```
>>> from vllm import ModelRegistry
>>> ModelRegistry.get_supported_archs()
['AquilaModel', 'AquilaForCausalLM', 'BaiChuanForCausalLM', 'BaichuanForCausalLM', 'BloomForCausalLM', 'ChatGLMModel', 'ChatGLMForConditionalGeneration', 'CohereForCausalLM', 'DbrxForCausalLM', 'DeciLMForCausalLM', 'DeepseekForCausalLM', 'FalconForCausalLM', 'GemmaForCausalLM', 'GPT2LMHeadModel', 'GPTBigCodeForCausalLM', 'GPTJForCausalLM', 'GPTNeoXForCausalLM', 'InternLMForCausalLM', 'InternLM2ForCausalLM', 'JAISLMHeadModel', 'LlamaForCausalLM', 'LlavaForConditionalGeneration', 'LlavaNextForConditionalGeneration', 'LLaMAForCausalLM', 'MistralForCausalLM', 'MixtralForCausalLM', 'QuantMixtralForCausalLM', 'MptForCausalLM', 'MPTForCausalLM', 'MiniCPMForCausalLM', 'OlmoForCausalLM', 'OPTForCausalLM', 'OrionForCausalLM', 'PhiForCausalLM', 'Phi3ForCausalLM', 'QWenLMHeadModel', 'Qwen2ForCausalLM', 'Qwen2MoeForCausalLM', 'RWForCausalLM', 'StableLMEpochForCausalLM', 'StableLmForCausalLM', 'Starcoder2ForCausalLM', 'ArcticForCausalLM', 'XverseForCausalLM', 'Phi3SmallForCausalLM', 'MistralModel']
```

youkaichao:
any progress here? I think we can add a tag vLLM, so that when a model is supported by vLLM, it can have a vLLM tag, like https://huggingface.co/meta-llama/Llama-2-7b-hf .

krampstudio:
@simon-mo @EliMCosta @mgoin we're back on the integration. I've merged it with #744 

Let me know if the snippets are ok for you.
It renders like this:

Screenshot 2024-09-02 at 16 59 52
Screenshot 2024-09-02 at 17 00 10

youkaichao:
@krampstudio I would recommend using the new cli `vllm serve` instead of the long `python -m vllm.entrypoints.xxx`

krampstudio:
@youkaichao I've updated the snippets (and the screenshots in the comment above)

packages/tasks/src/local-apps.ts
```diff
@@ -58,9 +58,28 @@ export type LocalApp = {
 	  }
 );
 
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
-function isGgufModel(model: ModelData) {
-	return model.tags.includes("gguf");
+function isGgufModel(model: ModelData): boolean {
+	return model.config?.quantization_config?.quant_method === "gguf" || model.tags.includes("gguf");
```
julien-c, line 62: are we sure about the first part?

julien-c (commented):
looks in good shape!

packages/tasks/src/local-apps.ts
```diff
@@ -127,6 +146,47 @@ const snippetLocalAI = (model: ModelData, filepath?: string): LocalAppSnippet[]
...
+		`		"messages": [`,
+		`			{"role": "user", "content": "Hello!"}`,
+		`		]`,
+		`	}'`,
+	];
+	return [
+		{
+			title: "Install from pip",
+			setup: ["# Install vLLM from pip:", "pip install vllm"].join("\n"),
+			content: ["# Load and run the model:", `vllm serve --model "${model.id}"`, ...runCommand].join("\n"),
```
mgoin, line 166:
```suggestion
			content: ["# Load and run the model:", `vllm serve "${model.id}"`, ...runCommand].join("\n"),
```

packages/tasks/src/local-apps.ts
krampstudio: no, let's rely only on the tag

pcuenca (commented):
Looks good!

PR #694, opened 2024-05-22
Add Titan RTX to list

cakiki:
source: https://www.nvidia.com/content/dam/en-zz/Solutions/titan/documents/titan-rtx-for-creators-us-nvidia-1011126-r6-web.pdf

Wasn't sure whether to call it Titan RTX (official name) or RTX Titan (nicer alphabetical UX for the dropdown) so I went with the latter.

### DIFF
--- a/packages/tasks/src/hardware.ts
+++ b/packages/tasks/src/hardware.ts
@@ -108,6 +108,10 @@ export const SKUS = {
 				tflops: 30.6,
 				memory: [12, 10],
 			},
+			"RTX Titan": {
+				tflops: 16.3,
+				memory: [24],
+			},			
 		},
 		AMD: {
 			MI300: {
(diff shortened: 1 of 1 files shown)


packages/tasks/src/hardware.ts
```diff
@@ -108,6 +108,10 @@ export const SKUS = {
 				tflops: 30.6,
 				memory: [12, 10],
 			},
+			"RTX Titan": {
+				tflops: 16.3,
```
johko, line 112: When based on FP16 performance (as mentioned in the comments further above), this should be the right value
(checked on https://www.techpowerup.com/gpu-specs/titan-rtx.c3311)
```suggestion
				tflops: 32.62,
```

julien-c (approved):
yay! 

i own two of those so i'll be happy to have them in my HF inventory

PR #697, opened 2024-05-22
Update hardware.ts to inclide RTX 3070 Ti Laptop

mattezell:


julien-c (approved):

pcuenca (approved):

packages/tasks/src/hardware.ts
```diff
@@ -108,6 +108,10 @@ export const SKUS = {
 				tflops: 30.6,
 				memory: [12, 10],
 			},
+			"RTX 3070 Ti Laptop": {
+				tflops: 16.60,
```
coyotte508, line 112:
```suggestion
				tflops: 16.6,
```

mattezell: Sorry about that... I should have pulled down the source and followed the flow, including linting - I was lazy and just used the web editor. Thanks for the fix.

PR #698, opened 2024-05-22
feat: higher memory options added.

aemreusta:
256 and 512 memory options added to DEFAULT_MEMORY_OPTIONS


julien-c (approved):
wow you have 512 GB CPU RAM?

PR #701, opened 2024-05-22
Add RTX 2060 Mobile specs

johko:
Adding a GPU poor option :wink: 
Specs with FP16 according to this page: https://www.techpowerup.com/gpu-specs/geforce-rtx-2060-mobile.c3348

### DIFF
--- a/packages/tasks/src/hardware.ts
+++ b/packages/tasks/src/hardware.ts
@@ -115,7 +115,7 @@ export const SKUS = {
       "RTX 2060 Mobile": {
 				tflops: 9.22,
 				memory: [6],
-      }
+      },
 		},
 		AMD: {
 			MI300: {
--- a/packages/tasks/src/hardware.ts
+++ b/packages/tasks/src/hardware.ts
@@ -31,7 +31,7 @@ export interface HardwareSpec {
 	memory?: number[];
 }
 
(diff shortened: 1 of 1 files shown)


julien-c (approved):

packages/tasks/src/hardware.ts
```diff
@@ -31,7 +31,7 @@ export interface HardwareSpec {
 	memory?: number[];
 }
 
-export const DEFAULT_MEMORY_OPTIONS = [8, 16, 24, 32, 40, 48, 64, 80, 96, 128];
+export const DEFAULT_MEMORY_OPTIONS = [6, 8, 16, 24, 32, 40, 48, 64, 80, 96, 128];
```
coyotte508, line 34: Should we change this line? It's not related IMO

pcuenca: I agree, that should not be necessary.

johko: yup, sorry somehow thought that was needed at first but you're totally right. reverted it

packages/tasks/src/hardware.ts
```diff
@@ -112,6 +112,10 @@ export const SKUS = {
 				tflops: 16.6,
 				memory: [8],
 			},
+      "RTX 2060 Mobile": {
+				tflops: 9.22,
+				memory: [6],
+      }
```
coyotte508, line 118:
```suggestion
      },
```

PR #703, opened 2024-05-22
Ignore unsupported pipelines for transformers

osanseviero:
If a model has a type without a corresponding pipeline in transformers, we don't show the `pipeline` snippet but we still show the `AutoModel` one. This is useful for example for `sentence-similarity` models

Screenshot from 2024-05-22 18-35-06
Screenshot from 2024-05-22 18-34-43

https://huggingface.slack.com/archives/C02EMARJ65P/p1716294199591799

### DIFF
--- a/packages/tasks/src/model-libraries-snippets.ts
+++ b/packages/tasks/src/model-libraries-snippets.ts
@@ -1,4 +1,5 @@
 import type { ModelData } from "./model-data";
+import { LIBRARY_TASK_MAPPING } from "./library-to-tasks";
 
 const TAG_CUSTOM_CODE = "custom_code";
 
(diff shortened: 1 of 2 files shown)


packages/tasks/src/model-libraries-snippets.ts
```diff
@@ -438,7 +439,7 @@ export const transformers = (model: ModelData): string[] => {
 		].join("\n");
 	}
 
-	if (model.pipeline_tag) {
+	if (model.pipeline_tag && LIBRARY_TASK_MAPPING.transformers?.includes(model.pipeline_tag)) {
```
tomaarsen, line 442: This looks very sensible 👍

julien-c (approved):
lgtm!

PR #706, opened 2024-05-22
Added NVIDIA V100 GPUs

nahidalam:
Added NVIDIA V100 GPUs

### DIFF
--- a/packages/tasks/src/hardware.ts
+++ b/packages/tasks/src/hardware.ts
@@ -136,6 +136,10 @@ export const SKUS = {
 				tflops: 32.62,
 				memory: [24],
 			},
+			"V100": {
+				tflops: 130,
+				memory: [32],
+			},
 		},
 		AMD: {
 			MI300: {


coyotte508:
Given its computational power it should probably be ordered above or below the A100 family cc @pcuenca ?

packages/tasks/src/hardware.ts
```diff
@@ -136,6 +136,10 @@ export const SKUS = {
 				tflops: 32.62,
 				memory: [24],
 			},
+			"V100": {
+				tflops: 130,
+				memory: [32],
```
pcuenca, line 141:
```suggestion
				memory: [32, 16],
```

There are variants with 16 GB of RAM: https://images.nvidia.com/content/technologies/volta/pdf/volta-v100-datasheet-update-us-1165301-r5.pdf

packages/tasks/src/hardware.ts
```diff
@@ -136,6 +136,10 @@ export const SKUS = {
 				tflops: 32.62,
 				memory: [24],
 			},
+			"V100": {
+				tflops: 130,
```
pcuenca, line 140: We are using the float16 performance as a reference, as it's an usual workload for ML. In this case, we can estimate it to be about double the [single-precision in this table](https://images.nvidia.com/content/technologies/volta/pdf/volta-v100-datasheet-update-us-1165301-r5.pdf), or the 28.26 reported [here](https://www.techpowerup.com/gpu-specs/tesla-v100-pcie-16-gb.c2957).

```suggestion
				tflops: 28.26,
```

pcuenca (commented):
Thanks for the contribution! We are using FP16 as the reference for tflops. In terms of placement, I think it makes sense to keep it after the Titan as it's less powerful and belongs to a previous generation.

coyotte508 (approved):

nahidalam:
Looks like the format check failed, though not sure for which line https://github.com/huggingface/huggingface.js/actions/runs/9208329481/job/25330626278?pr=706

packages/tasks/src/hardware.ts
```diff
@@ -160,6 +160,10 @@ export const SKUS = {
 				tflops: 32.62,
 				memory: [24],
 			},
+			"V100": {
```
coyotte508, line 163:
```suggestion
			V100: {
```

pcuenca:
Thanks @coyotte508 for the fix, sorry I missed on review :)

coyotte508:
> Thanks @coyotte508 for the fix, sorry I missed on review :)

Nothing to worry, I improved the CI to show the diff 

(maybe we could even add a bot that listens to @format comments :))

julien-c:
> Given its computational power it should probably be ordered above or below the A100 family cc @pcuenca ?

i think we should aim to order chronologically, WDYT?

pcuenca:
> order chronologically

yes, chronological family order sounds great, I can address it in a PR tomorrow

julien-c:
why not put it in the right place in this PR @coyotte508?

coyotte508:
> why not put it in the right place in this PR @coyotte508?

It's at its correct place, it was released in 2017 and the RTX titan above was released in 2018.

julien-c:
Ah cool, thanks

PR #708, opened 2024-05-23
added rtx 3050 mobile and also added 6GB to DEFAULT_MEMORY_OPTIONS

Om-Alve:
Added info about RTX 3050 mobile 6GB and also added 6GB to the DEFAULT_MEMORY_OPTIONS list

### DIFF
--- a/packages/tasks/src/hardware.ts
+++ b/packages/tasks/src/hardware.ts
@@ -154,8 +154,8 @@ export const SKUS = {
 			},
 			"RTX 3050 Mobile":{
 				tflops: 7.639,
-				memory:[6],
-			}
+				memory: [6],
+			},
 			"RTX 2060 Mobile": {
 				tflops: 9.22,
 				memory: [6],
--- a/packages/tasks/src/hardware.ts
+++ b/packages/tasks/src/hardware.ts
@@ -31,7 +31,7 @@ export interface HardwareSpec {
 	memory?: number[];
 }
 
-export const DEFAULT_MEMORY_OPTIONS = [8, 16, 24, 32, 40, 48, 64, 80, 96, 128, 256, 512];
+export const DEFAULT_MEMORY_OPTIONS = [6, 8, 16, 24, 32, 40, 48, 64, 80, 96, 128, 256, 512];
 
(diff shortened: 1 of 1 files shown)


coyotte508:
Thank you for the PR!

The `DEFAULT_MEMORY_OPTIONS` is used for CPU RAM, can you make a separate PR for this change? I'm not sure 6GB RAM is very common with latops

Om-Alve:
> The `DEFAULT_MEMORY_OPTIONS` is used for CPU RAM, can you make a separate PR for this change? I'm not sure 6GB RAM is very common with laptops

Ah, I'm sorry I thought that they were the default values for VRAM, should I undo that change in this PR?

packages/tasks/src/hardware.ts
```diff
@@ -128,6 +128,10 @@ export const SKUS = {
 				tflops: 12.74,
 				memory: [12, 8],
 			},
+			"RTX 3050 Mobile":{
+				tflops: 7.639,
+				memory:[6],
```
pcuenca, line 133: The first version was slower and had 4 GB: https://www.techpowerup.com/gpu-specs/geforce-rtx-3050-mobile.c3788

Let's keep this one for now (Numbers match this reference: https://www.techpowerup.com/gpu-specs/geforce-rtx-3050-mobile-refresh-6-gb.c3971)

pcuenca (approved):
Thanks a lot!

packages/tasks/src/hardware.ts
```diff
@@ -152,6 +152,10 @@ export const SKUS = {
...
+			}
```
pcuenca, line 158:
```suggestion
			},
```

packages/tasks/src/hardware.ts
```diff
@@ -152,6 +152,10 @@ export const SKUS = {
 				tflops: 12.74,
 				memory: [12, 8],
 			},
+			"RTX 3050 Mobile":{
+				tflops: 7.639,
+				memory:[6],
```
pcuenca, line 157:
```suggestion
				memory: [6],
```

PR #713, opened 2024-05-24
Add new text-to-image LocalApp  JoyFusion(macOS)

jiangdi0924:
JoyFusion is a native application for macOS , built upon Stable Diffusion and CoreML technologies. It can locally  run CoreML-based text-to-image models.

👇Here is Icon SVG

joyfusion-icon


pcuenca:
Hi @jiangdi0924, I'm not sure if you still want to proceed with this PR, but I tested it anyway and found that the deeplink does not work in my case. Is it pending an App Store update?

PR #719, opened 2024-05-28
Add Jellybox Local App

phudtran:
Hi Huggingface team!

[Jellybox](https://jellybox.com) is a local first app for running both language models, and image generation.
Here is a current features list:

1. Download models from Huggingface
2. In depth configuration for creating templates for chatting with language models
3. Allow language models to execute code in a container sandbox (prompted with unprompted coming soon)
4. Windows (Nvidia (CUDA) & AMD (ROCm)) and macOS support
5. Image generation templates, with per template image galleries
6. Deep link support

### DIFF
--- a/packages/tasks/src/local-apps.ts
+++ b/packages/tasks/src/local-apps.ts
@@ -110,15 +110,17 @@ export const LOCAL_APPS = {
 		displayOnModelPage: isGgufModel,
 		deeplink: (model) => new URL(`sanctum://open_from_hf?model=${model.id}`),
 	},
-	jellybox : {
+	jellybox: {
 		prettyLabel: "Jellybox",
 		docsUrl: "https://jellybox.com",
 		mainTask: "text-generation",
 		displayOnModelPage: (model) =>
 			isGgufModel(model) ||
-			model.library_name === "diffusers" && model.tags.includes("safetensors") && (model.pipeline_tag === "text-to-image" || model.tags.includes("lora")),
+			(model.library_name === "diffusers" &&
+				model.tags.includes("safetensors") &&
+				(model.pipeline_tag === "text-to-image" || model.tags.includes("lora"))),
 		deeplink: (model) => {
-			if(isGgufModel(model)) {
+			if (isGgufModel(model)) {
 				return new URL(`jellybox://llm/models/huggingface/LLM/${model.id}`);
 			} else if (model.tags.includes("lora")) {
 				return new URL(`jellybox://image/models/huggingface/ImageLora/${model.id}`);
--- a/packages/tasks/src/local-apps.ts
+++ b/packages/tasks/src/local-apps.ts
@@ -110,6 +110,23 @@ export const LOCAL_APPS = {
 		displayOnModelPage: isGgufModel,
 		deeplink: (model) => new URL(`sanctum://open_from_hf?model=${model.id}`),
 	},
+	jellybox : {
+		prettyLabel: "Jellybox",
+		docsUrl: "https://jellybox.com",
+		mainTask: "text-generation",
+		displayOnModelPage: (model) =>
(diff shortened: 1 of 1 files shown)


julien-c:
> Also, for apps that support multiple pipeline types, would you guys consider changing
> 
> ```
> mainTask: PipelineType;
> ```
> 
> to
> 
> ```
> tasks: [PipelineType];
> ```

This is only to categoize the app in our current settings UX though. So it's not super important imo

packages/tasks/src/local-apps.ts
```diff
@@ -103,6 +103,23 @@ export const LOCAL_APPS = {
...
+	jellybox : {
+		prettyLabel: "Jellybox",
+		docsUrl: "https://jellybox.com",
+		mainTask: "text-generation",
+		displayOnModelPage: (model) =>
+			isGgufModel(model) ||
+			model.library_name === "diffusers" && model.tags.includes("safetensors") && (model.pipeline_tag === "text-to-image" || model.tags.includes("lora")),
+		deeplink: (model) => {
+			if(isGgufModel(model)) {
+				return new URL(`jellybox://llm/models/huggingface/LLM/${model.id}`);
```
krampstudio, line 115: This will create for example `jellybox://llm/models/huggingface/LLM/microsoft/Phi-3-mini-4k-instruct-gguf`  
for me the deeplink opens Jellybox but doesn't select the model to download 
Screenshot 2024-05-30 at 16 27 41

It seems to be working in your video but with an internal id. Can you update the format?

phudtran: It's fixed in the latest update!
I was using the id from this previously.https://github.com/huggingface/huggingface.js/blob/f94376b3b763d53e4f54055c06d6ed34d219b003/packages/hub/src/lib/list-models.ts#L43

I'm currently working on converting the icon to a vector. Will reply here with the SVG once it's ready.

https://github.com/huggingface/huggingface.js/assets/22720475/04ea7f84-fb7d-4fc2-87e1-b34763c057e3

phudtran:
> Hey @phudtran do you have the icon in SVG?

I'm currently working on converting the icon to a vector. Will reply here with the SVG once it's ready.

phudtran:
@krampstudio here's the SVG

jellybox_icon_vector

phudtran:
Sorry for all the edits to the SVG. Illustrator was exporting with inline styles which caused issues with Github's CSP. Had to go in and manually change them to attributes.

packages/tasks/src/local-apps.ts
krampstudio: It works well with that version. Thanks!

packages/tasks/src/local-apps.ts
```diff
@@ -110,6 +110,23 @@ export const LOCAL_APPS = {
...
+			} else if (model.tags.includes("lora")) {
+				return new URL(`jellybox://image/models/huggingface/ImageLora/${model.id}`);
+			} else {
+				return new URL(`jellybox://image/models/huggingface/Image/${model.id}`);
```
julien-c, line 126: btw take a look at https://github.com/huggingface/huggingface.js/pull/738 in case you'd like to support opening from an optional specific file inside the repo! thanks!

phudtran: Thanks for the heads up! Taking a look.

krampstudio (approved):

krampstudio:
@phudtran this is now live on huggingface.co
Please don't hesitate to share the news with the Jellybox users

phudtran:
Thank you @krampstudio and @julien-c!

Quick question, is there any change I need to do on my side to get Jellybox to be selectable as the default application for both text generation and text-to-image?

image

PR #720, opened 2024-05-29
docs: improve llama.cpp install instructions.

Vaibhavs10:
Does two things:

1. Makes homebrew install the default and points users to build instructions for other platforms (note: our previous instructions were also only for Mac - this PRs makes the wording more generalised)
2. moves from `-m` to `--hf-file` this would make sure that the models are cached

### DIFF
--- a/packages/tasks/src/local-apps.ts
+++ b/packages/tasks/src/local-apps.ts
@@ -48,16 +48,14 @@ function isGgufModel(model: ModelData) {
 
 const snippetLlamacpp = (model: ModelData): string[] => {
 	return [
-		`
-## Install and build llama.cpp with curl support
-git clone https://github.com/ggerganov/llama.cpp.git 
-cd llama.cpp
-LLAMA_CURL=1 make
+		`## Install llama.cpp via brew or from source https://github.com/ggerganov/llama.cpp?tab=readme-ov-file#build
+
+brew install llama.cpp
 `,
 		`## Load and run the model
-./main \\
+llama \\
 	--hf-repo "${model.id}" \\
-	-m file.gguf \\
+	--hf-file file.gguf \\
 	-p "I believe the meaning of life is" \\
 	-n 128`,
 	];


pcuenca (approved):
Nice, thank you!

Vaibhavs10:
cc: @gary149 - can I get your blessings to merge this PR, too, please? 🙏

packages/tasks/src/local-apps.ts
```diff
@@ -48,16 +48,14 @@ function isGgufModel(model: ModelData) {
...
 const snippetLlamacpp = (model: ModelData): string[] => {
 	return [
-		`
-## Install and build llama.cpp with curl support
-git clone https://github.com/ggerganov/llama.cpp.git 
-cd llama.cpp
-LLAMA_CURL=1 make
+		`## Install llama.cpp via brew or from source https://github.com/ggerganov/llama.cpp?tab=readme-ov-file#build
+
+brew install llama.cpp
```
julien-c, line 53: i would keep the two options

julien-c, line 53:
```suggestion
		`## Install llama.cpp via brew
brew install llama.cpp

## or from source with curl support
git clone https://github.com/ggerganov/llama.cpp
cd llama.cpp
LLAMA_CURL=1 make
```

WDYT?

julien-c (approved):
lgtm with my suggestion, wdyt?

packages/tasks/src/local-apps.ts
Vaibhavs10: The only problem her is that 

```
git clone https://github.com/ggerganov/llama.cpp && cd llama.cpp && LLAMA_CURL=1 make  
```

would compile to metal only by default (so it'd have the same effect as brew)

People need to pass different flags for different backends

CUDA Build would require `LLAMA_CUDA=1` (ref: https://github.com/ggerganov/llama.cpp?tab=readme-ov-file#blas-build), different for BLAS, VULKAN, etc.

Maybe I'm overthink this, and the way you suggest is okay. but, thought it's good to flag this. Let me know what ya think!

pcuenca: I also thought that it'd be nice to give more options. We could potentially say after the `brew` command something like "This will enable super-fast Metal inference on a Mac. For more options or to target other backends, you can install from source with curl support. For example, to compile with CUDA on Linux..."

julien-c: edited my suggestion, WDYT?

Vaibhavs10: Yeah! This looks like a good compromise to me. I think the key is to say compile with `LLAMA_CURL`.

Vaibhavs10:
Sounds good to me, I will wait for @pcuenca to confirm that it looks good to him, too (since it's slightly deviating from his original idea) before merging.

pcuenca (approved):

Vaibhavs10:
I don't know why the E2E error fails, will wait for @coyotte508's guidance

osanseviero (approved):
Nice!  The tests pass after re-running.

PR #721, opened 2024-05-29
doc: update audio related tasks page.

Vaibhavs10:
cc: @sanchit-gandhi @ylacombe

### DIFF
--- a/packages/tasks/src/tasks/automatic-speech-recognition/about.md
+++ b/packages/tasks/src/tasks/automatic-speech-recognition/about.md
@@ -39,12 +39,12 @@ data = query("sample1.flac")
 You can also use libraries such as [transformers](https://huggingface.co/models?library=transformers&pipeline_tag=automatic-speech-recognition&sort=downloads), [speechbrain](https://huggingface.co/models?library=speechbrain&pipeline_tag=automatic-speech-recognition&sort=downloads), [NeMo](https://huggingface.co/models?pipeline_tag=automatic-speech-recognition&library=nemo&sort=downloads) and [espnet](https://huggingface.co/models?library=espnet&pipeline_tag=automatic-speech-recognition&sort=downloads) if you want one-click managed Inference without any hassle.
 
 ```python
+# pip install --upgrade transformers
+
 from transformers import pipeline
 
-with open("sample.flac", "rb") as f:
-  data = f.read()
+pipe = pipeline("automatic-speech-recognition", "openai/whisper-large-v3")
 
-pipe = pipeline("automatic-speech-recognition", "openai/whisper-large-v2")
 pipe("sample.flac")
 # {'text': "GOING ALONG SLUSHY COUNTRY ROADS AND SPEAKING TO DAMP AUDIENCES IN DRAUGHTY SCHOOL ROOMS DAY AFTER DAY FOR A FORTNIGHT HE'LL HAVE TO PUT IN AN APPEARANCE AT SOME PLACE OF WORSHIP ON SUNDAY MORNING AND HE CAN COME TO US IMMEDIATELY AFTERWARDS"}
 ```
@@ -57,7 +57,7 @@ import { HfInference } from "@huggingface/inference";
 const inference = new HfInference(HF_TOKEN);
 await inference.automaticSpeechRecognition({
 	data: await (await fetch("sample.flac")).blob(),
-	model: "openai/whisper-large-v2",
+	model: "openai/whisper-large-v3",
 });
 ```
 
--- a/packages/tasks/src/tasks/automatic-speech-recognition/about.md
+++ b/packages/tasks/src/tasks/automatic-speech-recognition/about.md
@@ -18,7 +18,7 @@ The use of Multilingual ASR has become popular, the idea of maintaining just a s
 
 ## Inference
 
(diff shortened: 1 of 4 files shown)


packages/tasks/src/tasks/automatic-speech-recognition/about.md
```diff
@@ -18,7 +18,7 @@ The use of Multilingual ASR has become popular, the idea of maintaining just a s
 
 ## Inference
 
-The Hub contains over [~9,000 ASR models](https://huggingface.co/models?pipeline_tag=automatic-speech-recognition&sort=downloads) that you can use right away by trying out the widgets directly in the browser or calling the models as a service using Inference Endpoints. Here is a simple code snippet to do exactly this:
+The Hub contains over [~17,000+ ASR models](https://huggingface.co/models?pipeline_tag=automatic-speech-recognition&sort=downloads) that you can use right away by trying out the widgets directly in the browser or calling the models as a service using Inference Endpoints. Here is a simple code snippet to do exactly this:
```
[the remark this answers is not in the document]
osanseviero: Yes, this is Serverless Inference API. And I agree the sentence afterwards "you can use libraries..." is misleading.

sanchit-gandhi (approved):
Thanks for the updates @Vaibhavs10! A few suggestions for the ASR task page:
1. In the [Inference](https://huggingface.co/tasks/automatic-speech-recognition#inference) section, before giving the Transformers code-snippet, should we also include instructions to pip install the library? Given the task page is targeted at beginners, IMO it's good to be more verbose here
```sh
pip install --upgrade transformers
```
2. There's an unnecessary op that we do in the Transformers code-snippet, which we can remove:
```diff
from transformers import pipeline

- with open("sample.flac", "rb") as f:
-  data = f.read()

pipe = pipeline("automatic-speech-recognition", "openai/whisper-large-v2")
pipe("sample.flac")
# {'text': "GOING ALONG SLUSHY COUNTRY ROADS AND SPEAKING TO DAMP AUDIENCES IN DRAUGHTY SCHOOL ROOMS DAY AFTER DAY FOR A FORTNIGHT HE'LL HAVE TO PUT IN AN APPEARANCE AT SOME PLACE OF WORSHIP ON SUNDAY MORNING AND HE CAN COME TO US IMMEDIATELY AFTERWARDS"}
```
3. For the datasets, instead of replacing `librispeech_asr` with `espnet/yodas`, should we replace `openslr` instead? `librispeech_asr` is quite well known among the ASR community, so it would be nice to keep it displayed there to make users aware it's present on the Hub (and can be used as an alternative to Kaldi LibriSpeech). Whereas `openslr` is not really used as a multilingual dataset anymore and `espnet/yodas` can be viewed as a replacement
4.  For the models, should we replace `facebook/s2t-small-mustc-en-fr-st` with `distil-whisper/distil-large-v3`? The former is again quite outdated (most users would be better off using `openai/whisper-large-v3` which we already promote as the first model on the list)

packages/tasks/src/tasks/automatic-speech-recognition/about.md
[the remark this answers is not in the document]
Vaibhavs10: Makes sense! and.. thanks for the suggestion!

Vaibhavs10:
Nice @sanchit-gandhi - addressed your review - Instead of distil-whisper I chose seamless-m4t just for diversity reasons. lmk if that works.

packages/tasks/src/tasks/text-to-speech/about.md
```diff
@@ -62,4 +62,5 @@ await inference.textToSpeech({
 - [A guide on Fine-tuning Whisper For Multilingual ASR with 🤗Transformers](https://huggingface.co/blog/fine-tune-whisper)
 - [Speech Synthesis, Recognition, and More With SpeechT5](https://huggingface.co/blog/speecht5)
 - [Optimizing a Text-To-Speech model using 🤗 Transformers](https://huggingface.co/blog/optimizing-bark)
+- [Train your own TTS models with Parler-TTS](https://github.com/huggingface/parler-tts)
```
ylacombe, line 65: Should we reorganize the useful resources section? 
There's the guide on fine-tuning whisper that should be put last since it's not directly related to TTS!

packages/tasks/src/tasks/text-to-speech/about.md
```diff
@@ -62,4 +62,5 @@ await inference.textToSpeech({
...
 -
```
ylacombe, line 66:
```suggestion
```

packages/tasks/src/tasks/text-to-speech/data.ts
```diff
@@ -4,8 +4,8 @@ const taskData: TaskDataCustom = {
 	canonicalId: "text-to-audio",
 	datasets: [
```
ylacombe, line 5: I currently don't see any datasets on the task page, even though there's supposed to be LibriTTS, am I missing something ?

ylacombe (approved):

packages/tasks/src/tasks/text-to-speech/about.md
Vaibhavs10: Done! Good shout!

packages/tasks/src/tasks/automatic-speech-recognition/about.md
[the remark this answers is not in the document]
osanseviero: Can you also change the paragraph after the code snippet please?

packages/tasks/src/tasks/text-to-speech/data.ts
osanseviero: Maybe due to the rename to https://huggingface.co/datasets/keithito/lj_speech

PR #729, opened 2024-06-01
Update GGUF quantization types

CISC:
Bring `GGMLQuantizationType` up to date; adds `I8`, `I16`, `I32`, `I64`, `F64`, `IQ1_M` and `BF16`.

Added in:
* ggerganov/llama.cpp#6045
* ggerganov/llama.cpp#6062
* ggerganov/llama.cpp#6302
* ggerganov/llama.cpp#6412

### DIFF
--- a/packages/gguf/src/types.ts
+++ b/packages/gguf/src/types.ts
@@ -29,6 +29,13 @@ export enum GGMLQuantizationType {
 	IQ3_S = 21,
 	IQ2_S = 22,
 	IQ4_XS = 23,
+	I8 = 24,
+	I16 = 25,
+	I32 = 26,
+	I64 = 27,
+	F64 = 28,
+	IQ1_M = 29,
+	BF16 = 30,
 }
 
 export enum GGUFValueType {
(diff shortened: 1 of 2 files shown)


packages/gguf/src/types.ts
```diff
@@ -29,6 +29,13 @@ export enum GGMLQuantizationType {
 	IQ3_S = 21,
 	IQ2_S = 22,
 	IQ4_XS = 23,
+	I8 = 24,
```
mishig25, line 32: @CISC could you paste here in the comments the source from ggml repo for these newly added types?

CISC: Done

mishig25 (approved):
lgtm ! Thanks so much for the contribution. I think if you want, you can even document `BF16` in the ggml format docs https://github.com/ggerganov/ggml/blob/master/docs/gguf.md?plain=1#L146-L151 😊

julien-c (commented):
Do we need to regenerate the table in https://huggingface.co/docs/hub/gguf?

PR #737, opened 2024-06-05
[llama.cpp snippet] Present both options

mishig25:
follow up to https://github.com/huggingface/huggingface.js/pull/720

just like we show two snippets for transformers (pipeline & automodel) (example [here](https://huggingface.co/mistralai/Codestral-22B-v0.1?library=transformers))

Screenshot 2024-06-05 at 14 51 22

we should show the two options for llama.cpp clearly as well. The two options are: installing from brew and installing from source

Screenshot 2024-06-05 at 15 37 22

### DIFF
--- a/packages/tasks/src/local-apps.ts
+++ b/packages/tasks/src/local-apps.ts
@@ -51,7 +51,7 @@ const snippetLlamacpp = (model: ModelData): string[] => {
 		`# Option 1: use llama.cpp with brew
 brew install llama.cpp
 
-# Build llama.cpp then load and run the model
+# Load and run the model
 llama \\
 	--hf-repo "${model.id}" \\
 	--hf-file file.gguf \\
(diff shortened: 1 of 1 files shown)


julien-c (approved):
lgtm

packages/tasks/src/local-apps.ts
```diff
@@ -48,20 +48,25 @@ function isGgufModel(model: ModelData) {
...
 brew install llama.cpp
 
-## or from source with curl support
-## see llama.cpp README for compilation flags to optimize for your hardware
-git clone https://github.com/ggerganov/llama.cpp
-cd llama.cpp
-LLAMA_CURL=1 make
-`,
-		`## Load and run the model
+# Load and run the model
```
gary149, line 54:
```suggestion
# Build llama.cpp then load and run the model
```

gary149: No it was better before can you revert @mishig25 ?

mishig25: https://github.com/huggingface/huggingface.js/pull/737/commits/1ef3a5e8a5577e7c1eb73d546ff0388794af43aa

PR #743, opened 2024-06-06
add a list of dataset libraries

severo:
see
https://github.com/huggingface/moon-landing/pull/9352#discussion_r1627957713

### DIFF
--- a/packages/tasks/src/dataset-libraries.ts
+++ b/packages/tasks/src/dataset-libraries.ts
@@ -20,10 +20,6 @@ export interface DatasetLibraryUiElement {
 	 * URL to library's docs
 	 */
 	docsUrl?: string;
-	/**
-	 * URL to library's docs in the Hugging Face documentation
-	 */
-	hfDocsUrl?: string;
 }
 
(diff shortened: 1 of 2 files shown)


packages/tasks/src/dataset-libraries.ts
```diff
@@ -0,0 +1,67 @@
...
+	 */
+	repoUrl:     string;
+	/**
+	 * URL to library's docs
+	 */
+	docsUrl?:    string;
+	/**
+	 * URL to library's docs in the Hugging Face documentation
+	 */
+	hfDocsUrl?:  string;
```
julien-c, line 26: imo `docsUrl` can always point to hub docs, which in turn point to the library docs (like for models)

julien-c: that way we have one less URL to define

julien-c (approved):
lgtm with my suggestion

packages/tasks/src/dataset-libraries.ts
severo: done

PR #745, opened 2024-06-10
feat(ci): add trufflehog secrets detection

McPatate:
### What does this PR do?

Adding a GH action to scan for leaked secrets on each commit.

### DIFF
--- a/.github/workflows/trufflehog.yml
+++ b/.github/workflows/trufflehog.yml
@@ -1,0 +1,21 @@
+on:
+  push:
+
+name: Secret Leaks
+
+permissions:
+  contents: read
+  id-token: write
+  issues: write
+  pull-requests: write
+
+jobs:
+  trufflehog:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v4
+      with:
+        fetch-depth: 0
+    - name: Secret Scanning
+      uses: trufflesecurity/trufflehog@main


.github/workflows/trufflehog.yml
```diff
@@ -0,0 +1,15 @@
...
+name: Secret Leaks
+
+jobs:
+  trufflehog:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v4
+      with:
+        fetch-depth: 0
```
coyotte508, line 13: Is there a reason to fetch everything?

McPatate: Though we could make it so it goes through the whole history if we wanted, cf the actions code: https://github.com/trufflesecurity/trufflehog/blob/3a029ea1932f93109bf9310607ed0f426483a891/action.yml

coyotte508: Shallow cloning sounds good!

About making it go through the whole history, pretty sure nothing's been merged with secrets ever, but we can run it once if we want (no need to run it every time)

coyotte508:
Job seems to fail: https://github.com/huggingface/huggingface.js/actions/runs/9517113121/job/26234899723

McPatate:
Looks like you used a `'` in your commit message and the script wasn't expecting it. Let me check if I can find a way to circumvent.

PR #746, opened 2024-06-10
Tasks: update snippet

krampstudio:
follow up of #699

### DIFF
--- a/packages/tasks/src/local-apps.ts
+++ b/packages/tasks/src/local-apps.ts
@@ -37,6 +37,7 @@ export type LocalApp = {
 	| {
 			/**
 			 * And if not (mostly llama.cpp), snippet to copy/paste in your terminal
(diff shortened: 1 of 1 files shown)


packages/tasks/src/local-apps.ts
```diff
@@ -54,7 +55,7 @@ brew install llama.cpp
 # Load and run the model
 llama \\
 	--hf-repo "${model.id}" \\
-	--hf-file file.gguf \\
+	--hf-file {{GGUF_FILE}} \\
```
osanseviero, line 58: Nice, thanks! 

Do you have a screenshot of how this looks with multiple GGUF files?

krampstudio: yes, 

with multiple files: 
Capture d’écran 2024-06-10 à 12 12 25

with a single file:

gary149 (approved):

PR #748, opened 2024-06-10
Add recurrentgemma and gemma.cpp as libraries

osanseviero:


Vaibhavs10 (approved):
LGTM, confirmed by looking through all the repos here: https://huggingface.co/models?other=gemma.cpp

They all only seem to have `*.sbs` file format.

pcuenca (commented):
LGTM

SBrandeis (approved):

julien-c (approved):
nice – should they have icons in the internal codebase?

Wauplin (commented):
LGTM!

About icons, couldn't we make the process public -same as the library definitions-? Suggesting it since we have quite a lot of those, especially now that we are adding local apps as well.

Wauplin (approved):
LGTM!

About icons, couldn't we make the process public -same as the library definitions-? Suggesting it since we have quite a lot of those, especially now that we are adding local apps as well.

PR #749, opened 2024-06-10
Add new fields for computing downloads

Kakulukian:
I totally forgot to update the package with the new fields we have to reduce some load:
`path_extension` | `path_prefix` | `path_filename`.

Simplified to use query language (boolean language) which is more straightforward to understand than DSL

### DIFF
--- a/packages/tasks/src/model-libraries-downloads.ts
+++ b/packages/tasks/src/model-libraries-downloads.ts
@@ -6,15 +6,13 @@
  * Read this doc about download stats on the Hub:
  *
  * https://huggingface.co/docs/hub/models-download-stats
- *
+ * Available fields:
+ *  - path: the complete path of the model
+ *  - path_prefix: the prefix of the path of the model
+ *  - path_extension: the extension of the path of the model
+ *  - path_filename: the extension of the path of the model
  * see also:
- * https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-bool-query.html
+ * https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-query-string-query.html
  */
 
-export type ElasticBoolQueryFilter =
-	// match a single filename
-	| { term?: { path: string } }
-	// match multiple possible filenames
-	| { terms?: { path: string[] } }
-	// match a wildcard
-	| { wildcard?: { path: string } };
+export type ElasticSearchQuery = string;
--- a/packages/tasks/src/model-libraries-downloads.ts
+++ b/packages/tasks/src/model-libraries-downloads.ts
@@ -7,10 +7,10 @@
  *
  * https://huggingface.co/docs/hub/models-download-stats
  * Available fields:
- *  - path: the complete path of the model
- *  - path_prefix: the prefix of the path of the model
- *  - path_extension: the extension of the path of the model
- *  - path_filename: the extension of the path of the model
+ *  - path: the complete file path (relative) (e.g: "prefix/file.extension")
+ *  - path_prefix: the prefix of the file path (e.g: "prefix", empty if no prefix)
+ *  - path_extension: the extension of the file path (e.g: "extension")
+ *  - path_filename: the name of the file path (e.g: "file")
  * see also:
  * https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-query-string-query.html
  */
--- a/packages/tasks/src/model-libraries-downloads.ts
+++ b/packages/tasks/src/model-libraries-downloads.ts
@@ -8,7 +8,7 @@
  * https://huggingface.co/docs/hub/models-download-stats
  * Available fields:
  *  - path: the complete file path (relative) (e.g: "prefix/file.extension")
- *  - path_prefix: the prefix of the file path (e.g: "prefix", empty if no prefix)
+ *  - path_prefix: the prefix of the file path (e.g: "prefix/", empty if no prefix)
  *  - path_extension: the extension of the file path (e.g: "extension")
  *  - path_filename: the name of the file path (e.g: "file")
  * see also:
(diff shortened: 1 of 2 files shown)


packages/tasks/src/model-libraries.ts
```diff
@@ -34,7 +34,7 @@ export interface LibraryUiElement {
 	 * By default, those files are counted:
 	 * "config.json", "config.yaml", "hyperparams.yaml", "meta.yaml"
 	 */
-	countDownloads?: ElasticBoolQueryFilter;
+	countDownloads?: ElasticBoolQueryFilter[];
```
julien-c, line 37: what is the behavior of the array? is it an `AND`?

Kakulukian: AND

SBrandeis (approved):
nice !!!

packages/tasks/src/model-libraries-downloads.ts
```diff
@@ -6,15 +6,13 @@
...
  */
 
-export type ElasticBoolQueryFilter =
-	// match a single filename
-	| { term?: { path: string } }
-	// match multiple possible filenames
-	| { terms?: { path: string[] } }
-	// match a wildcard
-	| { wildcard?: { path: string } };
+export type ElasticSearchQuery = string;
```
julien-c, line 18: i haven't read the linked internal PR yet but is the move to a string (vs. an object) a good idea?

julien-c: or is it a query language natively supported by Elastic?

packages/tasks/src/model-libraries.ts
```diff
@@ -140,18 +134,14 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		docsUrl: "https://aphp.github.io/edsnlp/latest/",
 		filter: false,
 		snippets: snippets.edsnlp,
-		countDownloads: {
-			wildcard: { path: "*/config.cfg" },
-		},
+		countDownloads: `path_filename:"config.cfg" AND _exists_:"path_prefix"`,
```
julien-c, line 137:
```suggestion
		countDownloads: `path_filename:"config.cfg"`,
```

i would guess we probably don't care about this part i.e. if the file is at the root, we can count it

packages/tasks/src/model-libraries.ts
```diff
@@ -140,18 +134,14 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
...
 	},
 	elm: {
 		prettyLabel: "ELM",
 		repoName: "elm",
 		repoUrl: "https://github.com/slicex-ai/elm",
 		filter: false,
-		countDownloads: {
-			wildcard: { path: "*/slicex_elm_config.json" },
-		},
+		countDownloads: `path_filename:"slicex_elm_config.json" AND _exists_:"path_prefix"`,
```
julien-c, line 144: same here

packages/tasks/src/model-libraries.ts
```diff
@@ -497,9 +459,7 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		prettyLabel: "WhisperKit",
 		repoName: "WhisperKit",
 		repoUrl: "https://github.com/argmaxinc/WhisperKit",
-		countDownloads: {
-			wildcard: { path: "*/model.mil" },
-		},
+		countDownloads: `path_filename:"model.mil" AND _exists_:"path_prefix"`,
```
julien-c, line 462: same

packages/tasks/src/model-libraries.ts
Kakulukian: I prefer have at least one example and how we can achieve this

julien-c: sure, ok to keep one

packages/tasks/src/model-libraries-downloads.ts
Kakulukian: | is it a query language natively supported by Elastic?

Though the good operation yes (`query_string`)

I think it'll simplify how to contribute because DSL is not common vs a basic boolean language

packages/tasks/src/model-libraries.ts
Kakulukian, line 144:
```suggestion
		countDownloads: `path_filename:"slicex_elm_config.json"`,
```

packages/tasks/src/model-libraries.ts
Kakulukian:
```suggestion
		countDownloads: `path_filename:"model" AND path_extension:"mil" AND _exists_:"path_prefix"`,
```

julien-c (commented):
will let other reviewers review, but this is fine for me 👍

Pierrci (commented):
LGTM

Kakulukian:
Can another code owner review it? The syntax for counting downloads will evolve

packages/tasks/src/model-libraries.ts
```diff
@@ -523,9 +485,7 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		prettyLabel: "WhisperKit",
 		repoName: "WhisperKit",
 		repoUrl: "https://github.com/argmaxinc/WhisperKit",
-		countDownloads: {
-			wildcard: { path: "*/model.mil" },
-		},
+		countDownloads: `path_filename:"model" AND path_extension:"mil" AND _exists_:"path_prefix"`,
```
pcuenca, line 488: Interesting example with `_exists_`

packages/tasks/src/model-libraries.ts
```diff
@@ -100,7 +96,7 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		repoName: "audioseal",
 		repoUrl: "https://github.com/facebookresearch/audioseal",
 		filter: false,
-		countDownloads: { wildcard: { path: "*.pth" } },
+		countDownloads: "path_extension:pt",
```
pcuenca, line 99: I think it's `pth`, can you confirm @Vaibhavs10?

Kakulukian: ah yes forgot one character

PR #751, opened 2024-06-12
Add Stable Diffusion 3 as library

NielsRogge:
This PR ensures downloads stats will work for https://huggingface.co/stabilityai/stable-diffusion-3-medium.

### DIFF
--- a/packages/tasks/src/model-libraries.ts
+++ b/packages/tasks/src/model-libraries.ts
@@ -394,7 +394,7 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		countDownloads: { term: { path: "model.safetensors" } },
 		snippets: snippets.stable_audio_tools,
 	},
-	"stable-diffusion-3": {
+	"diffusion-single-file": {
 		prettyLabel: "Stable Diffusion 3",
 		repoName: "stable-diffusion-3",
 		repoUrl: "https://github.com/comfyanonymous/ComfyUI",
--- a/packages/tasks/src/model-libraries.ts
+++ b/packages/tasks/src/model-libraries.ts
@@ -394,6 +394,15 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		countDownloads: { term: { path: "model.safetensors" } },
 		snippets: snippets.stable_audio_tools,
 	},
+	"stable-diffusion-3": {
+		prettyLabel: "Stable Diffusion 3",
+		repoName: "stable-diffusion-3",
+		repoUrl: "https://github.com/comfyanonymous/ComfyUI",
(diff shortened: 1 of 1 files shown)


packages/tasks/src/model-libraries.ts
```diff
@@ -394,6 +394,15 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		countDownloads: { term: { path: "model.safetensors" } },
 		snippets: snippets.stable_audio_tools,
 	},
+	"stable-diffusion-3": {
```
osanseviero, line 397: There's some ongoing discussion on Slack of whether to do this or not. If we go ahead, we'll need a different name, because otherwise all repos with a tag `stable-diffusion-3` will get this library automatically.

apolinario:
Currently Stability AI changed the repo to `diffusers` to not lose download stats. But I agree maybe a specific library may make more sense. Naming-wise, it's a bit tricky, how about: `diffusion-single-file` ?

It is the name we use internally for diffusion models that are not on diffusers format (comfyUI, AUTOMATIC1111), also the name [from_single_file](https://huggingface.co/docs/diffusers/v0.29.0/en/api/loaders/single_file#diffusers.loaders.FromSingleFileMixin.from_single_file) in the diffusers library

packages/tasks/src/model-libraries.ts
apolinario, line 397:
```suggestion
	"diffusion-single-file": {
```

osanseviero: LGTM

packages/tasks/src/model-libraries.ts
```diff
@@ -394,6 +394,15 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
...
+		prettyLabel: "Stable Diffusion 3",
```
apolinario, line 398:
```suggestion
		prettyLabel: "Diffusion Single File",
```

packages/tasks/src/model-libraries.ts
```diff
@@ -394,6 +394,15 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
...
+		repoName: "stable-diffusion-3",
```
apolinario, line 399:
```suggestion
		repoName: "diffusion-single-file",
```

packages/tasks/src/model-libraries.ts
julien-c: or Comfy?

julien-c (approved):

osanseviero (approved):
Discussed quickly offline, let's proceed with it 🔥

PR #752, opened 2024-06-13
core: ✨ add new package space-header

enzostvs:
Add the `mini header` into your project if running on a custom/sub domain

Screenshot 2024-06-13 at 11 24 02 AM

Example:

```js
import { init } from "@huggingface/space-header"

// ...

init("enzostvs/lora-studio")
```

Must be called once the page has been mounted. If this is not the case, a log error is displayed to prevent the user

we've decided to stick with vanilla JS, which may evolve if necessary. (web component, etc)


packages/space-header/tsconfig.json
julien-c, line 1: make this file shorter?

packages/space-header/src/inject_to_body.ts
```diff
@@ -0,0 +1,11 @@
+export const inject_to_body = (
```
jbilcke, line 1: perhaps we should add a class name to make it easier for integrators to change the position?

that way they don't have to change their underlying app/page design in case the header gets in the way of something (toolbar etc)

packages/space-header/src/index.ts
```diff
@@ -0,0 +1,23 @@
+import { inject_to_body} from "./inject_to_body"
```
jbilcke, line 1: maybe we could add a prettifier to have unified spaces / semicolons, but that's just code-smetics

packages/space-header/tsconfig.json
enzostvs: done!

packages/space-header/src/index.ts
enzostvs: The code was on another project before (without prettier), so should be good now!

packages/space-header/src/inject_to_body.ts
jbilcke: Ahh got it! 

I missed the `box.setAttribute("id", "huggingface-space-header");`

all good 👍

jbilcke:
Thank you @enzostvs  for doing this!

for some context I had started my [own independent version based on Tailwind](https://huggingface.co/spaces/jbilcke-hf/ai-stories-factory/blob/main/src/components/chrome/index.tsx) (my goal was to stay close to the original using Tailwind ie. save time in case the original design changes, let tailwind generate a tiny optimized stylesheet, let the framework generate a tiny static .js code injector etc)

but I never finished as I got involved in other things, and I think you approach works fine too (this header shouldn't change that often)..

plus no dependencies which is great! 👌

packages/space-header/src/inject_to_body.ts
coyotte508: Related: I think we should allow users to provide a "target" and if so, rather than appending to `body`, append to that target.

Like with Svelte: (but same for any other framework)

https://svelte.dev/docs/client-side-component-api#creating-a-component

eg `init(space, {target: "..."})`

pngwn:
Thanks for this PR, looks great, just have a few questions from a gradio PoV, where we may not know the user and space name up front. 

- we already make this request, and would need to make an extra one to get the space name anyway (you only know the space name when using this manually, we don’t have it on a subdomain), would it be possible to skip the API request and just pass in the Space object instead of a name?
- the origin check isn’t sufficient for gradio as gradio apps can be hosted anywhere and all non Hugging Face urls will ‘pass’ the check causing the widget to render (colab/ locally/ etc). This isn’t anything that requires changes here but just mentioning it, we will handle this on the gradio side.

packages/space-header/src/inject_to_body.ts
enzostvs: @coyotte508 should be good now! tell me if you had something different in mind 🤗

enzostvs:
> Thanks for this PR, looks great, just have a few questions from a gradio PoV, where we may not know the user and space name up front.
> 
> * we already make this request, and would need to make an extra one to get the space name anyway (you only know the space name when using this manually, we don’t have it on a subdomain), would it be possible to skip the API request and just pass in the Space object instead of a name?

@pngwn just pushed a change to cover your first point, now you can pass the `namespace` or directly the space information, here is the recommended informations 
```
export interface Space {
	id: string;
	likes: number;
	author: string;
}
```

pngwn:
Amazing. Thanks!

pngwn (approved):
Looks great from the gradio side! Thanks @enzostvs!

julien-c:
> * the origin check isn’t sufficient for gradio as gradio apps can be hosted anywhere and all non Hugging Face urls will ‘pass’ the check causing the widget to render (colab/ locally/ etc). This

Yes, you will use the [Spaces env variables](https://huggingface.co/docs/hub/en/spaces-overview#helper-environment-variables) to check whether the Gradio app is running on Spaces, correct, @pngwn?

pngwn:
@julien-c yus! We will also hide when embedded via the web component I think as there space info is already in the wrapper. 

One case we may not be able to figure out is if someone embeds using an iframe. I'm not sure what we want to do in that case, should we just show it?

coyotte508 (approved):

packages/space-header/src/inject_to_body.ts
coyotte508: In my mind it was more of a string that can be passed to `document.querySelector`, but a `HTMLElement` works too

jbilcke (approved):

julien-c:
> One case we may not be able to figure out is if someone embeds using an iframe. I'm not sure what we want to do in that case, should we just show it?

you mean the gradio app runs on Spaces, but the raw URL like https://julien-c-coqui.hf.space is embedded in an iframe on a different website?

I would display it in that case yes (but no strong opinion)

PR #758, opened 2024-06-17
Update model-libraries.ts to count download stats for nemo models

HeyyyyyyG:


osanseviero:
Discussing internally because this may not be a valid query

packages/tasks/src/model-libraries.ts
```diff
@@ -280,7 +280,7 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		repoUrl: "https://github.com/NVIDIA/NeMo",
 		snippets: snippets.nemo,
 		filter: true,
-		countDownloads: { wildcard: { path: "*.nemo" } },
+		countDownloads: { wildcard: { path: "*.nemo", "model_config.yaml" } },
```
julien-c, line 283: i think this is not valid JS or JSON

odelalleau, line 283: Note that counting all downloads of `model_config.yaml" would be over-estimating downloads, since I expect a number of people will check the config without actually downloading the model.

Maybe `common.pt` is a better file to use for this purpose?

HeyyyyyyG: I follow the same practice as `paddlenlp` and `peft`, so I assume tracking model_config is ok?

HeyyyyyyG: Sorry can you elaborate? I see `speechbrain` is [tracking](https://github.com/huggingface/huggingface.js/blob/426228b630f90674553dd2b7738e01051001e8e3/packages/tasks/src/model-libraries.ts#L389) `hyperparams.yaml`

osanseviero: Yes, I think this is a non-issue. Looking at a configuration file will not trigger a download count unless downloaded.

osanseviero:
Closing in favor of https://github.com/huggingface/huggingface.js/pull/749

PR #759, opened 2024-06-18
Update model-libraries.ts and add count statistics for glyph-byt5 (https://arxiv.org/abs/2403.09622)

lzy-tony:


packages/tasks/src/model-libraries.ts
```diff
@@ -214,6 +214,15 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 			term: { path: "gliner_config.json" },
 		},
 	},
+	glyph-byt5: {
```
osanseviero, line 217:
```suggestion
	"glyph-byt5": {
```

packages/tasks/src/model-libraries.ts
```diff
@@ -214,6 +214,15 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
...
+		prettyLabel: "Glyph-ByT5",
+		repoName: "Glyph-ByT5",
+		repoUrl: "https://github.com/AIGText/Glyph-ByT5",
+		filter: false,
+		countDownloads: {
+			term: { path: "checkpoints/byt5_model.pt" },
```
osanseviero, line 223: I think https://huggingface.co/GlyphByT5/Glyph-SDXL-v2/tree/main follows a different structure unfortunately

packages/tasks/src/model-libraries.ts
osanseviero, line 217: Some repos will need to have `library_name: glyph-byt5` before we merge this

osanseviero (commented):
Very nice PR!

packages/tasks/src/model-libraries.ts
lzy-tony: Hi, thanks for the review, I have added library name in https://huggingface.co/GlyphByT5/Glyph-SDXL-v2/tree/main

packages/tasks/src/model-libraries.ts
```diff
@@ -214,6 +214,15 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 			term: { path: "gliner_config.json" },
 		},
 	},
+	"glyph-byt5": {
+		prettyLabel: "Glyph-ByT5",
+		repoName: "Glyph-ByT5",
+		repoUrl: "https://github.com/AIGText/Glyph-ByT5",
+		filter: false,
+		countDownloads: {
+			term: { path: "checkpoints/byt5_model.pt" },
```
osanseviero, line 223: Overall LGTM. Just note that we strongly suggest using conventions that will make it easier to scale up. For example, if you change the checkpoint name, this query won't work. To handle this, we usually use a file such as `config.json` or another query file that will always be there for all repos under a library. Would you consider doing something like this for the glyph repos?

lzy-tony: Thanks for the advice! We would consider such a format in the future, but at the moment we would prefer keeping the current format.

osanseviero: SG!

osanseviero (approved):
LGTM!

lzy-tony:
Hi, do we still need more reviews for this pr to merge?

osanseviero:
LGTM! Let's go :rocket:

lzy-tony:
It seems like the matching rules have been changed before merging (https://github.com/huggingface/huggingface.js/pull/749), so unfortunately this pr might need to be updates :(

PR #760, opened 2024-06-18
add hunyuan Dit library

xianbaoqian:


osanseviero:
I don't see any repos setting the library with `hunyuan-dit` yet (https://huggingface.co/models?library=hunyuan-dit) so this won't work without models there

julien-c:
@osanseviero @SBrandeis @Wauplin maybe we should add a `exampleHfRepo: URL`  property to `LibraryUiElement ` to always link to example Hub model

xianbaoqian:
> I don't see any repos setting the library with hunyuan-dit yet (https://huggingface.co/models?library=hunyuan-dit) so this won't work without models there

There are a few PRs that need to be merged.

packages/tasks/src/model-libraries.ts
```diff
@@ -223,6 +223,14 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 			terms: { path: ["ckpt/tensor00000_000", "ckpt-0/tensor00000_000"] },
 		},
 	},
+	"hunyuan-dit": {
+		prettyLabel: "HunyuanDiT",
+		repoName: "HunyuanDiT",
+		repoUrl: "https://github.com/Tencent/HunyuanDiT",
+		countDownloads: {
+			terms: { path: ["pytorch_model_ema.pt", "pytorch_model_distill.pt"] },
```
osanseviero, line 231: Is there any file that is consistently across repos? E.g. vocab.txt or a config.json?

xianbaoqian: Not really. They chose not to add a metadata file, only using weight file for download metrics. They suggested to use the two .pt files are the two main weight file, one for normal weight and another one for distilled weight. The two files are mutually exclusive and won't be seen in the same repo.

xianbaoqian:
https://huggingface.co/models?library=hunyuan-dit&sort=trending

PRs of changing the library has been merged now. PTAL

osanseviero (approved):
LGTM

PR #763, opened 2024-06-18
[`integration`] Add `bm25s` library for bm25 retrieval models/indices

tomaarsen:
Hello!

## Pull Request overview
* Add BM25S as a library on the Hub. 

## Links
* Repository: https://github.com/xhluca/bm25s
* Models: https://huggingface.co/models?other=bm25s

## Details
BM25S is a very new library for efficient BM25: an important algorithm for full-text search. It can be combined with vector search (e.g. Sentence Transformers' domain) for hybrid search, which is commonly used and very powerful. I'm considering mentioning this library more in Sentence Transformers and its example/docs, so it'll be nice to get the "Use in Library" button & download count for these models/indices. I've set `filter` to False as it's not a big library currently.

cc @xhluca for context: this should add a "Use this model" button in the top right of all models on the Hub that have a `bm25s` tag in the model card metadata. It'll then show the following snippet:
```python
from bm25s.hf import BM25HF

retriever = BM25HF.load_from_hub("${model.id}")
```

- Tom Aarsen


xhluca:
Lgtm! Thanks for the pr.

packages/tasks/src/model-libraries.ts
```diff
@@ -111,6 +111,16 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 			wildcard: { path: "*.npz" },
 		},
 	},
+	bm25s: {
```
osanseviero, line 114: LGTM, but let's try to switch to explicitly setting this as a library (`library_name: bm25s`) in the repos rather than using `tags`

osanseviero (approved):
LGTM, very cool 🔥

packages/tasks/src/model-libraries.ts
tomaarsen: Agreed, I'll assist with getting that done

julien-c (commented):
cool addition

packages/tasks/src/model-libraries.ts
xhluca: done!

PR #766, opened 2024-06-19
✅ Add support for audioseal as a library.

Vaibhavs10:
Library: https://github.com/facebookresearch/audioseal
HF Hub weights: https://huggingface.co/facebook/audioseal

### DIFF
--- a/packages/tasks/src/model-libraries-snippets.ts
+++ b/packages/tasks/src/model-libraries-snippets.ts
@@ -48,21 +48,21 @@ model = BaseModel.from_pretrained("${model.id}")`,
 
 export const audioseal = (model: ModelData): string[] => {
 	const watermarkSnippet = `# Watermark Generator
-	from audioseal import AudioSeal
+from audioseal import AudioSeal
+
+model = AudioSeal.load_generator("${model.id}")
+# pass a tensor (tensor_wav) of shape (batch, channels, samples) and a sample rate
+wav, sr = tensor_wav, 16000
 	
-	model = AudioSeal.load_generator("${model.id}")
-	# pass a tensor (tensor_wav) of shape (batch, channels, samples) and a sample rate
-	wav, sr = tensor_wav, 16000
-	
-	watermark = model.get_watermark(wav, sr)
-	watermarked_audio = wav + watermark`;
+watermark = model.get_watermark(wav, sr)
+watermarked_audio = wav + watermark`;
 
 	const detectorSnippet = `# Watermark Detector
-	from audioseal import AudioSeal
+from audioseal import AudioSeal
 
-	detector = AudioSeal.load_detector("${model.id}")
+detector = AudioSeal.load_detector("${model.id}")
 	
-	result, message = detector.detect_watermark(watermarked_audio, sr)`;
+result, message = detector.detect_watermark(watermarked_audio, sr)`;
 	return [watermarkSnippet, detectorSnippet];
 };
 
--- a/packages/tasks/src/model-libraries-snippets.ts
+++ b/packages/tasks/src/model-libraries-snippets.ts
@@ -46,24 +46,25 @@ export const asteroid = (model: ModelData): string[] => [
 model = BaseModel.from_pretrained("${model.id}")`,
 ];
 
-export const audioseal = (model: ModelData): string[] => [
-	`# Watermark Generator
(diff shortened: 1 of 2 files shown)


Vaibhavs10:
adding the code snippet

packages/tasks/src/model-libraries-snippets.ts
```diff
@@ -46,6 +46,18 @@ export const asteroid = (model: ModelData): string[] => [
 model = BaseModel.from_pretrained("${model.id}")`,
 ];
 
+export const audioseal = (model: ModelData): string[] => [
+	`from audioseal import AudioSeal
+
+model = AudioSeal.load_generator("${model.id}")
+
+wav, sr = tensor_wav, 16000
```
osanseviero, line 54: Could we update to show what `tensor_wav` is please?

packages/tasks/src/model-libraries.ts
```diff
@@ -95,6 +95,14 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		snippets: snippets.audiocraft,
 		filter: false,
 	},
+	audioseal: {
+		prettyLabel: "Audioseal",
+		repoName: "audioseal",
+		repoUrl: "https://github.com/facebookresearch/audioseal",
+		filter: false,
+		countDownloads: { wildcard: { path: "*.pth" } },
```
osanseviero, line 103: won't this double count downloads?

osanseviero (commented):
Nice! 🔥

packages/tasks/src/model-libraries.ts
Vaibhavs10: Not quite, only one model is used at a time, either the generator or the detector. They both also have different code snippets, maybe I should add both to the snippets. Let moi add.

Vaibhavs10: Fixed in the latest commit. I think it should be fine since both checkpoints are not used in one go.

packages/tasks/src/model-libraries-snippets.ts
Vaibhavs10: Fixed in the latest commit.

packages/tasks/src/model-libraries-snippets.ts
```diff
@@ -46,6 +46,25 @@ export const asteroid = (model: ModelData): string[] => [
...
+from audioseal import AudioSeal
+
+model = AudioSeal.load_generator("${model.id}")
+# pass a tensor (tensor_wav) of shape (batch, channels, samples) and a sample rate
+wav, sr = tensor_wav, 16000
+
+watermark = model.get_watermark(wav, sr)
+watermarked_audio = wav + watermark
+
+# Watermark Detector
```
osanseviero, line 60: The function returns a list of strings, so you can actually return multiple snippets rather than hacking them in a single one (see `transformers` for example)

Vaibhavs10: Let me know if this doesn't work! 🤗

packages/tasks/src/model-libraries-snippets.ts
```diff
@@ -46,6 +46,26 @@ export const asteroid = (model: ModelData): string[] => [
 model = BaseModel.from_pretrained("${model.id}")`,
 ];
 
+export const audioseal = (model: ModelData): string[] => {
+	const watermarkSnippet = `# Watermark Generator
+	from audioseal import AudioSeal
```
osanseviero, line 51: I think you need to remove the initial space
```suggestion
from audioseal import AudioSeal
```

packages/tasks/src/model-libraries.ts
```diff
@@ -95,6 +95,14 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		snippets: snippets.audiocraft,
 		filter: false,
 	},
+	audioseal: {
+		prettyLabel: "Audioseal",
```
pcuenca, line 99:
```suggestion
		prettyLabel: "AudioSeal",
```

I think that's what they use in their repo.

pcuenca (commented):
Nice! 🔥

osanseviero (approved):
LGTM with Pedro's suggestion

Vaibhavs10:
Thanks! Fixed! Merging! 🚀

PR #767, opened 2024-06-19
👾Fix chunk count not incrementing when parsing additional chunks in a local file

ryan-the-crayon:
When parsing a local GGUF and the metadata section does not fit in the same chunk, the parser would try to load more chunks. However, in the local file implementation range view, the chunk count is not incremented, resulting it not actually loading in more data.


mishig25:
cc: @ngxson

ngxson:
LGTM. Thanks! I don't understand why I missed this line of code :eyes:

mishig25:
@ryan-the-crayon thanks a lot for the contribution! Anyway, you can add a test case ? https://github.com/huggingface/huggingface.js/blob/e2689f8c7b02d8eb05e3b7869f8e02f7e495ea93/packages/gguf/src/gguf.spec.ts#L230-L241

PR #770, opened 2024-06-24
fix: fix outdated count statistics for glyph-byt5

lzy-tony:
the previous version is outdated for glyph-byt5, update format to match new matching rules


pcuenca (approved):
Looks good to me, deferring to @Kakulukian for confirmation :)

Kakulukian (approved):

PR #773, opened 2024-06-25
✨ Count downloads for audiocraft

nateraw:
Hey there, wanted to add download counts for `audiocraft` to see how many folks are downloading [songstarter-v0.2](https://huggingface.co/nateraw/musicgen-songstarter-v0.2). :)

My model doesn't have transformers ckpts in it, so the counts are missed. Existing musicgen models from FB are only being counted if the user is using transformers too I think.

So I updated the logic here to count on `state_dict.bin`, which is `audiocraft` checkpoint, or `config.json`, which captures downloads from `transformers`.

### DIFF
--- a/packages/tasks/src/model-libraries.ts
+++ b/packages/tasks/src/model-libraries.ts
@@ -90,6 +90,7 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		repoUrl: "https://github.com/facebookresearch/audiocraft",
 		snippets: snippets.audiocraft,
 		filter: false,
+		countDownloads: `path:"state_dict.bin" OR path:"config.json"`,
 	},
 	audioseal: {
 		prettyLabel: "AudioSeal",
--- a/packages/tasks/src/model-libraries.ts
+++ b/packages/tasks/src/model-libraries.ts
@@ -90,7 +90,7 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		repoUrl: "https://github.com/facebookresearch/audiocraft",
 		snippets: snippets.audiocraft,
 		filter: false,
-		countDownloads: `path:"state_dict.bin" OR path:"config.json"`,
+		countDownloads: `path:"state_dict.bin"`,
 	},
 	audioseal: {
 		prettyLabel: "AudioSeal",


julien-c (approved):

Wauplin (approved):
Nice!

packages/tasks/src/model-libraries.ts
```diff
@@ -90,6 +90,7 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		repoUrl: "https://github.com/facebookresearch/audiocraft",
 		snippets: snippets.audiocraft,
 		filter: false,
+		countDownloads: `path:"state_dict.bin" OR path:"config.json"`,
```
osanseviero, line 93: Nice, this will also enable download stats for repos such as https://huggingface.co/facebook/magnet-small-10secs 

cc @Vaibhavs10

osanseviero, line 93: At the moment, the current download setup will just take into account the primary library for download stats. That means that if a repo is both Audiocraft and transformers, and transformers is the primary library, audiocraft checkpoints will not be taken into account. (there's an issue for this)

Due to this, I would make this filter just about audiocraft (so just keep `path:"state_dict.bin"`). WDYT?

osanseviero (commented):
Very nice!

packages/tasks/src/model-libraries.ts
Vaibhavs10: Also in favor of just keeping `state_dict.bin`

Vaibhavs10 (commented):
Thanks for the PR Nate. Quick note: `config.json` is not used in `audiocraft` it directly just downloads the `state_dict.bin`

As it is the only relevant one for AudioCraft - MusicGen, AudioGen and MAGNeT models we have seperate ones for AudioSeal and Encodec (Transformers checkpoint so doesn't quite matter)

ref: https://github.com/facebookresearch/audiocraft/blob/adf0b04a4452f171970028fcf80f101dd5e26e19/audiocraft/models/loaders.py#L40 (look for `_get_state_dict` function calls)

I still don't know why I did not add download counts in my earlier PR in Feb: https://github.com/huggingface/huggingface.js/pull/490

packages/tasks/src/model-libraries.ts
nateraw, line 93:
```suggestion
		countDownloads: `path:"state_dict.bin"`,
```
Cool, we'll just do this then?

Vaibhavs10 (approved):
LGTM! ❤️ - We can go ahead and merge this bad boi!

PR #777, opened 2024-06-27
gguf: test local file with big metadata

ngxson:
Ref comment: https://github.com/huggingface/huggingface.js/pull/767#issuecomment-2181173686

This file have 32MB of just metadata (no tensor), can be useful for testing or benchmarking.

Due to its large size, running the download process inside the test case may cause timeout (so I moved it to `beforeAll`)

### DIFF
--- a/packages/gguf/src/gguf.spec.ts
+++ b/packages/gguf/src/gguf.spec.ts
@@ -1,5 +1,5 @@
-import { describe, expect, it } from "vitest";
-import { GGMLQuantizationType, gguf, ggufAllShards, parseGgufShardFilename } from "./gguf";
+import { beforeAll, describe, expect, it } from "vitest";
+import { GGMLQuantizationType, GGUFParseOutput, gguf, ggufAllShards, parseGgufShardFilename } from "./gguf";
 import fs from "node:fs";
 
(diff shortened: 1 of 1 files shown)


julien-c (approved):
lgtm

packages/gguf/src/gguf.spec.ts
```diff
@@ -12,8 +13,19 @@ const URL_V1 =
 	"https://huggingface.co/tmadge/testing/resolve/66c078028d1ff92d7a9264a1590bc61ba6437933/tinyllamas-stories-260k-f32.gguf";
 const URL_SHARDED_GROK =
 	"https://huggingface.co/Arki05/Grok-1-GGUF/resolve/ecafa8d8eca9b8cd75d11a0d08d3a6199dc5a068/grok-1-IQ3_XS-split-00001-of-00009.gguf";
+const URL_BIG_METADATA = "https://huggingface.co/ngxson/test_gguf_models/resolve/main/gguf_test_big_metadata.gguf";
```
mishig25, line 16:
```suggestion
const URL_BIG_METADATA = "https://huggingface.co/ngxson/test_gguf_models/resolve/c72c0f045c36b7c01f348c5fbba424cd6a52f770/gguf_test_big_metadata.gguf";
```

to make the test case more robust, rather than `branch/file`, we can use `commit/file`

mishig25 (approved):
🔥

PR #787, opened 2024-07-08
Add Exllamav2 to local-apps.ts

KT313:


packages/tasks/src/local-apps.ts
```diff
@@ -91,6 +91,13 @@ export const LOCAL_APPS = {
 		displayOnModelPage: isGgufModel,
 		snippet: snippetLlamacpp,
 	},
+	"exllamav2": {
+		prettyLabel: "exllamav2",
+		docsUrl: "https://github.com/turboderp/exllamav2",
+		mainTask: "text-generation",
+		displayOnModelPage: isGgufModel,
+		snippet: snippetLlamacpp,
```
Vaibhavs10, line 99: Does EXL2 support running directly with llama.cpp? this should be a different snippet no?

PR #788, opened 2024-07-08
chore: add mlc-llm as a library.

Vaibhavs10:


packages/tasks/src/model-libraries.ts
```diff
@@ -304,6 +304,13 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		filter: false,
 		countDownloads: `path:"model.safetensors"`,
 	},
+	"mlc-llm": {
+		prettyLabel: "MLC-LLM",
+		repoName: "MLC-LLM",
+		repoUrl: "https://github.com/mlc-ai/mlc-llm",
+		filter: true,
```
Vaibhavs10, line 311: Setting to `True` because this would effect at least 150 model checkpoints: https://huggingface.co/mlc-ai

(we'd need to make sure that all of them have `mlc-llm` tag)

Vaibhavs10 (commented):
Discussing with the MLC team if we should add a snippet here or not, since the model checkpoints would be compatible with both `web-llm` as well as `mlc-llm`

packages/tasks/src/model-libraries.ts
```diff
@@ -304,6 +304,13 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		filter: false,
 		countDownloads: `path:"model.safetensors"`,
 	},
+	"mlc-llm": {
+		prettyLabel: "MLC-LLM",
+		repoName: "MLC-LLM",
+		repoUrl: "https://github.com/mlc-ai/mlc-llm",
```
Wauplin, line 310:
```suggestion
		repoUrl: "https://github.com/mlc-ai/mlc-llm",
		docsUrl: "https://llm.mlc.ai/docs/",
```

Seems they have docs as well (https://llm.mlc.ai/docs/)

packages/tasks/src/model-libraries.ts
Wauplin, line 311:
```suggestion
		filter: false,
```

`filter: true` would add mlc-llm library to the libraries list on the left side on hf.co/models. Given it's quite small for now ([3 models](https://huggingface.co/models?other=mlc-llm)?), let's have it hidden.
image

packages/tasks/src/model-libraries.ts
```diff
@@ -304,6 +304,13 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
...
+		filter: true,
+		countDownloads: `path_extension:"json"`,
```
Wauplin, line 312: Is [mlc-chat-config.json](https://huggingface.co/mlc-ai/Llama-3-8B-Instruct-q0f16-MLC/blob/main/mlc-chat-config.json) supposed to be the only target file? If yes, let's be explicit 

```suggestion
		countDownloads: `path:"mlc-chat-config.json"`,
```

(or tokenizer.json / tokenizer_config.json ?). Better to count once when the whole repo is downloaded.

packages/tasks/src/model-libraries.ts
Wauplin: Sorry, I didn't see https://github.com/huggingface/huggingface.js/pull/788#discussion_r1668737141 before reviewing

Vaibhavs10: ref: https://github.com/huggingface/huggingface.js/pull/788#discussion_r1668737141

packages/tasks/src/model-libraries.ts
Vaibhavs10: That's a good point, I'll confirm which file should we use here.

packages/tasks/src/model-libraries.ts
Vaibhavs10: Let's do `filter: false` to begin with and then update this as the changes trickle down.

packages/tasks/src/model-libraries.ts
```diff
@@ -304,6 +304,14 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		filter: false,
 		countDownloads: `path:"model.safetensors"`,
 	},
+	"mlc-llm": {
+		prettyLabel: "MLC-LLM",
+		repoName: "MLC-LLM",
+		repoUrl: "https://github.com/mlc-ai/mlc-llm",
+		docsUrl: "https://llm.mlc.ai/docs/",
+		filter: true,
```
Wauplin, line 312:
```suggestion
		filter: false,
```

(as per https://github.com/huggingface/huggingface.js/pull/788/files#r1670222328)

Wauplin (approved):

PR #793, opened 2024-07-10
Rapatriate Inference types to moon

julien-c:


packages/tasks/src/model-data.ts
```diff
@@ -27,8 +12,9 @@ export interface ModelData {
 	id: string;
 	/**
 	 * Whether or not to enable inference widget for this model
+	 * TODO(type it)
 	 */
-	inference: InferenceDisplayability;
+	inference: string;
```
coyotte508, line 17:
```suggestion
	inference: string | "Yes" | ...;
```

To keep some documentation while being lax?

julien-c: but see https://github.com/huggingface-internal/moon-landing/pull/10618 (internal) it's gonna change quite a bit

coyotte508 (approved):

julien-c:
merging, even though i'm surprised that it doesn't break the (archived) widget build, given it uses this type

(we can just move the type into the archived widgets if it's simpler)

coyotte508:
> merging, even though i'm surprised that it doesn't break the (archived) widget build, given it uses this type
> 
> (we can just move the type into the archived widgets if it's simpler)

widgets was removed from `pnpm-workspace.yml`, so it's not being built or anything in the CI

PR #795, opened 2024-07-11
[hub.js] List models by inference status

mishig25:


SBrandeis:
`inferenceStatus` is not a stable API yet (name and values might change)
Let's not rely on it yet

PR #797, opened 2024-07-11
add dataset libraries: distilabel, fiftyone, argilla

severo:


packages/tasks/src/dataset-libraries.ts
```diff
@@ -53,6 +53,18 @@ export const DATASET_LIBRARIES_UI_ELEMENTS = {
...
+		repoName: "distilabel",
+		repoUrl: "https://github.com/argilla-io/distilabel",
+		docsUrl: "https://distilabel.argilla.io",
+	},
+	fiftyone: {
+		prettyLabel: "FiftyOne",
+		repoName: "fiftyone",
+		repoUrl: "https://github.com/voxel51/fiftyone",
+		docsUrl: "https://docs.voxel51.com",
+	},
```
burtenshaw, line 67: As requested, it would be great to also include Argilla here 🤗.

```suggestion
	},
        argilla: {
                prettyLabel: "Argilla",
                repoName: "argilla",
                repoUrl: "https://github.com/argilla-io",
                docsUrl: "https://argilla-io.github.io/argilla",
	},
```

severo: Perfect, thanks!

packages/tasks/src/dataset-libraries.ts
```diff
@@ -53,6 +53,24 @@ export const DATASET_LIBRARIES_UI_ELEMENTS = {
...
+	argilla: {
+		prettyLabel: "Argilla",
+		repoName: "argilla",
+		repoUrl: "https://github.com/argilla-io/argilla",
+		docsUrl: "https://argilla-io.github.io/argilla",
+	},
```
julien-c, line 73: what would the snippet be when a dataset is advertising that it can be opened with Argilla?

burtenshaw: Good question. We're currently working on it.

There are two points of integration between a dataset repo and an Argilla dataset:
- a preconfigured argilla dataset with `.argilla` directory can be loaded in Argilla like this ([next release](https://github.com/argilla-io/argilla/pull/5161)):
```python
import argilla as rg

client = rg.Argilla(api_url="<api_url>", api_key="<api_key>")
dataset = rg.Dataset.from_hub(repo_id="<repo_id>")
```
- any dataset repo's rows can be loaded in Argilla, if the dataset's columns are compatible with the Argilla dataset settings:
```python
from datasets import load_dataset

dataset = rg.Dataset(...) # a dataset with compatible settings
hf_dataset = load_dataset("<api_url>")
dataset.records.log(records=hf_dataset)
```

severo: Note that this PR is not directly related to showing a snippet. The main goal is to show "argilla" under the "Libraries" tags group when a dataset has the "argilla" user-defined tag, and enable filtering the datasets on this dataset.

We don't plan to add snippets for other libraries (ie. under the "Use this dataset") in the near future, or at least, it's another discussion. See https://github.com/huggingface/dataset-viewer/issues/2986.

burtenshaw: Thanks @severo . I figured that, but just wanted to give context. We're actively working on our hub integrations so I'll follow https://github.com/huggingface/dataset-viewer/issues/2986.

severo:
can we merge this?

osanseviero (approved):
LGTM :+1:

PR #798, opened 2024-07-11
Add Video-Text-to-Text task 

merveenoyan:
new task for VideoLLaVA/PLLaVA-like models cc @NielsRogge @zucchini-nlp

logo is a big weird tbh I couldn't find a better one it looks like this (feel free to recommend new ones)
Screenshot 2024-07-11 at 19 48 55


NielsRogge (approved):
Thanks for adding!

packages/tasks/src/pipelines.ts
```diff
@@ -652,6 +652,11 @@ export const PIPELINE_DATA = {
 		modality: "cv",
 		color: "indigo",
 	},
+	"video-text-to-text": {
```
osanseviero, line 655: There are no models nor datasets with this tag yet (see https://huggingface.co/models?other=video-text-to-text&sort=trending for example). I would add `hideInModels` and `hideInDatasets` to avoid having big buttons not leading to any repos

osanseviero (commented):
Cool :fire:

packages/tasks/src/pipelines.ts
merveenoyan: there's a lot of models and some datasets too, I opened PRs to a lot of them

zucchini-nlp: btw does this tag apply for video captioning datasets?

merveenoyan: @zucchini-nlp if they have to input "caption this video"-like prompt yes, if not, we need to have a separate task called `video-to-text` (e.g. like difference between blip and llava)

zucchini-nlp: ah, i see, thanks!

NielsRogge: Yes for pure captioning we'd need the separate `video-to-text` task (similar to `image-to-text`)

merveenoyan: @osanseviero some PRs are merged, should I merge this?

merveenoyan:
@pcuenca can you review this?

packages/tasks/src/pipelines.ts
pcuenca: I would still `hideInDatasets`

packages/widgets/src/lib/components/Icons/IconVideoTextToText.svelte
pcuenca, line 1: noob here, is there a way to preview this?

pcuenca (approved):
There are a few models now (and more where the PRs have not been merged yet). I'm supportive of merging after including `hideInDatasets`, just like in `image-text-to-text`.

merveenoyan:
@osanseviero can you review this again?

osanseviero:
There are open comments from @pcuenca

packages/tasks/src/pipelines.ts
```diff
@@ -652,6 +652,11 @@ export const PIPELINE_DATA = {
...
+		name: "Video-Text-to-Text",
+		modality: "multimodal",
+		color: "blue",
```
pcuenca, line 658:
```suggestion
		color: "blue",
		hideInDatasets: true,
```

As mentioned, I think we can merge if `hideInDatasets` is added. Do you see a problem with it @merveenoyan?

PR #800, opened 2024-07-15
avoid duplicate header, check the ancestor origin

enzostvs:
the condition wasn't good due to the iframe, so now im checking the ancestor origin instead of .origin

### DIFF
--- a/packages/space-header/src/index.ts
+++ b/packages/space-header/src/index.ts
@@ -13,7 +13,7 @@ async function main(initialSpace: string | Space, options?: Options) {
 		window.location?.ancestorOrigins ?? {
 			0: window.document.referrer,
 		}
-	).some((origin) => origin.startsWith("https://huggingface.co"));
+	).some((origin) => new URL(origin)?.origin === "https://huggingface.co");
 	if (has_huggingface_ancestor) return;
 
 	inject_fonts();
--- a/packages/space-header/src/index.ts
+++ b/packages/space-header/src/index.ts
@@ -9,7 +9,10 @@ import { inject } from "./inject";
 async function main(initialSpace: string | Space, options?: Options) {
 	if (window === undefined) return console.error("Please run this script in a browser environment");
 	// Don't run on huggingface.co to avoid duplicate headers
-	if (window.location?.origin === "https://huggingface.co") return;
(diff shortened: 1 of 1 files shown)


packages/space-header/src/index.ts
```diff
@@ -9,7 +9,10 @@ import { inject } from "./inject";
 async function main(initialSpace: string | Space, options?: Options) {
 	if (window === undefined) return console.error("Please run this script in a browser environment");
 	// Don't run on huggingface.co to avoid duplicate headers
-	if (window.location?.origin === "https://huggingface.co") return;
+	const has_huggingface_ancestor = Object.values(window.location?.ancestorOrigins ?? {}).some((origin) =>
```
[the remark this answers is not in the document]
enzostvs: arf! good point @coyotte508 
Just pushed a wordaround for Firefox, wdyt ?

packages/space-header/src/index.ts
```diff
@@ -9,7 +9,12 @@ import { inject } from "./inject";
 async function main(initialSpace: string | Space, options?: Options) {
 	if (window === undefined) return console.error("Please run this script in a browser environment");
 	// Don't run on huggingface.co to avoid duplicate headers
-	if (window.location?.origin === "https://huggingface.co") return;
+	const has_huggingface_ancestor = Object.values(
+		window.location?.ancestorOrigins ?? {
+			0: window.document.referrer,
+		}
+	).some((origin) => origin.includes("huggingface.co"));
```
coyotte508, line 16: I think the `includes` check is too simplistic, eg `abchuggingface.co` would work.

Is it possible to have a better check, eg with `new URL(x).origin === "https://huggingface.co"` or something?

enzostvs: Yes sure
I switched to a .startsWith, because referrer could be https://huggingface.co/spaces/xxxx, while ancestorOrigins is just the domain origin, so should work in both case

coyotte508: `startsWith` is also vulnerable - eg `huggingface.co.evil.com`, can we use `new URL(x).origin` and do an equality check?

enzostvs: should be good! cc @coyotte508 tell me what do you think

coyotte508 (approved):

Issue #794, opened 2024-07-11
GGUF:  the file quantization type is not the GGMLQuantizationType.


snowyu:
There are two kinds of quantization in llama.cpp, don't confuse them:

* GGMLQuantizationType(`ggml_type`): for tensor https://github.com/ggerganov/llama.cpp/blob/7a221b672e49dfae459b1af27210ba3f2b5419b6/ggml/include/ggml.h#L354
* The GGUF file quantization type(`general.file_type`):  for file https://github.com/ggerganov/llama.cpp/blob/7a221b672e49dfae459b1af27210ba3f2b5419b6/include/llama.h#L131

If "general.file_type" is not configured, the following algorithm is used to guess the quantization type of the file:

https://github.com/ggerganov/llama.cpp/blob/7a221b672e49dfae459b1af27210ba3f2b5419b6/src/llama.cpp#L3751C1-L3802C65

julien-c:
Yes. What is the precise issue in this repo?

snowyu:
@julien-c 

1. The "general.file_type" missing enum FileQuantizationType like `GGMLQuantizationType`
2. Should guess the fileQuantType from metadata if no  "general.file_type".

And some tests incorrectly use the `GGMLQuantizationType` as "general.file_type":

https://github.com/huggingface/huggingface.js/blame/8d6fe81cd25936f65975d65eade246064ad48f7b/packages/gguf/src/gguf.spec.ts#L137

https://github.com/huggingface/huggingface.js/blame/8d6fe81cd25936f65975d65eade246064ad48f7b/packages/gguf/src/gguf.spec.ts#L174

julien-c:
maybe cc @ngxson (not sure)

ngxson:
Yes you’re correct @snowyu . general.file_type is the quantization scheme (i.e MOSTLY_*). I will push a fix later (sorry I’m quite busy atm)

@julien-c FYI, it’s because quantized model usually use mixed types. For example norm layer can always stay at f32 or f16, while other tensors can be Qk. This improves model performance with a little cost of space. Hence the word « mostly » used in type name

mishig25:
@snowyu thanks for your comment !

> The "general.file_type" missing enum FileQuantizationType like GGMLQuantizationType

[...]

PR #806, opened 2024-07-17
feat: add GGMLFileQuantizationType and apply to test

snowyu:
@mishig25 that's it for #794


julien-c:
cc @ngxson too

ngxson (approved):
LGTM. Thanks :+1:

julien-c (approved):
will let @mishig25 do a final review and merge 👌 

Thanks a lot @snowyu!

ngxson:
FYI, I added the `MOSTLY_` prefix in the last commit, to better reflect the type name from ggml (see [here](https://github.com/ggerganov/llama.cpp/blob/c8ddce85606d9fb6e30745b6e4fe103eecadc73f/ggml/include/ggml.h#L412-L440))

The reason is because many operations in ggml only support F32 for 1d tensors. So in fact, gguf file is never "purely" quantized, but rather being a mix between quantized type and F32.

julien-c:
BTW, i also propose to display the enum's key name in a tooltip inside the GGUF file viewer, like this:

image

([internal PR](https://github.com/huggingface-internal/moon-landing/pull/10943))

julien-c (approved):

julien-c:
i'll let you merge @ngxson!

snowyu:
@ngxson be careful,  the const is not in `ggml.h`, it's in [llama.h](https://github.com/ggerganov/llama.cpp/blob/7a221b672e49dfae459b1af27210ba3f2b5419b6/include/llama.h#L131).

ngxson:
Yeah I linked to the incorrect file, but the content is not changed anyway, I only added `MOSTLY_` on top of your commit. (So everything is still correct)

PR #813, opened 2024-07-19
Add possibility to collapse the header

enzostvs:
Demo:
https://github.com/user-attachments/assets/61c990a1-33cf-4368-bd8c-a43b71187778


enzostvs:
Idea:
The space icon could be a link to the space ?

pngwn:
Or as an alternative the space icon could take the place of the arrow/ maximise icon.

pngwn:
But thank you @enzostvs. I'll give this a test and see how it works in a gradio UI.

coyotte508:
Conceptual LGTM, feel free to merge :)

pngwn:
Made some tweaks after chatting with @julien-c @gary149 @coyotte508.

- The minimise button just removes the header now, so I have removed the unnessecary code.
- Added a way to remove the element from the page. We discussed a few options but I just went with the simplest and safest, an object containing the element. This allows us to expand this interface in the future without breaking changes but doesn't tie us into any specific cleanup API cc @coyotte508.

packages/space-header/src/header/components/content/index.ts
```diff
@@ -0,0 +1,26 @@
+import type { Space } from "../../../type";
+import { Like } from "../like";
+import { Avatar } from "./avatar";
+import { Namespace } from "./namespace";
+import { Separation } from "./separation";
+import { Username } from "./username";
+
+export const Content = (space: Space): HTMLDivElement => {
+	console.log(space);
```
pngwn, line 9: remove

packages/space-header/src/header/components/like/index.ts
```diff
@@ -14,9 +14,10 @@ export const Like = (space: Space): HTMLAnchorElement => {
 	box.style.display = "flex";
 	box.style.flexDirection = "row";
 	box.style.alignItems = "center";
-	box.style.margin = "0 12px";
+	box.style.margin = "0 0 0 12px";
 	box.style.fontSize = "14px";
 	box.style.paddingLeft = "4px";
+	box.style.textDecoration = "none";
```
pngwn, line 20: Fixes styling bug.

packages/space-header/src/index.ts
```diff
@@ -23,10 +23,16 @@ async function main(initialSpace: string | Space, options?: Options) {
 	if (typeof initialSpace === "string") {
 		space = await get_space(initialSpace);
 		if (space === null) return console.error("Space not found");
+	} else {
+		space = initialSpace;
 	}
```
pngwn, line 28: There was a bug with passing the space information directly, we checked the  `initialSpace` value but never assigned it to there `space` variable if it isn't a string. This is what gets passed along, so this mechanism wasn't working previously.

coyotte508 (approved):

PR #801, opened 2024-07-15
Add 3d task descriptions

dylanebert:
Added contents of image-to-3d and text-to-3d tasks

PR #814, opened 2024-07-20
3D task revisions

dylanebert:
Follow-up from #801

### DIFF
--- a/packages/tasks/src/tasks/image-to-3d/about.md
+++ b/packages/tasks/src/tasks/image-to-3d/about.md
@@ -14,9 +14,9 @@ Meshes are the standard representation of 3D in industry.
 
 ### Inference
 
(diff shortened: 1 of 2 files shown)


dylanebert:
I've added revisions based on comments that I missed in the diff of #801

dylanebert:
Regarding the [task output](https://github.com/huggingface/huggingface.js/blob/8a9ebac778bb1ab1e51fa570a3374dec3142dd9c/packages/tasks/src/tasks/image-to-3d/data.ts#L22C4-L26C6), which is currently a text filename - typically, models return the path to the 3D file as output. [Gradio Model3D](https://www.gradio.app/docs/gradio/model3d) accepts 3D filepaths as input, but not images.

Is there a more elegant way to handle this?

merveenoyan (approved):
thanks a lot!!!

PR #816, opened 2024-07-23
Handle user and org avatars

pngwn:
Currently if there is no avatar then the space header can display a broken image link.

This PR adds a simple check to see if the avatar exists. If it doesn't then we don't add the DOM for it.

I decided to do it outside of the 'component' rendering just to keep things cleaner + sync after the main function but I'm happy to change this.

Could be optimised to make the fetches in parallel but the difference is probably minor in most cases.

Screenshot:

Screenshot 2024-07-23 at 10 19 56

cc @enzostvs @coyotte508 @julien-c

### DIFF
--- a/packages/space-header/src/network_utils.ts
+++ b/packages/space-header/src/network_utils.ts
@@ -4,15 +4,18 @@ export const get_space = async (space_id: string): Promise<Space | null> => {
 	try {
 		const response = await fetch(`https://huggingface.co/api/spaces/${space_id}`);
 		const data = await response.json();
+		console.log(data);
 		return data as Space;
 	} catch (error) {
 		return null;
 	}
 };
 
-export const check_avatar = async (username: string): Promise<boolean> => {
+export const check_avatar = async (username: string, type: "user" | "org" = "user"): Promise<boolean> => {
+	const route = type === "user" ? "users" : "organizations";
+
 	try {
-		const response = await fetch(`https://huggingface.co/api/users/${username}/avatar`);
+		const response = await fetch(`https://huggingface.co/api/${route}/${username}/avatar`);
 		return response.ok;
 	} catch (error) {
 		return false;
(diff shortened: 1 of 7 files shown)


pngwn:
Update: I have also now handled the case of user/ or avatars which was causing the original issue. I've left the conditional logic in just in case but can remove it if needed.

packages/space-header/src/network_utils.ts
```diff
@@ -0,0 +1,23 @@
...
+export const check_avatar = async (username: string, type: "user" | "org" = "user"): Promise<boolean> => {
+	const route = type === "user" ? "users" : "organizations";
+
+	try {
+		const response = await fetch(`https://huggingface.co/api/${route}/${username}/avatar`);
+		return response.ok;
+	} catch (error) {
+		return false;
+	}
+};
```
coyotte508, line 23: Let's keep that in two separate files IMO (inside a specific folder if you want)

Or when @enzostvs comes back he can decide :)

pngwn:
We can also close this if it is handled in the hub, some of this code is redundant if that PR gets merged. 

We could keep the defensive conditional if we wanted to be safe though.

julien-c:
@pngwn we do need the user vs orgs handling, but i'll let others review!

pngwn:
I'll leave this here and let @enzostvs weigh in when they are back as it is their baby. Happy to make any changes.

Unless we want to get a fix out sooner?

enzostvs (approved):
Thanks for making the change, I hadn't taken orgs into account unfortunately.
I just split the utils function in 2 files, otherwise the change is good for me! 👌

coyotte508 (approved):
@enzostvs feel free to merge

packages/space-header/src/utils/get_space.ts
```diff
@@ -1,9 +1,10 @@
-import type { Space } from "./type";
+import type { Space } from "./../type";
 
 export const get_space = async (space_id: string): Promise<Space | null> => {
 	try {
 		const response = await fetch(`https://huggingface.co/api/spaces/${space_id}`);
 		const data = await response.json();
+		console.log(data);
```
julien-c, line 7: remaining console.log, intended?

enzostvs: no, will remove it, ty

PR #818, opened 2024-07-23
feat: add Ollama to local-apps.ts

sammcj:
I saw there were a couple of other PRs to add Ollama but they had problems with the deeplinks, however - I think this could possibly be a solution - instead of trying to link to an exact model - this would search for the model on the hub.

Another option might be to do something similar to llama.cpp using Ollama's import command with the GGUF from huggingface - https://github.com/ollama/ollama/blob/main/docs/import.md

### DIFF
--- a/packages/tasks/src/local-apps.ts
+++ b/packages/tasks/src/local-apps.ts
@@ -91,6 +91,13 @@ export const LOCAL_APPS = {
 		displayOnModelPage: isGgufModel,
 		snippet: snippetLlamacpp,
 	},
+	ollama: {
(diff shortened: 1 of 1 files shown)


Vaibhavs10:
Duplicate of this: https://github.com/huggingface/huggingface.js/pull/687

Please see the rationale/ reasoning behind it in the PR.

julien-c:
ok we might actualy revive this 🔥  see https://github.com/huggingface/huggingface.js/pull/687#issuecomment-2411824724

PR #822, opened 2024-07-26
Add polars to the list of supported dataset libraries

severo:
cc @nameexhaustion

Preliminary step before adding Polars in the Hub's code (cc https://github.com/huggingface/dataset-viewer/pull/2997 for reference)

### DIFF
--- a/packages/tasks/src/dataset-libraries.ts
+++ b/packages/tasks/src/dataset-libraries.ts
@@ -71,6 +71,12 @@ export const DATASET_LIBRARIES_UI_ELEMENTS = {
 		repoUrl: "https://github.com/argilla-io/argilla",
 		docsUrl: "https://argilla-io.github.io/argilla",
 	},
+	polars: {
+		prettyLabel: "Polars",
+		repoName: "polars",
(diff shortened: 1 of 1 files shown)


lhoestq (approved):

packages/tasks/src/dataset-libraries.ts
```diff
@@ -71,6 +71,12 @@ export const DATASET_LIBRARIES_UI_ELEMENTS = {
 		repoUrl: "https://github.com/argilla-io/argilla",
 		docsUrl: "https://argilla-io.github.io/argilla",
 	},
+	polars: {
+		prettyLabel: "Polars",
+		repoName: "polars",
+		repoUrl: "https://github.com/pola-rs/polars",
+		docsUrl: "https://docs.pola.rs/",
```
julien-c, line 78: should be the docs in our docs

severo: OK? i'll change for all the libs that we have docs for

severo: https://github.com/huggingface/huggingface.js/pull/825

Issue #612, opened 2024-04-08
[ASR Widget] “Browse for file” is unresponsive


ymoslem:
For automatic speech recognition models like [Whisper](https://huggingface.co/openai/whisper-large-v3), the “Browse for file” button is not responsive. When clicking it, the file navigator opens; however, when selecting an audio file, nothing happens. On the other hand, “Record from browser” works just fine. Also, drag and drop works.

## How to reproduce

* Go to [Whisper](https://huggingface.co/openai/whisper-large-v3) page (or any ASR model)
* Click “Browse for file”
* Browse for an audio file

## Result

I cannot see an error; it seems that the page reloads when the “Browse for file” is clicked. When we select an audio file, its name does not appear on the screen. When we click “Compute”, an error says “You must select or record an audio file” because there is no audio file uploaded.

coyotte508:
Fix works: https://huggingface.co/openai/whisper-large-v3

PR #824, opened 2024-07-28
Fix file upload button

dlavrenuek:
Fixes https://github.com/huggingface/huggingface.js/issues/612

### Description

Fixes the file upload button.

The problem is caused by `<button>` element that wraps file upload input having no `type` attribute. When used inside a form it defaults to `type="submit"` ([REF](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button)) which causes the page to reload (or submitting the form to the same url) at the same time as opening the upload dialog. The upload dialog itself does nothing in this case because it is mounted to the old page (before reload).

Bildschirmfoto 2024-07-28 um 19 48 31


coyotte508 (approved):
Thank you for the contribution!

PR #825, opened 2024-07-29
use HF docs

severo:


julien-c (approved):
thanks, that way it's more consistent with models

PR #831, opened 2024-08-03
Add sam2 to libraries

merveenoyan:
Enable stats for sam2 models


merveenoyan:
@osanseviero can you take a look? 👀

packages/tasks/src/model-libraries.ts
```diff
@@ -435,6 +435,13 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		repoUrl: "https://github.com/jbloomAus/SAELens",
 		filter: false,
 	},
+	"sam": {
```
osanseviero, line 438: This would impact lots of models https://huggingface.co/models?other=sam , including SAM1 release

Should we use `sam2` here?

osanseviero, line 438:
```suggestion
	"sam2": {
```

packages/tasks/src/model-libraries.ts
```diff
@@ -435,6 +435,13 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		repoUrl: "https://github.com/jbloomAus/SAELens",
 		filter: false,
 	},
+	sam2: {
+		prettyLabel: "sam",
+		repoName: "sam",
```
osanseviero, line 440: These should also say SAM2 I think

merveenoyan:
I thought of leaving pretty name the same due to model type but fixed

Vaibhavs10 (approved):
Was browsing through open PRs, took the liberty to take a pass through! 🤗 

GG & thanks! There are models visible through the `sam2` tag as well: https://huggingface.co/models?other=sam2

packages/tasks/src/model-libraries.ts
```diff
@@ -436,6 +436,13 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		snippets: snippets.saelens,
 		filter: false,
 	},
+	sam2: {
+		prettyLabel: "sam2",
+		repoName: "sam2",
```
osanseviero, line 441: The repo name is segment-anything-2 actually, I would go with that

PR #833, opened 2024-08-04
Add LocalAI to local-apps

mudler:
Hello :wave: 

This PR adds LocalAI (https://localai.io) to the huggingface local-apps. It natively supports models directly from huggingface in gguf format by running localai e.g. with `huggingface://repo-id/file`  ( docs: https://localai.io/docs/getting-started/models/#run-models-via-uri ) .

It also supports transformers - however it needs a configuration file and can't be natively used, so piggybacking only to gguf files for now.


Vaibhavs10 (commented):
Conceptually LGTM! I haven't tested this yet, does it require any specific setup for Mac or CUDA devices?

mudler:
> Conceptually LGTM! I haven't tested this yet, does it require any specific setup for Mac or CUDA devices?

Only if using docker - you would need to pull specific images for e.g. CUDA. If using the single-binary both Mac and CUDA would work. That is however handled automatically by using the install script!

packages/tasks/src/local-apps.ts
```diff
@@ -126,6 +136,13 @@ export const LOCAL_APPS = {
 		deeplink: (model, filepath) =>
 			new URL(`lmstudio://open_from_hf?model=${model.id}${filepath ? `&file=${filepath}` : ""}`),
 	},
+	"localai": {
```
mudler, line 139:
```suggestion
	localai: {
```

packages/tasks/src/local-apps.ts
```diff
@@ -99,6 +99,16 @@ const snippetLlamacpp = (model: ModelData, filepath?: string): LocalAppSnippet[]
 	];
 };
 
+const snippetLocalAI = (model: ModelData, filepath?: string): string[] => {
+	return [
+		`# Option 1: use LocalAI with docker
+docker run -p 8080:8080 --name localai -v $PWD/models:/build/models localai/localai:latest-cpu huggingface://${model.id}/${filepath ?? "{{GGUF_FILE}}"}`,
```
mudler, line 105:
```suggestion
docker run -p 8080:8080 --name localai -v $PWD/models:/build/models localai/localai:latest-cpu huggingface://${
			model.id
		}/${filepath ?? "{{GGUF_FILE}}"}`,
```

enzostvs:
Hi @mudler 
Can you send us your logo in SVG format please ? 🤗

packages/tasks/src/local-apps.ts
```diff
@@ -99,6 +99,18 @@ const snippetLlamacpp = (model: ModelData, filepath?: string): LocalAppSnippet[]
 	];
 };
 
+const snippetLocalAI = (model: ModelData, filepath?: string): string[] => {
+	return [
+		`# Option 1: use LocalAI with docker
```
Vaibhavs10, line 104: One last comment on this:

Let's switch the order of the snippets, as I assume the number of people who will use this with docker are far less than directly from the binaries.

Question for @enzostvs - do you think doing something like we did for llama.cpp here: https://github.com/huggingface/huggingface.js/blob/3a8651fbc6508920475564a692bf0e5b601d9343/packages/tasks/src/local-apps.ts#L65 (since we have two snippets make sense)

mudler: inverted as requested in https://github.com/huggingface/huggingface.js/pull/833/commits/3fca3d6b1bcedbd1306eedbf4269667f360beae4 !

Vaibhavs10: Perfect! @mudler sorry for the back and forth, do you think you'll be able to split the snippets the way I linked above? This way the users will have a much better `readability` experience on the hub.

mudler: > Perfect! @mudler sorry for the back and forth, do you think you'll be able to split the snippets the way I linked above? This way the users will have a much better `readability` experience on the hub.

I can try, javascript isn't really my cup of tea :sweat_smile:

mudler: Updated in https://github.com/huggingface/huggingface.js/pull/833/commits/1d5064ce6335f91841b8d4cf6816c31dc05061c0 - let me know if that looks good!

packages/tasks/src/local-apps.ts
```diff
@@ -99,6 +99,29 @@ const snippetLlamacpp = (model: ModelData, filepath?: string): LocalAppSnippet[]
 	];
 };
 
+const snippetLocalAI = (model: ModelData, filepath?: string): LocalAppSnippet[] => {
+	const command = (binary: string) =>
+		[
+			`huggingface://${model.id}/${filepath ?? "{{GGUF_FILE}}"}`,
```
pcuenca, line 105: The actual commands seem to be missing (i.e., `binary` is unused) :)

Vaibhavs10 (approved):
Nice! LGTM! Can you run the linter on this so that the code is formatted properly? :🙏

If you are using VS code then you can just install the `prettier` extension and it'd automatically format it for you upon save! 🤗

mudler:
> Nice! LGTM! Can you run the linter on this so that the code is formatted properly? :🙏
> 
> If you are using VS code then you can just install the `prettier` extension and it'd automatically format it for you upon save! 🤗

that did the trick, thank you!

enzostvs (approved):

Vaibhavs10:
Thanks again for the contribution @mudler 🤗

mudler:
> Thanks again for the contribution @mudler 🤗

Thank you for the quick review! :heart_hands:

enzostvs:
Hi @mudler 
LocalAI has been added to the local apps on the HF hub 🤗 
You could communicate on it too, would be really nice! 🥳 
Thanks again for your contribution!

Screenshot 2024-08-14 at 9 36 05 AM
Screenshot 2024-08-14 at 9 39 47 AM

PR #838, opened 2024-08-06
Add support for Relik

osanseviero:
Library: https://github.com/SapienzaNLP/relik
Some model repos (adding `library_name` soon): https://huggingface.co/collections/sapienzanlp/relik-retrieve-read-and-link-665d9e4a5c3ecba98c1bef19

The repos already use `config.yaml` so downloads work out of the box already

cc @Riccorl

### DIFF
--- a/packages/tasks/src/model-libraries-snippets.ts
+++ b/packages/tasks/src/model-libraries-snippets.ts
@@ -363,6 +363,12 @@ export const pyannote_audio = (model: ModelData): string[] => {
 	return pyannote_audio_model(model);
 };
 
+const relik = (model: ModelData) => [
+	`from relik import Relik
+ 
+relik = Relik.from_pretrained("${model.id}")`,
+];
+
 const tensorflowttsTextToMel = (model: ModelData): string[] => [
 	`from tensorflow_tts.inference import AutoProcessor, TFAutoModel
 
--- a/packages/tasks/src/model-libraries.ts
+++ b/packages/tasks/src/model-libraries.ts
@@ -429,6 +429,13 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		filter: false,
 		countDownloads: `path:"tokenizer.model"`,
 	},
+	relik: {
+		prettyLabel: "Relik",
+		repoName: "Relik",
+		repoUrl: "https://github.com/SapienzaNLP/relik",
+		snippets: snippets.relik,
+		filter: false,
+	},
 	saelens: {
 		prettyLabel: "SAELens",
 		repoName: "SAELens",


Vaibhavs10 (approved):
Looks good, we'd need to open PRs to the model repos before merging. As there are no repos listed with `relik` tag as of now: https://huggingface.co/models?other=relik

NielsRogge (approved):
Thanks for adding!

Riccorl:
@Vaibhavs10 I have added the `relik` tag to our pretrained configurations that work directly with the `Relik.from_pretrained()` method.

packages/tasks/src/model-libraries-snippets.ts
```diff
@@ -363,6 +363,12 @@ export const pyannote_audio = (model: ModelData): string[] => {
 	return pyannote_audio_model(model);
 };
 
+const relik = (model: ModelData) => [
```
pcuenca, line 366:
```suggestion
export const relik = (model: ModelData) => [
```

pcuenca:
Noob question, haven't used the library yet. I'm not sure I'm understanding the pieces right, but it looks like a repo such as https://huggingface.co/sapienzanlp/relik-relation-extraction-nyt-large is a pointer for https://huggingface.co/sapienzanlp/relik-reader-deberta-v3-large-nyt (see https://huggingface.co/sapienzanlp/relik-relation-extraction-nyt-large/blob/main/config.yaml#L12), as well as the index / embeddings here: https://huggingface.co/sapienzanlp/relik-retriever-small-nyt-document-index (from https://huggingface.co/sapienzanlp/relik-relation-extraction-nyt-large/blob/main/config.yaml#L9).

Are all of them going to have the relik `library_name`? Will we be counting downloads multiple times?

LittlePea13:
Hi there @pcuenca, as of now you are right, the "upper" config (i.e. [sapienzanlp/relik-relation-extraction-nyt-large](https://huggingface.co/sapienzanlp/relik-relation-extraction-nyt-large)) points to the different components the overall system requires (i.e. reader and retrievers). We found this easier, so that different configs can point to the same components, i.e. the same reader can use a small retriever or a base one, different index, and such. This way it leads to no need to redownload the same components in different "upper" configs, if that makes sense.

As @Riccorl said, for now we tagged with the library name only the "upper" configs, that means only those that are loaded with `Relik.from_pretrained()`. Not sure what this means in terms of download counts nor if you have another suggested way of dealing with these matters.

pcuenca:
Hello @LittlePea13 👋, thanks for the clarifications! Tagging the "upper" configs makes sense to me. With this setup, I believe that when you use `Relik.from_pretrained()` you'll get +1 added to the download counts of the config, the reader and the retriever, but that's perfectly ok as long as it's clear how it works and you can get useful information from the data. For example, you may get insights on how often individual components are used as part of the pipelines, in addition to knowing the high-level configurations that are used the most.

Vaibhavs10 (commented):
I have no clue why the linter fails, prettier doesn't complain locally 😅

Also, sorry for messing up the merge commit, I was just curious to see the reason why the CI was failing.

osanseviero:
Fixed lint issue in https://github.com/huggingface/huggingface.js/pull/838/commits/ca2a19c32fcda088d62f650e21e7acf37972b9e8

I'll merge this ahead

PR #840, opened 2024-08-07
add llama-cpp-python snippet.

Vaibhavs10:


packages/tasks/src/model-libraries-snippets.ts
```diff
@@ -261,6 +261,24 @@ backbone = keras_nlp.models.Backbone.from_preset("hf://${model.id}")
 `,
 ];
 
+export const llama_cpp_python = (model: ModelData): string[] => [
+	`from llama_cpp import Llama
+llm = Llama.from_pretrained(
+repo_id="${model.id}",
+filename="<QUANT FILE NAME>",
```
Vaibhavs10, line 268: Not sure what's the best way to deal with quant file name here?

packages/tasks/src/model-libraries.ts
```diff
@@ -314,6 +314,13 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		filter: false,
 		countDownloads: `path:"liveportrait/landmark.onnx"`,
 	},
+	"llama-cpp-python": {
+		prettyLabel: "llama-cpp-python",
+		repoName: "llama-cpp-python",
+		repoUrl: "https://github.com/abetlen/llama-cpp-python",
+		snippets: snippets.llama_cpp_python,
+		filter: true,
```
julien-c, line 322:
```suggestion
```

julien-c: models won't be tagged explicitly as llama-cpp-python through https://github.com/huggingface-internal/moon-landing/pull/10885, only implicitly

packages/tasks/src/model-libraries-snippets.ts
```diff
@@ -261,6 +261,24 @@ backbone = keras_nlp.models.Backbone.from_preset("hf://${model.id}")
...
+llm.create_chat_completion(
+		messages = [
+			{
+				"role": "user",
+				"content": "What is the capital of France?"
+			}
+		]
+)
+	
+`,
```
julien-c, line 279:
```suggestion
	`from llama_cpp import Llama

llm = Llama.from_pretrained(
	repo_id="${model.id}",
	filename="{{GGUF_FILE}}",
)

llm.create_chat_completion(
		messages = [
			{
				"role": "user",
				"content": "What is the capital of France?"
			}
		]
)
```

julien-c (approved):
You can use `{{GGUF_FILE}}` to display a selector like for llama.cpp, i've opened a PR to support this in our internal codebase

packages/tasks/src/model-libraries.ts
julien-c: let's remove it ok?

julien-c (approved):
let's merge to get the internal PR to be green:)

PR #740, opened 2024-06-06
Add new text-to-image LocalApp JoyFusion(macOS)

jiangdi0924:
JoyFusion is a native application for macOS , built upon Stable Diffusion and CoreML technologies. It can locally run CoreML-based text-to-image models.

Currently, only `CoreML model ZIP file` imports are supported, but more file types will be supported in the future.
For now, you can find CoreML files in the following repos or community:
https://huggingface.co/Norton0924/Joyfusion/tree/main
https://huggingface.co/coreml-community

👇Here is  App Icon SVG
joyfusion (1)

👇Demo Video

https://github.com/huggingface/huggingface.js/assets/21988177/2db2c186-a0dc-4707-8ce5-de1e74071edb

PR #844, opened 2024-08-07
add new type hasZipFilePaths

enzostvs:
Added into HF Hub a new `hasZipFilePaths` field to know if there is at least 1 zip file in the model repository, could be use into hf.js tasks too, especially for local-apps which are using .zip files only (like https://github.com/huggingface/huggingface.js/pull/740)

`hasZipFilePaths` is a boolean


julien-c (commented):
hmm this sounds a bit specific use case, i think we can find something simpler

Let me think about it tomorrow

julien-c:
see https://github.com/huggingface/huggingface.js/pull/740#issuecomment-2274151754

PR #845, opened 2024-08-08
Tasks: add watch mode

krampstudio:
```
pnpm install
cd pacakges/tasks
pnpm run watch
```

cc @coyotte508 when you're back


packages/tasks/package.json
```diff
@@ -25,6 +25,7 @@
 		"format:check": "prettier --check .",
 		"prepublishOnly": "pnpm run inference-codegen && git diff --name-only --exit-code src && pnpm run build",
 		"build": "tsup src/index.ts --format cjs,esm --clean && tsc --emitDeclarationOnly --declaration",
+		"watch": "tsup src/index.ts --format cjs,esm --clean --watch",
```
julien-c, line 28: there's no need for the `tsc --declaration` part of the watch command?

julien-c (approved):
Thanks, this is super useful

packages/tasks/package.json
krampstudio: oh, yes, I thought it was only for type checking but it generates the declarations map too. Fixing it

packages/tasks/package.json
```diff
@@ -25,6 +25,9 @@
 		"format:check": "prettier --check .",
 		"prepublishOnly": "pnpm run inference-codegen && git diff --name-only --exit-code src && pnpm run build",
 		"build": "tsup src/index.ts --format cjs,esm --clean && tsc --emitDeclarationOnly --declaration",
+		"watch:export": "tsup src/index.ts --format cjs,esm --watch",
+		"watch:types": "tsc --emitDeclarationOnly --declaration --watch",
+		"watch": "npm-run-all --parallel watch:export watch:types",
```
julien-c, line 30: parallel is fine, or should we first tsup, then tsc (i.e. sequentially)?

(i don't know enough to know)

krampstudio: the process won't exit in watch mode, so only the 1st process will run in sequential mode.

PR #852, opened 2024-08-15
wrong deeplink for diffusionbee

enzostvs:
linked to https://huggingface.slack.com/archives/C07777H04GL/p1723671712888659

### DIFF
--- a/packages/tasks/src/local-apps.ts
+++ b/packages/tasks/src/local-apps.ts
@@ -232,7 +232,7 @@ export const LOCAL_APPS = {
 		mainTask: "text-to-image",
 		macOSOnly: true,
 		displayOnModelPage: (model) => model.library_name === "diffusers" && model.pipeline_tag === "text-to-image",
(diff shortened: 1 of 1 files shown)


packages/tasks/src/local-apps.ts
```diff
@@ -232,7 +232,7 @@ export const LOCAL_APPS = {
 		mainTask: "text-to-image",
 		macOSOnly: true,
 		displayOnModelPage: (model) => model.library_name === "diffusers" && model.pipeline_tag === "text-to-image",
-		deeplink: (model) => new URL(`diffusionbee://open_from_hf?model=${model.id}`),
+		deeplink: (model) => new URL(`https://diffusionbee.com/huggingface_import?model_id=${model.id}`),
```
julien-c, line 235:
```suggestion
		deeplink: (model) => new URL(`diffusionbee://huggingface_sd_model/${model.id}`),
```

would also work, which one do we prefer? cc @divamgupta

cf. our Slack discussion

PR #692, opened 2024-05-21
[jinja] Add support for macros and `| tojson`

xenova:
Adds support for Macros. See here for more information: https://jinja.palletsprojects.com/en/3.1.x/templates/#macros

cc @Rocketknight1

PR #854, opened 2024-08-17
feat(jinja): add string rstrip and lstrip methods

snowyu:


xenova:
Thanks so much for the PR! Could you restructure the unit tests in this way: https://github.com/huggingface/huggingface.js/pull/692/commits/3738c685f483d79712189be6196ae9634535246a? Thanks!

xenova (changes requested):
Thanks so much for the PR! Could you restructure the unit tests in this way: https://github.com/huggingface/huggingface.js/pull/692/commits/3738c685f483d79712189be6196ae9634535246a? Thanks!

Also, strangely enough, jinja doesn't have a `| ltrim` or `| rtrim` filter (only `| trim`; see [docs](https://jinja.palletsprojects.com/en/3.1.x/templates/#jinja-filters.trim)), so adding these two functions will be useful 👍

snowyu:
Done. 

btw, I've implemented user-defined filters using JavaScript's original value methods: `toValue()`, `toString()`, and `toJSON()` mainly. It's a huge change. I'm not sure if I should submit a pull request yet.

xenova (approved):
Tests pass - LGTM!

xenova:
btw, do you have an example of a chat template which uses these methods? Would be nice to add as an end-to-end unit test

snowyu:
@xenova here is my WIP project PPE([Programmable Prompt Engine](https://github.com/offline-ai/ppe)) CLI:https://github.com/offline-ai/cli. 

Any suggestions are very welcome.

My changes are here: https://github.com/isdk/ai-tool.js/tree/main/src/utils/prompt/template/jinja

unit tests:
* https://github.com/isdk/ai-tool.js/blob/main/src/utils/prompt/template/jinja/test/interpreter.test.js#L179
* https://github.com/isdk/ai-tool.js/blob/main/src/utils/prompt/hf-prompt-template.test.ts#L94

User story scenario:

I've written the PPE [json.ai.yaml](https://github.com/offline-ai/cli/blob/main/lib/json.ai.yaml) lib which convert any content to json.

I've found that LLM are better at understanding natural language than json.

So, the user can use the `Object::toString`(without modifing the json library) convert json to string(natural language).

```yaml
---
output:
  type: "object"
  properties:
    sentiment:
      type: "string"
      description: "Sentiment (positive or negative)"
content: >-
  Needed a nice lamp for my bedroom, and this one had additional storage and not too high of a price point. Got it fast.
  The string to our lamp broke during the transit and the company happily sent over a new one. Came within a few days as well. It was easy to put together.
---
|fn |-
  # add toString() method to output object
  function output$toString(value) {
    return 'the nature lang for output JSON Schema...'
  }
# call the json lib
-> json(content=content, output=output)
```

PR #856, opened 2024-08-17
Account for the fact that there non-llama.cpp gguf files now

julien-c:
[companion internal PR](https://github.com/huggingface-internal/moon-landing/pull/10946)

**TL;DR:** we want the ability to discriminate non-llama.cpp gguf repos

I've picked `${architecture}.context_length` as a property that, if it's in the gguf file, we assume it's compatible with llama.cpp

Another option was to check whether `architecture` is inside the list in `@huggingface/gguf` (and keep this list up to date with upstream, maybe through some CI)

Both options are possible, so let me know what you think is best

### DIFF
--- a/packages/tasks/src/local-apps.ts
+++ b/packages/tasks/src/local-apps.ts
@@ -58,6 +58,7 @@ export type LocalApp = {
 	  }
 );
 
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
 function isGgufModel(model: ModelData) {
 	return model.tags.includes("gguf");
 }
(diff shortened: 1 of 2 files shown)


ngxson (approved):
Cool feature!

For the CI issue: the `prettier` check is failed (not the `lint` check). You can fix by running `pnpm run format`

Vaibhavs10 (approved):
Sounds good, did couple of spot checks to find any offending repos by looking through random (unconventional) GGUFs.

This seems to be good proxy, good find 😄 

For me, context length is much better proxy than architecture (since we'll need to keep the architecture list fully synced at all times - which gets updated quite a bit)

osanseviero (approved):
LGTM, with the risk being that other projects use context_length in their GGUF as well

julien-c:
> (since we'll need to keep the architecture list fully synced at all times - which gets updated quite a bit)

Yes, but IMO in the future we (e.g. @ngxson 😁) can automate the syncing of our architecture list with upstream through some CI

PR #857, opened 2024-08-19
[Text-to-video] Update modality

NielsRogge:
I noticed the "text-to-video" task is currently under computer vision, should be multimodal.

### DIFF
--- a/packages/tasks/src/pipelines.ts
+++ b/packages/tasks/src/pipelines.ts
@@ -585,7 +585,7 @@ export const PIPELINE_DATA = {
 	},
 	"text-to-video": {
 		name: "Text-to-Video",
-		modality: "cv",
+		modality: "multimodal",
 		color: "green",
 	},
 	"image-text-to-text": {


pcuenca:
`text-to-image` is currently CV as well. There have been various discussions about what "multimodal" should include, the latest update was done [in this PR](https://github.com/huggingface/huggingface.js/pull/477) where some comments were shared. cc @merveenoyan as well for opinions / clarifications.

packages/tasks/src/pipelines.ts
```diff
@@ -585,7 +585,7 @@ export const PIPELINE_DATA = {
 	},
 	"text-to-video": {
 		name: "Text-to-Video",
-		modality: "cv",
+		modality: "multimodal",
```
osanseviero, line 588: This was done in https://github.com/huggingface/huggingface.js/pull/477 as part of existing multimodal discussions in Slack (search for this PR). I'm happy to reconsider this specific case

NielsRogge: For me, multimodal is anything that is not single modality

* If only images are involved => cv
* if only text is involed => nlp
* if image + text, or any other combination of modalities (like text and video) => multimodal

osanseviero: Image generation and captioning involve both text and images. I think the discussion had converged in cardinality of modalities of input or output. Please refer to the existing discussions

PR #859, opened 2024-08-21
Allow to customize parallelDownload for ggufAllShards

Kakulukian:
Could be related to https://huggingface.slack.com/archives/C02EMARJ65P/p1724166668261449

### DIFF
--- a/packages/gguf/src/gguf.ts
+++ b/packages/gguf/src/gguf.ts
@@ -9,6 +9,7 @@ export { GGUF_QUANT_DESCRIPTIONS } from "./quant-descriptions";
 
 export const RE_GGUF_FILE = /\.gguf$/;
 export const RE_GGUF_SHARD_FILE = /^(?<prefix>.*?)-(?<shard>\d{5})-of-(?<total>\d{5})\.gguf$/;
+const PARALLEL_DOWNLOADS = 20;
 
 export interface GgufShardFileInfo {
 	prefix: string;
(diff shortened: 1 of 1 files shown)


packages/gguf/src/gguf.ts
```diff
@@ -401,8 +401,12 @@ export async function ggufAllShards(
 		 */
 		fetch?: typeof fetch;
 		additionalFetchHeaders?: Record<string, string>;
+		parallelDownloads?: number;
 	}
 ): Promise<{ shards: GGUFParseOutput[]; parameterCount: number }> {
+	if (params?.parallelDownloads && params.parallelDownloads < 1) {
```
ngxson, line 407:
```suggestion
  const parallelDownloads = params?.parallelDownloads ?? 20;
	if (parallelDownloads < 1) {
```

packages/gguf/src/gguf.ts
```diff
@@ -413,7 +417,7 @@ export async function ggufAllShards(
 			urls.push(`${prefix}-${shardIdx.toString().padStart(5, "0")}-of-${total.toString().padStart(5, "0")}.gguf`);
 		}
 
-		const PARALLEL_DOWNLOADS = 20;
+		const PARALLEL_DOWNLOADS = params?.parallelDownloads ?? 20;
```
ngxson, line 420: Nitpicks: This variable should no longer be all caps, since it's no longer a constant.

My suggestion is to remove this line and replace line 423 with `parallelDownloads`

Pierrci (approved):

PR #875, opened 2024-08-29
Add diffusionkit to local-apps

ardaatahan:
Opening this PR as per @Vaibhavs10's suggestion [here](https://github.com/huggingface/huggingface.js/pull/869#pullrequestreview-2262720881) regarding DiffusionKit LocalApps integration.

This PR:

1. Adds `diffusionkit` to the LOCAL_APPS object in [local-apps.ts](https://github.com/huggingface/huggingface.js/blob/main/packages/tasks/src/local-apps.ts).
2. Adds `snippetDiffusionKit` function that generates the snippet string for `diffusionkit` in the same file.


Vaibhavs10 (approved):
Thanks a lot for opening this PR and sorry for the delay. Conceptually approving the PR as the snippet looks good to me.

Will let @enzostvs chime in about what else is needed w.r.t to showing the snippet for the Local App itself.

Note: Please make sure to run the linter to make the linter tests pass.

packages/tasks/src/local-apps.ts
```diff
@@ -127,6 +127,39 @@ const snippetLocalAI = (model: ModelData, filepath?: string): LocalAppSnippet[]
...
+			`  --seed 1001 \\`,
+			`  --step ${model.tags.includes("flux") ? 4 : 50} \\`,
+			`  --cfg ${model.tags.includes("flux") ? 0. : 5} \\`,
+			`  --output ~/Desktop/out.png`,
+		].join("\n");
+	return [
+		{
+			title: "Install from pip",
+			setup: "pip install diffusionkit",
+			content: command("diffusionkit-cli"),
```
julien-c, line 148: if binary is always the same mayber simpler to hardcode it a `const command = string`

julien-c (commented):
let's maybe wait till the binary is packaged w/o pip for us to ship  this one (as it's a bit redundant with the model-library, imo)

Vaibhavs10 (commented):
Just to summarise discussions on next steps w/ @ZachNagengast @atiorh 

Argmax will put DiffusionKit over on brew and we'll update snippets accordingly.

PR #879, opened 2024-08-31
register py-feat library

ljchang:
The goal of this PR is to register a new library [py-feat](https://py-feat.org/) with huggingface and add supported tasks of `image-feature-extraction`.

Please let me know if I should make any changes.


packages/tasks/src/library-to-tasks.ts
```diff
@@ -31,6 +31,7 @@ export const LIBRARY_TASK_MAPPING: Partial<Record<ModelLibraryKey, PipelineType[
 	peft: ["text-generation"],
 	"pyannote-audio": ["automatic-speech-recognition"],
 	"sentence-transformers": ["feature-extraction", "sentence-similarity"],
+	py-feat: ["image-feature-extraction"],
```
pcuenca, line 34:
```suggestion
	"py-feat": ["image-feature-extraction"],
```

packages/tasks/src/model-libraries.ts
```diff
@@ -446,6 +446,14 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		snippets: snippets.pyannote_audio,
 		filter: true,
 	},
+	"py-feat": {
+		prettyLabel: "py-feat",
```
pcuenca, line 450: This is for display purposes. I've seen both `Py-FEAT` (in the [repo](https://github.com/cosanlab/py-feat)) and `Py-Feat` (in the [docs](https://py-feat.org/pages/intro.html)).

packages/tasks/src/model-libraries.ts
```diff
@@ -446,6 +446,14 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
...
+		repoName: "py-feat",
+		repoUrl: "https://github.com/cosanlab/py-feat",
+		docsUrl: "https://py-feat.org/",
+		snippets: snippets.py-feat,
```
pcuenca, line 454: This should be a code snippet that is adapted to the current model the user is visiting in the Hub, telling them how to use it in the library when they click on the "Use this Model" button.

packages/tasks/src/model-libraries.ts
```diff
@@ -446,6 +446,14 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
...
+		filter: true,
```
pcuenca, line 455: For this to work, there should be models in the Hub tagged as being compatible with the library. For instance, https://huggingface.co/py-feat/retinaface would be a candidate.

pcuenca (commented):
Thanks for the contribution @ljchang! Made some initial comments, please let us know if you have any questions!

cc @merveenoyan

packages/tasks/src/model-libraries.ts
ljchang: Thanks, will add

packages/tasks/src/library-to-tasks.ts
ljchang: thanks will change to quotes

packages/tasks/src/model-libraries.ts
ljchang: Thanks @pcuenca for the feedback. I think I'm still a little confused about how code snippets work with huggingface. I have tried to include snippets of code for how to use each model in the model cards, but I'm not sure how to add tags to this that can be referenced in the snippets attribute. Maybe it's just best to remove for now?

ljchang:
Thanks @pcuenca for the fast comments! I updated the PR based on your suggestions and removed the snippets reference for now until I better understand how this works.

packages/tasks/src/model-libraries.ts
```diff
@@ -446,6 +446,13 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		snippets: snippets.pyannote_audio,
 		filter: true,
 	},
+	"py-feat": {
+		prettyLabel: "Py-Feat",
+		repoName: "py-feat",
```
Wauplin, line 451:
```suggestion
		repoName: "Py-Feat",
```

packages/tasks/src/model-libraries.ts
```diff
@@ -446,6 +446,13 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
...
+		repoUrl: "https://github.com/cosanlab/py-feat",
+		docsUrl: "https://py-feat.org/",
+		filter: true,
```
Wauplin, line 454: Marking `filter: true,` will add the library in the left sidebar on https://huggingface.co/models. This is usually reserved to popular libraries with at least hundreds of models as we want to keep the list short. 

image

```suggestion
		filter: false,
```

Setting `filter: false` don't mean that it'll not be possible to filter by this library, it just means that it'll not be listed on this page. It's still possible to filter by `py-feat` by going to a py feat model and clicking on the tag.

packages/tasks/src/model-libraries.ts
```diff
@@ -446,6 +446,13 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		snippets: snippets.pyannote_audio,
 		filter: true,
 	},
+	"py-feat": {
```
Wauplin, line 449: For this filter to work, you will need to tag your models with `py-feat`. This will make them listed in https://huggingface.co/models?other=py-feat (0 models at the moment).

To tag a model, you must add an attribute to the model card metadata. It must be either `library_name` or under the `tags` list. I opened https://huggingface.co/py-feat/retinaface/discussions/1 as an example for you to reproduce on other models.

packages/tasks/src/library-to-tasks.ts
```diff
@@ -31,6 +31,7 @@ export const LIBRARY_TASK_MAPPING: Partial<Record<ModelLibraryKey, PipelineType[
 	peft: ["text-generation"],
 	"pyannote-audio": ["automatic-speech-recognition"],
 	"sentence-transformers": ["feature-extraction", "sentence-similarity"],
+	"py-feat": ["image-feature-extraction"],
```
Wauplin, line 34: As far as I know, this list is used for the inference snippet on the model pages. Since `py-feat` is not integrated with our inference service, we shouldn't list it here.

If you want `py-feat` models to be tagged as `image-feature-extraction`, you must tag them as such in the model card metadata (either using `pipeline_tag` or `tags`). See PR https://huggingface.co/py-feat/retinaface/discussions/1

Wauplin (commented):
Hi @ljchang, thanks for the PR! I've left a few more comments.

From https://github.com/huggingface/huggingface.js/pull/879#discussion_r1740189883, I feel there is a confusion about what this PR does and how the Hub works.
1. Each model on the Hub can have tags mostly based on the model card metadata. I've created a PR on one of your models to show you how to do it (https://huggingface.co/py-feat/retinaface). You can use a script to modify these metadata for all of your models automatically. [Here is a guide](https://huggingface.co/docs/huggingface_hub/guides/model-cards) showing you how to do that using the `huggingface_hub` Python library.
2. Once a model have tags, the Hub (i.e. our servers) can detect things like library or task. Based on that, we can display a snippet of code which is supposed to be the same for all models of the same library -expect the model id changing-. This is not an obligation for a first integration but can be helpful for end users. See for example how [black-forest-labs/FLUX.1-dev](https://huggingface.co/black-forest-labs/FLUX.1-dev?library=diffusers) can be used with the `diffusers` library. This advantage of document the model this way compared to having it in the model card description is that you are guaranteed all `py-feat` models will have it and you'll be able to maintain the snippet from a single place (i.e. from this repo). 
3. I saw that you are using the `PyTorchModelHubMixin` to push models to the Hub. Is that correct? If that's the case, models should be loadable using a `model.from_pretrained(modelId)` approach (might need to be adapted depending on your lib). When you use the `PyTorchModelHubMixin`, it's possible to define a library_name and pipeline_tag directly in your lib' so that all users pushing a model to the Hub will automatically have the correct tags. Check out [this guide](https://huggingface.co/docs/huggingface_hub/guides/integrations#model-card) for more details.

All of this to say:
- let's fix the model cards for now (e.g. add library_name + pipeline_tag)
- let's add snippets in a future PR once basic integration will be completed
- (optional) let's reorganize py-feat to benefit from `PyTorchModelHubMixin` feature if that's what you are using 

Let me know if you have any questions!

packages/tasks/src/library-to-tasks.ts
ljchang: that makes sense. I'll remove it.

packages/tasks/src/model-libraries.ts
ljchang: Thanks @Wauplin this is all very helpful!

1. I updated all of my model cards to have a library and pipeline tag
2. set filter to false
3. removed py-feat from libraries-to-task.ts
4. I'll submit a new PR in the future if we ever have a more standardized approach to access models.
5. We have added `PyTorchModelHubMixin` mixin to all of our pytorch models and have played with the upload and download features. We ran into an issue using the upload method with safetensors, where the weights somehow got changed through that process and messed up inference.  I wasn't able to find other reports of this happening, nor did I investigate why. I instead just manually uploaded the `.pth` file to ensure everything worked properly.

This is the first step for us in integrating huggingface into py-feat. We will continue to refactor our codebase to streamline the integration as we continue to learn how everything works.

Thanks to everyone for creating such a great resource and being very helpful and responsive to new users!

Wauplin (approved):
Thanks @ljchang! Everything looks great to merge now :) Once deployed (expect a 2-3 days delay), all models listed in https://huggingface.co/models?other=py-feat will be recognized as `Py-Feat` library.

Regarding https://github.com/huggingface/huggingface.js/pull/879#discussion_r1741050505:
1. :+1: 
2. :+1: 
3. :+1: 
4. Good! Let us know if we can be of any help on this.
5. If you ever investigate the errors you got with `PyTorchModelHubMixin`, I'd be interested to know! You can open an issue on https://github.com/huggingface/huggingface_hub if you encounter any bug. Otherwise [this guide](https://huggingface.co/docs/huggingface_hub/guides/integrations#a-more-complex-approach-class-inheritance) is the way to go.

One major advantage of using the mixin is that it'll push a `config.json` file to the Hub when pushing a model and pull it when using `from_pretrained`. This config file is the one used to count downloads on the Hub. This is why for example [py-feat/resmasknet](https://huggingface.co/py-feat/resmasknet) has 691 downloads while [py-feat/svm_au](https://huggingface.co/py-feat/svm_au) don't have any (see _Downloads are not tracked for this model._). If you don't manage to make the mixin work, it's still possible to make the download count work for all `Py-Feat` models by adding a rule "count on config.json OR .pth file". This can be added in a follow-up PR.

Wauplin:
The failing CI is unrelated to the PR so let's merge it :)

Issue #764, opened 2024-06-19
add Snapdragon X Elite / Plus CPU scores


sitaoo:
The Snapdragon X Elite / Plus series laptops have been officially released, they come with Qualcomm Adreno GPU and Qualcomm Hexagon NPU, they are very powerful when it comes to running large language models (LLMs).

**Please add the rating to `packages/tasks/src/hardware.ts`**

Reference link:
https://www.qualcomm.com/products/mobile/snapdragon/pcs-and-tablets/snapdragon-x-elite

| Processor              | Qualcomm Oryon CPU |                 |                               |                     | Qualcomm Adreno GPU | Qualcomm Hexagon NPU | Memory          |                   |
| :--------------------- | :----------------- | :-------------- | :---------------------------- | :------------------ | :------------------ | :------------------- | :-------------- | :---------------- |
|                        | **Cores**          | **Total cache** | **Max Multithread frequency** | **Dual core boost** | **TFLOPS**          | **NPU TOPS**         | **Memory Type** | **Transfer Rate** |
| **Snapdragon X Elite** |                    |                 |                               |                     |                     |                      |                 |                   |
| X1E-00-1DE             | 12                 | 42 MB           | 3.8 GHz                       | 4.3 GHz             | 4.6                 | 45                   | LPDDR5x         | 8448 MT/s         |
| X1E-84-100             | 12                 | 42 MB           | 3.8 GHz                       | 4.2 GHz             | 4.6                 | 45                   | LPDDR5x         | 8448 MT/s         |
| X1E-80-100             | 12                 | 42 MB           | 3.4 GHz                       | 4.0 GHz             | 3.8                 | 45                   | LPDDR5x         | 8448 MT/s         |
| X1E-78-100             | 12                 | 42 MB           | 3.4 GHz                       | None                | 3.8                 | 45                   | LPDDR5x         | 8448 MT/s         |
| **Snapdragon X Plus**  |                    |                 |                               |                     |                     |                      |                 |                   |
| X1P-64-100             | 10                 | 42 MB           | 3.4 GHz                       | None                | 3.8                 | 45                   | LPDDR5x         | 8448 MT/s         |

coyotte508:
Hi @sitaoo , thanks for the reference :) 

Feel free to make a PR yourself in https://github.com/huggingface/huggingface.js/blob/main/packages/tasks/src/hardware.ts if you want this to be added faster

julien-c:
[...]

PR #881, opened 2024-09-01
Adds Snapdragon X Elite and X Plus GPU scores to hardware.ts

mrkiura:
Added:
* Snapdragon X Elite
  * X1E-00-1DE
  * X1E-84-100
  * X1E-80-100
  * X1E-78-100
  
* Snapdragon X Plus
  * X1P-64-100
 
Addresses: #764 
Source: https://www.qualcomm.com/products/mobile/snapdragon/laptops-and-tablets/snapdragon-x-elite


coyotte508:
Even if it has an integrated GPU, I think maybe it belongs in CPUs?

Also what to do about NPU score @pcuenca / @julien-c ? They have 45 TOps in their NPU, and arguably the NPU is more important than the GPU for ML stuff.

Vaibhavs10 (commented):
I'd consider it a GPU still - in more long-term ideally, we'd be able to use the device info they put and customise snippets and even recommend models for their tasks down the line.

Besides, down the line when NPUs are more mainstream it might make sense to add it as a separate add-on.

Vaibhavs10 (commented):
Hey @mrkiura - thanks a ton for your contribution, please do make sure to run the linter so that the lint test passes.

If you use VS code then you can do so with the `prettier` extension too.

mrkiura:
Thanks @coyotte508 @Vaibhavs10 

I have updated the branch.

coyotte508 (approved):

julien-c (commented):
maybe of interest to @mfuntowicz too

mrkiura:
It seems like the Test/node and (Test/ browser) checks are failing because [this URL](https://github.com/huggingface/huggingface.js/blob/main/packages/hub/src/utils/WebBlob.spec.ts#L54) in packages/hub/src/utils/WebBlob.spec.ts references a file from runway ml's account, but Runway is no longer maintaining a hugging face organization. It probably needs to be addressed in a different PR.

Cc @julien-c @coyotte508 @Vaibhavs10

coyotte508:
> It seems like the Test/node and (Test/ browser) checks are failing because [this URL](https://github.com/huggingface/huggingface.js/blob/main/packages/hub/src/utils/WebBlob.spec.ts?rgh-link-date=2024-09-03T18%3A42%3A21Z#L54) in packages/hub/src/utils/WebBlob.spec.ts references a file from runwayml's account, but Runway is no longer maintaining a HuggingFace organization. It probably needs to be addressed in a different PR.
> 
> Cc @julien-c @coyotte508 @Vaibhavs10

yes you can ignore those tests

pcuenca (approved):
I tend to agree with @coyotte508, it somehow feels more natural to classify them as CPUs, but no strong opinion.

Vaibhavs10 (approved):
Good to merge then, we can revisit the classification when we have a bit more candidates? ✅

Vaibhavs10:
Thanks a lot again for the contribution @mrkiura ❤️

PR #900, opened 2024-09-09
Add sapiens as a library

merveenoyan:


packages/tasks/src/model-libraries.ts
```diff
@@ -506,6 +506,13 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		filter: true,
 		countDownloads: `path:"cfg.json"`,
 	},
+	sapiens: {
+		prettyLabel: "sapiens",
+		repoName: "sapiens",
+		repoUrl: "https://github.com/facebookresearch/sapiens",
+		filter: false,
+		countDownloads: `path_extension:"pt"`,
```
Wauplin, line 514: From what I can see in https://huggingface.co/models?other=sapiens, all models have a `config.json` file and no `.pt` files. I think it's best to continue to count downloads using `config.json` (the default rule) in that case. 
Except if the `config.json` file is not loaded when a onnx file is used and in that case we should count downloads using the onnx downloads instead (but counting `config.json` would still be preferable if it makes sense)

Wauplin (commented):
Look good to me except the download count rule. I saw that there are already 6 models under https://huggingface.co/models?other=sapiens which is nice :+1:

packages/tasks/src/model-libraries.ts
pcuenca: The models that are intended to be covered by this PR are [these ones](https://huggingface.co/facebook/sapiens). They don't have the tag yet, and they may be split into separate repos (either per-task or per-checkpoint).

Wauplin: Oh ok, thanks for the pointer :+1:  
Let's think of a clean up before merging then (maybe contact the other 6 repos to change their tag?)

pcuenca: cc @xenova. I think the tag is auto-inferred from the configuration (they are sapiens models), but they are ONNX versions.

Wauplin: So maybe 

```suggestion
		countDownloads: `path_extension:"pt" OR path_extension:"onnx"`,
```

would be appropriate to handle both cases. If all of them are `sapiens` models, that's perfect.

NielsRogge: It seems the extension is a bit flexible:
* .pth: https://huggingface.co/facebook/sapiens-depth-0.3b/tree/main
* .pt2 instead of .pt for the bfloat16 models: https://huggingface.co/facebook/sapiens-pretrain-0.3b-bfloat16/tree/main

Wauplin: But is https://github.com/facebookresearch/sapiens able to load all of these weights? This doesn't seem very unified :confused:

merveenoyan: addressed this

packages/tasks/src/model-libraries.ts
```diff
@@ -506,6 +506,13 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		filter: true,
 		countDownloads: `path:"cfg.json"`,
 	},
+	sapiens: {
```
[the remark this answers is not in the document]
merveenoyan: I have already addressed this. I'll open PRs today, this will not take effect immediately anyway so it can be merged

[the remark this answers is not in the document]
Vaibhavs10: Since the PR is for the `Sapiens` library, it'd probably make sense to wait till the merge your Hub PRs and then merge this PR. Otherwise, this PR/ downloads won't make much sense.

(you know this collab/ authors better, I'm commenting purely from hub perspective)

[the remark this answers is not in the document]
merveenoyan: I just send them a script to merge all let's see

merveenoyan:
they've merged all PRs adding sapiens tag @Wauplin

Wauplin (approved):
Looks good! Thanks for handling it @merveenoyan :)

pcuenca (approved):

PR #907, opened 2024-09-17
Update diffusers default snippets to be complete.

Vaibhavs10:
Slack chat for ref: https://huggingface.slack.com/archives/C04PJ0H35UM/p1726563327331199?thread_ts=1725899853.249099&cid=C04PJ0H35UM

PR #909, opened 2024-09-18
Improve prompting for diffusers default snippets

apolinario:
Addressing comments left on #907, specially for LoRAs

### DIFF
--- a/packages/tasks/src/model-libraries-snippets.ts
+++ b/packages/tasks/src/model-libraries-snippets.ts
@@ -146,7 +146,7 @@ const diffusers_default = (model: ModelData) => [
 
 pipe = DiffusionPipeline.from_pretrained("${model.id}")
 
-prompt = "${escapeQuotes(get_prompt_from_diffusers_model(model)) ?? diffusersDefaultPrompt}"
+prompt = "${get_prompt_from_diffusers_model(model) ?? diffusersDefaultPrompt}"
 image = pipe(prompt).images[0]`,
 ];
 
--- a/packages/tasks/src/model-libraries-snippets.ts
+++ b/packages/tasks/src/model-libraries-snippets.ts
@@ -9,7 +9,7 @@ function nameWithoutNamespace(modelId: string): string {
 	return splitted.length === 1 ? splitted[0] : splitted[1];
 }
 
-const escapeQuotes = (str: string): string =>  JSON.stringify(str);
+const escapeQuotes = (str: string): string => JSON.stringify(str);
 
 //#region snippets
 
@@ -75,9 +75,9 @@ function get_base_diffusers_model(model: ModelData): string {
 
 function get_prompt_from_diffusers_model(model: ModelData): string | undefined {
 	const prompt = (model.widgetData?.[0] as WidgetExampleTextInput).text ?? model.cardData?.instance_prompt;
-    if(prompt){
-        return escapeQuotes(prompt);
-    }
+	if (prompt) {
+		return escapeQuotes(prompt);
+	}
 }
 
 export const bertopic = (model: ModelData): string[] => [
--- a/packages/tasks/src/model-libraries-snippets.ts
+++ b/packages/tasks/src/model-libraries-snippets.ts
@@ -165,7 +165,7 @@ const diffusers_lora = (model: ModelData) => [
 pipe = DiffusionPipeline.from_pretrained("${get_base_diffusers_model(model)}")
 pipe.load_lora_weights("${model.id}")
 
-prompt = "${escapeQuotes(get_prompt_from_diffusers_model(model)) ?? diffusersDefaultPrompt}"
+prompt = "${get_prompt_from_diffusers_model(model) ?? diffusersDefaultPrompt}"
 image = pipe(prompt).images[0]`,
 ];
 
--- a/packages/tasks/src/model-libraries-snippets.ts
+++ b/packages/tasks/src/model-libraries-snippets.ts
@@ -9,6 +9,8 @@ function nameWithoutNamespace(modelId: string): string {
 	return splitted.length === 1 ? splitted[0] : splitted[1];
 }
 
+const escapeQuotes = (str: string | undefined): string | undefined => str?.replace(/"/g, '\\"');
+
 //#region snippets
 
 export const adapters = (model: ModelData): string[] => [
@@ -141,7 +143,7 @@ const diffusers_default = (model: ModelData) => [
 
 pipe = DiffusionPipeline.from_pretrained("${model.id}")
 
-prompt = "${get_prompt_from_diffusers_model(model) ?? diffusersDefaultPrompt}"
+prompt = "${escapeQuotes(get_prompt_from_diffusers_model(model)) ?? diffusersDefaultPrompt}"
 image = pipe(prompt).images[0]`,
 ];
 
@@ -160,7 +162,7 @@ const diffusers_lora = (model: ModelData) => [
 pipe = DiffusionPipeline.from_pretrained("${get_base_diffusers_model(model)}")
 pipe.load_lora_weights("${model.id}")
 
-prompt = "${get_prompt_from_diffusers_model(model) ?? diffusersDefaultPrompt}"
+prompt = "${escapeQuotes(get_prompt_from_diffusers_model(model)) ?? diffusersDefaultPrompt}"
 image = pipe(prompt).images[0]`,
 ];
 
--- a/packages/tasks/src/model-libraries-snippets.ts
+++ b/packages/tasks/src/model-libraries-snippets.ts
@@ -9,7 +9,7 @@ function nameWithoutNamespace(modelId: string): string {
 	return splitted.length === 1 ? splitted[0] : splitted[1];
 }
 
-const escapeQuotes = (str: string): string => JSON.stringify(str);
+const escapeStringForJson = (str: string): string => JSON.stringify(str);
 
 //#region snippets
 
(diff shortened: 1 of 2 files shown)


packages/tasks/src/model-libraries-snippets.ts
```diff
@@ -70,6 +70,12 @@ function get_base_diffusers_model(model: ModelData): string {
 	return model.cardData?.base_model?.toString() ?? "fill-in-base-model";
 }
 
+function get_prompt_from_diffusers_model(model: ModelData): string {
+	return model.cardData?.widget?.[0]?.text?.toString()
```
pcuenca, line 74:
```suggestion
	return model.widgetData?.[0].text
```

I think

julien-c: yes @pcuenca is right, this is the right way to access this

packages/tasks/src/model-libraries-snippets.ts
```diff
@@ -70,6 +70,12 @@ function get_base_diffusers_model(model: ModelData): string {
...
+	    ?? model.cardData?.instance_prompt?.toString()
```
pcuenca, line 75: I don't know where to get this from.

julien-c: this one's fine imo

pcuenca: I don't understand how it works though. Is it a part of `inference.parameters`? It's not part of https://huggingface.co/docs/api-inference/tasks/text-to-image

packages/tasks/src/model-libraries-snippets.ts
```diff
@@ -70,6 +70,12 @@ function get_base_diffusers_model(model: ModelData): string {
 	return model.cardData?.base_model?.toString() ?? "fill-in-base-model";
 }
 
+function get_prompt_from_diffusers_model(model: ModelData): string {
```
mishig25, line 73:
```suggestion
function get_prompt_from_diffusers_model(model: ModelData): string | undefined {
```

since the value `model.widgetData?.[0].text` is not guranteed to be there all the time

packages/tasks/src/model-libraries-snippets.ts
```diff
@@ -134,7 +140,7 @@ const diffusers_default = (model: ModelData) => [
 
 pipe = DiffusionPipeline.from_pretrained("${model.id}")
 
-prompt = "Astronaut in a jungle, cold color palette, muted colors, detailed, 8k"
+prompt = "${get_prompt_from_diffusers_model(model)}"
```
mishig25, line 145: following the comment [here](https://github.com/huggingface/huggingface.js/pull/909/files#r1764685115)

it should be something like:
```ts
const defaultPrompt = "Astronaut in a jungle, cold color palette, muted colors, detailed, 8k";
prompt = "${get_prompt_from_diffusers_model(model) ?? defaultPrompt}"
```

saying that if example prompt does not exist in model card, it should fallback to `defaultPrompt`

packages/tasks/src/model-libraries-snippets.ts
```diff
@@ -70,6 +70,14 @@ function get_base_diffusers_model(model: ModelData): string {
 	return model.cardData?.base_model?.toString() ?? "fill-in-base-model";
 }
 
+function get_prompt_from_diffusers_model(model: ModelData): string {
+	return (
+		model.cardData?.widget?.[0]?.text?.toString() ??
+		model.cardData?.instance_prompt?.toString() ??
```
mishig25, line 76: why do we need `toString()` ? aren't `.text` & `.instance_prompt` already strings ? as defined https://github.com/huggingface/huggingface.js/pull/909/files#diff-4dbb69668b0377cf24081f874e42ce011491bed90939745260b6f65d689dd941

packages/tasks/src/model-libraries-snippets.ts
apolinario: With `widgetData` it errors as it doesn't find the `text` property 
```
│ src/model-libraries-snippets.ts(75,26): error TS2339: Property 'text' does not exist on type 'WidgetExample'.
│   Property 'text' does not exist on type 'WidgetExampleChatInput<WidgetExampleOutput>'.
```

mishig25: for now you can do this:
```ts
(model.widgetData?.[0] as WidgetExampleTextInput).text
```

because the model card is for diffusers model, we can assume that the widgetData is for `text-to-image` pipeline which is `WidgetExampleTextInput`

mishig25: and you can remove https://github.com/huggingface/huggingface.js/blob/324e720c631ab871a920c371178e3d5dfdd7613b/packages/tasks/src/model-data.ts#L110-L115 since it is redundant to https://github.com/huggingface/huggingface.js/blob/324e720c631ab871a920c371178e3d5dfdd7613b/packages/tasks/src/model-data.ts#L92

packages/tasks/src/model-libraries-snippets.ts
```diff
@@ -70,6 +71,10 @@ function get_base_diffusers_model(model: ModelData): string {
 	return model.cardData?.base_model?.toString() ?? "fill-in-base-model";
 }
 
+function get_prompt_from_diffusers_model(model: ModelData): string | undefined {
+	return (model.widgetData?.[0] as WidgetExampleTextInput).text ?? model.cardData?.instance_prompt ?? undefined;
```
mishig25, line 75:
```suggestion
	return (model.widgetData?.[0] as WidgetExampleTextInput).text ?? model.cardData?.instance_prompt ?? undefined;
```

mishig25, line 75:
```suggestion
	return (model.widgetData?.[0] as WidgetExampleTextInput).text ?? model.cardData?.instance_prompt;
```

no need for explicit `undefined`

packages/tasks/src/model-libraries-snippets.ts
apolinario: I've added `instance_prompt` to `cardData` to fix the typing error, but I'm not completely sure this will make it pull correctly from the model card https://github.com/huggingface/huggingface.js/pull/909/files#diff-4dbb69668b0377cf24081f874e42ce011491bed90939745260b6f65d689dd941R110

mishig25:
looking great!

regarding this line:
`prompt = "${get_prompt_from_diffusers_model(model) ?? diffusersDefaultPrompt}"`

I think there can be edge case where user supplied prompt has quote like `"`, which means we need to escape  `"`.
Otherwise, there is a danger of producing broken python code like:
```python
prompt = "Astronaut holding sign that says "Hello""
```

packages/tasks/src/model-libraries-snippets.ts
apolinario: wdyt @mishig25 ?

mishig25: > I've added instance_prompt to cardData
checked in moon. What you did is correct 👍

julien-c: instance_prompt is validated to be a string in moon. But no big deal to add a toString

Vaibhavs10 (approved):
Conceptually looks great to me! Thanks for taking this to the finish line! 🐐

packages/tasks/src/model-libraries-snippets.ts
```diff
@@ -8,6 +9,8 @@ function nameWithoutNamespace(modelId: string): string {
 	return splitted.length === 1 ? splitted[0] : splitted[1];
 }
 
+const escapeQuotes = (str: string | undefined): string | undefined => str?.replace(/"/g, '\\"');
```
mishig25, line 12:
```suggestion
const escapeQuotes = (str: string): string => str.replace(/"/g, '\\"');
```
I think it would be cleaner if the function signature is strictly `string` inout & `string` output

packages/tasks/src/model-libraries-snippets.ts
```diff
@@ -129,12 +136,14 @@ depth = model.infer_image(raw_img) # HxW raw depth map in numpy
...
 
+const diffusersDefaultPrompt = "Astronaut in a jungle, cold color palette, muted colors, detailed, 8k";
+
 const diffusers_default = (model: ModelData) => [
 	`from diffusers import DiffusionPipeline
 
 pipe = DiffusionPipeline.from_pretrained("${model.id}")
 
-prompt = "Astronaut in a jungle, cold color palette, muted colors, detailed, 8k"
+prompt = "${escapeQuotes(get_prompt_from_diffusers_model(model)) ?? diffusersDefaultPrompt}"
```
mishig25, line 146:
```suggestion
prompt = "${get_prompt_from_diffusers_model(model) ?? diffusersDefaultPrompt}"
```

packages/tasks/src/model-libraries-snippets.ts
```diff
@@ -153,7 +162,7 @@ const diffusers_lora = (model: ModelData) => [
 pipe = DiffusionPipeline.from_pretrained("${get_base_diffusers_model(model)}")
 pipe.load_lora_weights("${model.id}")
 
-prompt = "Astronaut in a jungle, cold color palette, muted colors, detailed, 8k"
+prompt = "${escapeQuotes(get_prompt_from_diffusers_model(model)) ?? diffusersDefaultPrompt}"
```
mishig25, line 165:
```suggestion
prompt = "${get_prompt_from_diffusers_model(model) ?? diffusersDefaultPrompt}"
```

mishig25 (approved):

Issue #149, opened 2023-04-17
Refacto `credentials` to just `accessToken: string`


coyotte508:
`credentials: Credentials` is verbose and probably unnecessary future-proofing

`accessToken: AccessToken` would make more sense

PR #918, opened 2024-09-20
💥 Simpler credentials passing around

coyotte508:
Fix https://github.com/huggingface/huggingface.js/issues/149

Should have backwards compatibilty, just marked as deprecated

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,12 +14,12 @@
 
 await createRepo({
   repo: {type: "model", name: "my-user/nlp-model"},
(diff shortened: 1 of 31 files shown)


julien-c (approved):
this does seem simpler

Pierrci (approved):
btw, we don't support oauth? 😄

coyotte508:
> btw, we don't support oauth? 😄

OAuth token are prefixed by `hf_oauth` so they work fine

PR #925, opened 2024-09-23
Added CoTracker for download stats

lvoursl:
Hey! 

Thanks for the amazing library :) 

We are hosting CoTracker model checkpoint [here](https://huggingface.co/facebook/cotracker/tree/main) and would like to see the download count. This PR follows [this one](https://github.com/huggingface/huggingface.js/pull/561/files) to enable it. 

Hope it's correct way to do it, let me know if any changes are needed or there is a better way to do so.

### DIFF
--- a/packages/tasks/src/model-libraries.ts
+++ b/packages/tasks/src/model-libraries.ts
@@ -715,6 +715,13 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		countDownloads: `path:"model_vae_fp16.pt"`,
 		snippets: snippets.threedtopia_xl,
 	},
+	cotracker: {
+		prettyLabel: "cotracker",
+		repoName: "cotracker",
+		repoUrl: "https://github.com/facebookresearch/co-tracker",
+		filter: false,
+		countDownloads: `path_extension:"pt"`,
+	},
 } satisfies Record<string, LibraryUiElement>;
 
 export type ModelLibraryKey = keyof typeof MODEL_LIBRARIES_UI_ELEMENTS;
--- a/packages/tasks/src/model-libraries.ts
+++ b/packages/tasks/src/model-libraries.ts
@@ -204,6 +204,13 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		repoUrl: "https://github.com/cartesia-ai/cartesia_mlx",
 		snippets: snippets.cartesia_mlx,
 	},
+	cotracker: {
+		prettyLabel: "CoTracker",
+		repoName: "CoTracker",
+		repoUrl: "https://github.com/facebookresearch/co-tracker",
+		filter: false,
+		countDownloads: `path_extension:"pth"`,
+	},
 	edsnlp: {
 		prettyLabel: "EDS-NLP",
 		repoName: "edsnlp",
@@ -714,13 +721,6 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		filter: false,
 		countDownloads: `path:"model_vae_fp16.pt"`,
 		snippets: snippets.threedtopia_xl,
-	},
-	cotracker: {
-		prettyLabel: "CoTracker",
-		repoName: "CoTracker",
-		repoUrl: "https://github.com/facebookresearch/co-tracker",
-		filter: false,
-		countDownloads: `path_extension:"pt"`,
 	},
 } satisfies Record<string, LibraryUiElement>;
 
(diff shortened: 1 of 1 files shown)


osanseviero (approved):
Nice! It LGTM.

packages/tasks/src/model-libraries.ts
```diff
@@ -715,6 +715,13 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		countDownloads: `path:"model_vae_fp16.pt"`,
 		snippets: snippets.threedtopia_xl,
 	},
+	cotracker: {
+		prettyLabel: "cotracker",
+		repoName: "cotracker",
```
Wauplin, line 720:
```suggestion
		prettyLabel: "CoTracker",
		repoName: "CoTracker",
```
for what I can read on https://github.com/facebookresearch/co-tracker and https://co-tracker.github.io/, `"CoTracker"` seems to be the preferred casing.

Wauplin (approved):
Looks good! Would be a good occasion to update [CoTracker's GitHub README](https://huggingface.co/docs/huggingface_hub/index) and [HF Model Card](https://huggingface.co/facebook/cotracker) to showcase how to load the model from the Hugging Face Hub. For now it's showcasing 

```
cotracker = torch.hub.load("facebookresearch/co-tracker", "cotracker2").to(device)
```

but that won't count downloads on the HF Hub :hugs: 

If that can help, here is [a guide](https://huggingface.co/docs/huggingface_hub/guides/download) on how to download files using `huggingface_hub`.

packages/tasks/src/model-libraries.ts
lvoursl: Indeed, we usually use CoTracker. 

Curious what would it affect in this case? Right now our HF model (https://huggingface.co/facebook/cotracker/tree/main) tagged with `cotracker`, if we commit this change, shall we change tag as well in our existing model and use this tag in the future releases?

lvoursl: @Wauplin ping :)

Wauplin: Oops sorry for the delay @lvoursl  :hugs:

Your model been tracked as `cotracker` is good. In the PR changes you've made:
- `cotracker` (the key of the mapping you've added) is the tag that the model should have. Example [here](https://huggingface.co/google/gemma-2b-keras/blob/main/README.md?code=true#L2) with `keras-nlp`.
- `prettyName` is how it will be displayed for users. See example with `KerasNLP`:
image
- `repoName` (usually the same as `prettyName`) is how the library is called for the ["use this model" snippet](https://huggingface.co/google/gemma-2b-keras/blob/main/README.md?code=true&library=keras-nlp#L2):
image

lvoursl: Thanks for the clarification @Wauplin, appreciate it! I've committed the changes that you proposed, let me know if there anything else has to be done to merge it.

packages/tasks/src/model-libraries.ts
```diff
@@ -715,6 +715,13 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		countDownloads: `path:"model_vae_fp16.pt"`,
 		snippets: snippets.threedtopia_xl,
 	},
+	cotracker: {
+		prettyLabel: "CoTracker",
+		repoName: "CoTracker",
+		repoUrl: "https://github.com/facebookresearch/co-tracker",
+		filter: false,
+		countDownloads: `path_extension:"pt"`,
```
Wauplin, line 723:
```suggestion
		countDownloads: `path_extension:"pth"`,
```

I just noticed the file extension is not `pt` but `pth` in https://huggingface.co/facebook/cotracker/tree/main.

packages/tasks/src/model-libraries.ts
```diff
@@ -715,6 +715,13 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		countDownloads: `path:"model_vae_fp16.pt"`,
 		snippets: snippets.threedtopia_xl,
 	},
+	cotracker: {
```
Wauplin, line 718: Could you move this entry up in the mapping to keep it consistent with the alphabetical order. Thanks!

Wauplin (commented):
Thanks for the changes @lvoursl! I noticed two things to update before merging, otherwise we should be good to go :)

lvoursl:
@Wauplin I've fixed both things that you mentioned -- looks like good to go :)

Wauplin (approved):
Thanks @lvoursl! Everything looks good.
I'll merge this PR now. Expect 2-3 days to see it shipped to production :)

PR #929, opened 2024-09-26
[`integration`] Add Model2Vec as a library on the Hub

tomaarsen:
Hello!

## Pull Request overview
* Add Model2Vec as a library on the Hub
  * Including snippet

## Links
* Repository: https://github.com/MinishLab/model2vec
* Models: https://huggingface.co/models?library=model2vec

## Details
A Sentence Transformers integration is planned. Once released, I'll add a second snippet on how to load the model with Sentence Transformers. 

- Tom Aarsen

### DIFF
--- a/packages/tasks/src/model-libraries.ts
+++ b/packages/tasks/src/model-libraries.ts
@@ -412,7 +412,6 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		repoUrl: "https://github.com/MinishLab/model2vec",
 		snippets: snippets.model2vec,
 		filter: false,
-		countDownloads: `path_extension:"safetensors"`,
 	},
 	moshi: {
 		prettyLabel: "Moshi",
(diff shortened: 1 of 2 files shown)


packages/tasks/src/model-libraries.ts
```diff
@@ -406,6 +406,14 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		filter: false,
 		countDownloads: `path:"mlc-chat-config.json"`,
 	},
+	model2vec: {
+		prettyLabel: "Model2Vec",
+		repoName: "model2vec",
+		repoUrl: "https://github.com/MinishLab/model2vec",
+		snippets: snippets.model2vec,
+		filter: false,
+		countDownloads: `path_extension:"safetensors"`,
```
osanseviero, line 415: The repos also contain a `config.json` so downloads should be computed by default. Would that work?

osanseviero (commented):
Nice :fire:

packages/tasks/src/model-libraries.ts
tomaarsen: Let me double-check whether the `config.json` is actually loaded, or if it's only pushed.

tomaarsen: Looks like it's indeed used. Let me remove this line for simplicity.

osanseviero (approved):
LGTM!

Wauplin (approved):
A clean PR as I like them! Thanks! :hugs:

pcuenca (approved):

PR #937, opened 2024-10-01
:sparkle: FEAT(local-apps): add ComfyUI LocalApp

### LINKED ISSUE #855: [feat] Add comfy to local apps
image

tildebyte:
Closes #855

RE: the discussion in :arrow_up: - to the best of of my knowledge there's no way to ask vanilla Comfy[1] to d/l a model for you; it's completely manual

Related to above: `deeplink` is almost certainly incorrect, but 1) the local type checker was very pissed about it **not** being present 2) I don't know the HF API well enough to know what to put there.

Apologies for no local verification/testing; I'm not a TS person, and it would take me longer to figure out/set up than it took me to write the code :grin: 

[1] There are probably 3rd-party nodes to do so, but it's more likely that such a thing would build a list of its own in some way


tildebyte:
Sorry for the churn. Opening a new PR.

PR #938, opened 2024-10-02
[Python snippets] map `snippetConversationalWithImage`

mishig25:
Follow up to #927

python equavalent of https://github.com/huggingface/huggingface.js/blob/1bb5b31131c6990547087d91aebda2361e91dfad/packages/tasks/src/snippets/js.ts#L188

Because of this missing line, user does not see `python` amongst the options in the[ inference snippet](https://huggingface.co/meta-llama/Llama-3.2-11B-Vision-Instruct?inference_api=true) 
image


julien-c (commented):
lgtm

PR #949, opened 2024-10-05
feat: add `node-llama-cpp` library and CLI to GGUF models

giladgd:
This PR adds a code snippet of using [`node-llama-cpp`](https://node-llama-cpp.withcat.ai) on GGUF models, and also CLI commands to:
* Chat with the model
* Estimate the model compatibility with the current hardware without downloading it

[`node-llama-cpp`](https://node-llama-cpp.withcat.ai) makes using `llama.cpp` in Node.js easy, and also provides convenient CLI commands that can be used without even installing it (assuming Node.js is installed).

I couldn't find where to add the icon for `node-llama-cpp`, so in case it's not in this repo, you can use [this icon](https://github.com/withcatai/node-llama-cpp/blob/master/assets/icon.v3.svg).


packages/tasks/src/local-apps.ts
```diff
@@ -119,6 +119,23 @@ const snippetLlamacpp = (model: ModelData, filepath?: string): LocalAppSnippet[]
...
+		{
+			title: "Chat with the model",
+			content: [
+				`npx -y node-llama-cpp chat \\`,
+				`  --model "hf:${model.id}/${filepath ?? "{{GGUF_FILE}}"}" \\`,
+				`  --prompt 'Hi there!'`,
+			].join("\n"),
+		},
+		{
+			title: "Estimate the model compatibility with your hardware",
```
Vaibhavs10, line 133: This is very cool! 🔥

packages/tasks/src/local-apps.ts
```diff
@@ -200,6 +217,13 @@ export const LOCAL_APPS = {
 		displayOnModelPage: isLlamaCppGgufModel,
 		snippet: snippetLlamacpp,
 	},
+	"node-llama-cpp-cli": {
+		prettyLabel: "node-llama-cpp (CLI)",
+		docsUrl: "https://node-llama-cpp.withcat.ai",
+		mainTask: "text-generation",
```
Vaibhavs10, line 223: I think we can use the `conversational` tage here: https://huggingface.co/models?other=conversational

cc: @mishig25 WDYT?

packages/tasks/src/model-libraries.ts
```diff
@@ -448,6 +448,13 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		filter: true,
 		countDownloads: `path_extension:"nemo" OR path:"model_config.yaml"`,
 	},
+	"node-llama-cpp": {
```
Vaibhavs10, line 451: Are there any examples of where this is used? I don't see any tagged repos here: https://huggingface.co/models?other=node-llama-cpp

Note: Whichever repo has the tag `node-llama-cpp` would have the snippet you defined.

Vaibhavs10 (commented):
Thanks for the PR @giladgd - It's already looking very good.

re: LocalApps can you send the SVG of the logo for `node-llama-cpp` as well.

More meta comment - I think this would fit more generally for LocalApps atm, but let's wait to see what others think as well: cc: @julien-c

packages/tasks/src/local-apps.ts
osanseviero: `conversational` is not a task, it's a sub-task. I would expect this to be consistent with llama.cpp snippet, which uses `text-generation`

packages/tasks/src/local-apps.ts
```diff
@@ -119,6 +119,23 @@ const snippetLlamacpp = (model: ModelData, filepath?: string): LocalAppSnippet[]
...
 };
 
+const snippetNodeLlamaCppCli = (model: ModelData, filepath?: string): LocalAppSnippet[] => {
+	return [
+		{
+			title: "Chat with the model",
+			content: [
+				`npx -y node-llama-cpp chat \\`,
+				`  --model "hf:${model.id}/${filepath ?? "{{GGUF_FILE}}"}" \\`,
+				`  --prompt 'Hi there!'`,
```
osanseviero, line 129: Does this take care of chat templates automatically?

packages/tasks/src/model-libraries.ts
osanseviero: We should only have `node-llama-cpp` as a local app option, not a library, for consistency with the rest

osanseviero (commented):
Very nice :fire:

packages/tasks/src/model-libraries.ts
Vaibhavs10: Yes! read this: https://github.com/huggingface/huggingface.js/pull/949#pullrequestreview-2350820812

julien-c: agree

packages/tasks/src/local-apps.ts
giladgd: Yes

giladgd:
@Vaibhavs10 You can use [this SVG icon](https://github.com/withcatai/node-llama-cpp/blob/master/assets/icon.v3.svg) in places that need a small icon, and [this SVG logo](https://github.com/withcatai/node-llama-cpp/blob/master/assets/icon.svg) for places that need a medium-sized logo (note that this logo's colors change depending on the [`color-scheme`](https://developer.mozilla.org/en-US/docs/Web/CSS/color-scheme) CSS property used on the image)

packages/tasks/src/model-libraries.ts
julien-c: IMO let's first ship in local apps and we can always revisit later

giladgd: Alright. I removed the library code snippet

Vaibhavs10 (approved):
Thanks a lot for iterating here @giladgd - really appreciate it! We'll have a new release of LocalApps soon. There's an update required from our end (on our internal codebase) - this PR will be merged along with that!

cc: @enzostvs @krampstudio for vis 🤗

pcuenca (approved):

giladgd:
@Vaibhavs10 Do you know when it will show up on GGUF models?
I saw that you added Ollama yesterday, but `node-llama-cpp` is not there.

giladgd:
@julien-c I don't see it on the drop-down menu on my end (also when I'm not logged in).
I've tested on [this model](https://huggingface.co/mradermacher/Meta-Llama-3.1-8B-Instruct-GGUF):
image

giladgd:
Is it possible to show it by default alongside `llama.cpp` and Ollama?
`node-llama-cpp` makes it easy to run `llama.cpp` models locally without installing anything using a single `npx` command, so I think having it being shown by default and also for unauthenticated users would be very helpful.

PR #950, opened 2024-10-06
Rm `LocalAppSnippet.setup`

mishig25:
Follow up to https://github.com/huggingface/huggingface.js/issues/778 and https://github.com/huggingface/huggingface.js/pull/942

After https://github.com/huggingface/huggingface.js/pull/942, I've realized that a better absteraction/generalisation is just to make `LocalAppSnippet.content: string | string[]` & remove `LocalAppSnippet.setup` altogether. Instead put what was in `LocalAppSnippet.setup` as the first item in `LocalAppSnippet.content` string array

### DIFF
--- a/packages/tasks/src/local-apps.ts
+++ b/packages/tasks/src/local-apps.ts
@@ -7,10 +7,6 @@ export interface LocalAppSnippet {
 	 */
 	title: string;
 	/**
-	 * Optional setup guide
-	 */
-	setup?: string;
-	/**
 	 * Content (or command) to be run
 	 */
 	content: string | string[];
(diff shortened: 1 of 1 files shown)


Vaibhavs10 (commented):
No strong feelings but I do have a preference for explicit declerations over implicit.

Could we maybe add a comment in code here/ somewhere that the first line corresponds to setup? - We get a few contributors for LocalApps now, it'd be good to make it easier for them to understand.

ngxson:
In fact, my initial idea was to have the ability to hide the `setup` section if user already installed the package. It will be useful if the setup is multi-steps, so user can optionally hide it to gain some space.

Also agree with @Vaibhavs10 that explicitly having `setup` is more intuitive for contributors.

julien-c:
yes let's keep it imo

PR #954, opened 2024-10-07
[Local App Snippet] support non conversational LLMs

mishig25:
## Description

Most GGUF files on the hub are insutrct/conversational. However, not all of them. Previously, local app snippets assumed that all GGUFs are insutrct/conversational.

### vLLM

https://huggingface.co/meta-llama/Llama-3.2-3B?local-app=vllm

```json
mishig@machine:~$ curl -X POST "http://localhost:8000/v1/completions" \
        -H "Content-Type: application/json" \
        --data '{
                "model": "meta-llama/Llama-3.2-3B",
                "prompt": "Once upon a time",
                "max_tokens": 150,
                "temperature": 0.5
        }'

{"id":"cmpl-157aad50ba6d45a5a7e2641a3c8157dd","object":"text_completion","created":1728293162,"model":"meta-llama/Llama-3.2-3B","choices":[{"index":0,"text":" there was a man who was very generous and kind to everyone. He was a good man and a good person. One day he was walking down the street and he saw a man who was very poor and starving. The man was so hungry that he was crying and shaking. The man was so hungry that he was crying and shaking. The man was so hungry that he was crying and shaking. The man was so hungry that he was crying and shaking. The man was so hungry that he was crying and shaking. The man was so hungry that he was crying and shaking. The man was so hungry that he was crying and shaking. The man was so hungry that he was crying and shaking. The man was so hungry that he was crying and shaking","logprobs":null,"finish_reason":"length","stop_reason":null,"prompt_logprobs":null}],"usage":{"prompt_tokens":5,"total_tokens":155,"completion_tokens":150}}
```

### llama.cpp

https://huggingface.co/mlabonne/gemma-2b-GGUF?local-app=llama.cpp

```
llama-cli \
  --hf-repo "mlabonne/gemma-2b-GGUF" \
  --hf-file gemma-2b.Q2_K.gguf \
  -p "Once upon a time "
```

### llama-cpp-python

```python
from llama_cpp import Llama

llm = Llama.from_pretrained(
        repo_id="mlabonne/gemma-2b-GGUF",
        filename="gemma-2b.Q2_K.gguf",
)

output = llm(
        "Once upon a time ",
        max_tokens=512,
        echo=True
)

print(output)
```

PR #962, opened 2024-10-09
Add Text Generation Inference as a Local App 🧠

Vaibhavs10:


Vaibhavs10:
Particularly interested in @mishig25 & @Narsil's review for the Run commands! 🤗

packages/tasks/src/local-apps.ts
```diff
@@ -77,6 +77,9 @@ function isMarlinModel(model: ModelData): boolean {
 function isTransformersModel(model: ModelData): boolean {
 	return model.tags.includes("transformers");
 }
+function isTgiModel(model: ModelData): boolean {
+	return model.tags.includes("text-generation-inference");
```
Vaibhavs10, line 81: Note: this will be much more reliable once this PR: https://github.com/huggingface-internal/moon-landing/pull/11356 (internal) is merged

packages/tasks/src/local-apps.ts
```diff
@@ -180,6 +183,33 @@ const snippetVllm = (model: ModelData): LocalAppSnippet[] => {
...
 };
+const snippetTgi = (model: ModelData): LocalAppSnippet[] => {
+	const runCommand = [
+		"# Call the server using curl:",
+		`curl -X POST "http://localhost:8000/v1/chat/completions" \\`,
+		`	-H "Content-Type: application/json" \\`,
+		`	--data '{`,
+		`		"model": "${model.id}",`,
+		`		"messages": [`,
+		`			{"role": "user", "content": "What is the capital of France?"}`,
```
Vaibhavs10, line 194: This assumes that all models are instruct, we would need to update this once Mishig's PR is merged: https://github.com/huggingface/huggingface.js/pull/954

Or happy to merge this now @mishig25 - in case you want to take care of this in your PR - whatever you think is the best.

packages/tasks/src/local-apps.ts
```diff
@@ -214,6 +244,15 @@ export const LOCAL_APPS = {
 			(model.pipeline_tag === "text-generation" || model.pipeline_tag === "image-text-to-text"),
 		snippet: snippetVllm,
 	},
+	tgi: {
+		prettyLabel: "TGI",
+		docsUrl: "https://huggingface.co/docs/text-generation-inference/",
+		mainTask: "text-generation",
+		displayOnModelPage: (model: ModelData) =>
+			(isAwqModel(model) || isGptqModel(model) || isMarlinModel(model) || isTransformersModel(model)) &&
+			isTgiModel(model),
```
osanseviero, line 253: Should this only be in the case `isTgi`? Otherwise it seems we have two different heuristic sets for deciding whether a model is TGI-compatible, one for the tag and one for the display

Vaibhavs10: Good point, however, we can load all AWQ + GPTQ models in TGI as well.

Vaibhavs10: Let me see if the tag already covers these use-cases.

Vaibhavs10: fixed in https://github.com/huggingface/huggingface.js/pull/962/commits/933d60bd5a3a6c7b342c8934e5edadf5e381ac00

packages/tasks/src/local-apps.ts
```diff
@@ -180,6 +183,33 @@ const snippetVllm = (model: ModelData): LocalAppSnippet[] => {
...
+		`		"messages": [`,
+		`			{"role": "user", "content": "What is the capital of France?"}`,
+		`		]`,
+		`	}'`,
+	];
+	return [
+		{
+			title: "Use Docker images",
+			setup: [
+				"# Deploy with docker on Linux:",
```
julien-c, line 202: will there be one snippet section per OS?

Vaibhavs10: I was thinking of just having the linux one here (same as vllm)

packages/tasks/src/local-apps.ts
```diff
@@ -180,6 +183,33 @@ const snippetVllm = (model: ModelData): LocalAppSnippet[] => {
...
+				`docker run --gpus all \\`,
+				`	-v ~/.cache/huggingface:/root/.cache/huggingface \\`,
+				` 	-e HF_TOKEN=<secret>" \\`,
+				`	ghcr.io/huggingface/text-generation-inference:2.3.1 \\`,
```
Vaibhavs10, line 206:
```suggestion
				`	ghcr.io/huggingface/text-generation-inference:latest \\`,
```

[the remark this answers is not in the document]
Vaibhavs10: Fixed: https://github.com/huggingface/huggingface.js/pull/962/commits/66745b746c0b5e112f83c1f6b840a2bcac199d95

packages/tasks/src/local-apps.ts
julien-c: ah yes but that's just the guest OS, not the host, right?

Vaibhavs10: Oui! but, we use `http://localhost:8000/v1/chat/completions` in the usage snippet so it'll make it a bit complicated to have a different host.

packages/tasks/src/local-apps.ts
```diff
@@ -214,6 +244,13 @@ export const LOCAL_APPS = {
 			(model.pipeline_tag === "text-generation" || model.pipeline_tag === "image-text-to-text"),
 		snippet: snippetVllm,
 	},
+	tgi: {
+		prettyLabel: "TGI",
+		docsUrl: "https://huggingface.co/docs/text-generation-inference/",
+		mainTask: "text-generation",
+		displayOnModelPage: (model: ModelData) => isTgiModel(model),
```
osanseviero, line 251: I think
```suggestion
		displayOnModelPage: isTgiModel,
```

osanseviero (commented):
Looking good! If you have a screenshot of how this would look that would be great

packages/tasks/src/local-apps.ts
```diff
@@ -180,6 +183,33 @@ const snippetVllm = (model: ModelData): LocalAppSnippet[] => {
...
+		`			{"role": "user", "content": "What is the capital of France?"}`,
+		`		]`,
+		`	}'`,
+	];
+	return [
+		{
+			title: "Use Docker images",
+			setup: [
+				"# Deploy with docker on Linux:",
+				`docker run --gpus all \\`,
```
OlivierDehaene, line 203: Is it normal that we do not expose a port here?

Vaibhavs10: Thanks for the review - fixed: https://github.com/huggingface/huggingface.js/pull/962/commits/f74e4ba73d010bb3a589f2e4f7b8a03bd053977e

packages/tasks/src/local-apps.ts
Vaibhavs10: Fixed: https://github.com/huggingface/huggingface.js/pull/962/commits/f74e4ba73d010bb3a589f2e4f7b8a03bd053977e

Vaibhavs10 (commented):
The CI errors look unrelated to the PR.

Re: SVG for the LocalApp - what should we use? the same as the one used for TGI tags on the hub?

packages/tasks/src/local-apps.ts
```diff
@@ -180,6 +183,34 @@ const snippetVllm = (model: ModelData): LocalAppSnippet[] => {
...
+				`	-v ~/.cache/huggingface:/root/.cache/huggingface \\`,
+				` 	-e HF_TOKEN=<secret>" \\`,
```
pcuenca, line 205:
```suggestion
				` 	-e HF_TOKEN="<secret>" \\`,
```

packages/tasks/src/local-apps.ts
```diff
@@ -180,6 +183,34 @@ const snippetVllm = (model: ModelData): LocalAppSnippet[] => {
...
+				`	-p 8000:80 \\`,
+				`	ghcr.io/huggingface/text-generation-inference:latest \\`,
```
pcuenca, line 207: I think @Narsil is not very fond of encouraging the use of `latest` because it makes issues harder to track. But yes, using `2.3.1` would quickly become obsolete.

Vaibhavs10: Yeah! not sure what's the best way here

julien-c: `latest` is not the latest tagged release?

Vaibhavs10: But let's use latest for now (same as vllm)

pcuenca (approved):
Looks good to me!

PR #963, opened 2024-10-09
lmstudio deeplink for MLX models

yagil:
(Not yet tested)

LM Studio 0.3.4 and newer supports MLX - [reference](https://lmstudio.ai/blog/lmstudio-v0.3.4)

### DIFF
--- a/packages/tasks/src/local-apps.ts
+++ b/packages/tasks/src/local-apps.ts
@@ -82,6 +82,10 @@ function isLlamaCppGgufModel(model: ModelData) {
 	return !!model.gguf?.context_length;
 }
 
+function isMlxModel(model: ModelData) {
+	return model.tags.includes("mlx");
+}
+
 const snippetLlamacpp = (model: ModelData, filepath?: string): LocalAppSnippet[] => {
 	const command = (binary: string) =>
 		[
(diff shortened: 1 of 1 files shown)


packages/tasks/src/local-apps.ts
```diff
@@ -218,7 +222,7 @@ export const LOCAL_APPS = {
 		prettyLabel: "LM Studio",
 		docsUrl: "https://lmstudio.ai",
 		mainTask: "text-generation",
-		displayOnModelPage: isLlamaCppGgufModel,
+		displayOnModelPage: isLlamaCppGgufModel || isMlxModel,
```
julien-c, line 225:
```suggestion
		displayOnModelPage: (model) => isLlamaCppGgufModel(model) || isMlxModel(model),
```

untested too :)

yagil: that makes more sense 😅

Vaibhavs10 (commented):
Niceee! Looking forward to testing this 😄

packages/tasks/src/local-apps.ts
```diff
@@ -82,6 +82,10 @@ function isLlamaCppGgufModel(model: ModelData) {
 	return !!model.gguf?.context_length;
 }
 
+function isMlxModel(model: ModelData) {
+	return model.tags.includes("mlx");
```
julien-c, line 86:
```suggestion
	return model.library_name === "mlx";
```

or maybe like this @Vaibhavs10 if we want to be pedantic about it

julien-c (approved):

packages/tasks/src/local-apps.ts
julien-c: actually nvm because https://huggingface.co/mlx-community/Llama-3.2-3B-Instruct-4bit for instance is not library=mlx

julien-c:
CI failure is unrelated

Vaibhavs10 (approved):
Just tested with this: `lmstudio://open_from_hf?model=mlx-community/Llama-3.2-1B-Instruct-4bit`

Works perfectly! 🔥

packages/tasks/src/local-apps.ts
Vaibhavs10: Yes! Just `mlx` tag would be more reliable 👍

PR #973, opened 2024-10-18
[Tasks] Fix widget model for image-text-to-text task

hanouticelina:
This PR updates the widget model for `image-text-to-text` task to put `meta-llama/Llama-3.2-11B-Vision-Instruct` instead of `microsoft/kosmos-2-patch14-224` which is an `image-to-text` model.

Also, I'm not sure but this will maybe also fix the widget displayed in the `image-text-to-text` task [page](https://huggingface.co/tasks/image-text-to-text).

Screenshot 2024-10-18 at 15 17 21


pcuenca (commented):
Is it a problem that this model can't be used outside the EU? I think it's fine for serving, just asking out of caution.

Looks good to me otherwise, cc @merveenoyan

hanouticelina:
I think it won't be an issue for serving as you said since it's running through InferenceAPI

Wauplin (approved):
Looks like a sensible change to me as well :+1:

hanouticelina:
failing tests are unrelated so i guess we can merge this one

PR #974, opened 2024-10-20
Update hardware.ts

m626zNq:
Add more GPUs to the list, all fp16 performance.

### DIFF
--- a/packages/tasks/src/hardware.ts
+++ b/packages/tasks/src/hardware.ts
@@ -88,6 +88,10 @@ export const SKUS = {
 				tflops: 31.24,
 				memory: [24],
 			},
+			A2: {
+				tflops: 4.531, // source: https://www.techpowerup.com/gpu-specs/a2.c3848
+				memory: [16],
+			},
 			"RTX 4090": {
 				tflops: 82.58,
 				memory: [24],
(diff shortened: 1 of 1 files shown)


packages/tasks/src/hardware.ts
```diff
@@ -84,6 +84,10 @@ export const SKUS = {
 				tflops: 37.42,
 				memory: [48],
 			},
+			A16: { 
+				tflops: 17.972, // source: https://www.techpowerup.com/gpu-specs/a16-pcie.c3794
+				memory: [64], // this one you do not have to add, as its 4x.
```
pcuenca, line 89: Yeah, it's a bit confusing as this is a bundle of 4 GPUs in a single card. I'm not sure if we should report the per-GPU specs instead, as the use cases of 4x16 GB are different to 1x64 GB cards.

m626zNq: Alright I will just remove it since it will be inaccurate anyways, since its basically 1 gpu, but 4 parts? I am always confused on those.

pcuenca (approved):
Thank you! 🙌

packages/tasks/src/hardware.ts
pcuenca: Yes, another idea would be to use the per-unit specs: 4.49 tflops and 16 GB of RAM, but that could be somewhat confusing as well.

m626zNq:
> I see the CI failing in other PRs as well, merging.

That always fails tbh, thanks for merging to main!

PR #976, opened 2024-10-21
Update conversational inference API snippets

mishig25:
### Description

This PR updates inference snippet generating functions signatures. Before, the functions were generating `string`. Now, the function will generate `InferenceSnippet  | InferenceSnippet []`.

https://github.com/huggingface/huggingface.js/blob/5bc694b9e7b845d25743e6db30daef60f85883a7/packages/tasks/src/snippets/types.ts#L14-L17

Why do we need to generate `InferenceSnippet []`, not just `InferenceSnippet`?
Because for a given langauge (let's say), we wanna show multiple clients options (`huggingface_hub`, `openai`). (see the attached video below).

Also, this PR improves the conversational snippet greatly.

### Screen recording

https://github.com/user-attachments/assets/6bc982c5-a855-4879-b9be-ca023d2b59fa

### DIFF
--- a/packages/tasks/src/snippets/common.ts
+++ b/packages/tasks/src/snippets/common.ts
@@ -56,7 +56,7 @@ export function stringifyGenerationConfig(
 	return (
 		opts.start +
 		Object.entries(config)
-			.map(([key, val]) => `${quote}${key}${quote}${opts.attributeValueConnector}${quote}${val}${quote}`)
+			.map(([key, val]) => `${quote}${key}${quote}${opts.attributeValueConnector}${val}`)
 			.join(opts.sep) +
 		opts.end
 	);
--- a/packages/tasks/src/snippets/common.ts
+++ b/packages/tasks/src/snippets/common.ts
@@ -1,0 +1,63 @@
+import type { ChatCompletionInputMessage, GenerationParameters } from "../tasks";
+
+export interface StringifyMessagesOptions {
+	sep: string;
+	start: string;
+	end: string;
+	attributeKeyQuotes?: boolean;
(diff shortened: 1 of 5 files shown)


mishig25:
ready for review

Wauplin (approved):
Thanks @mishig25! Results looks amazing and will definitely settle the debate over on which snippets to display :smile: As mentioned on the moon-landing PR, I would tend to always return `InferenceSnippet []` to avoid confusions but if you think otherwise, I'm really fine keeping as it is now. 

Once this is merged, I can take care of adding InferenceClient snippets for many tasks in Python (similar to https://github.com/huggingface/huggingface.js/pull/971 but with your new structure). 

Note: this change will break the script that generates the Inference API docs ([here](https://github.com/huggingface/hub-docs/blob/main/scripts/api-inference/scripts/generate.ts)). I'll open a PR once this is merged/deployed.

julien-c:
are the selected tab + sub-tab sticky? (not sure if they should be, cc @gary149)

gary149:
> are the selected tab + sub-tab sticky? (not sure if they should be, cc @gary149)

Currently no (I think it's fine)

gary149 (commented):
Nice! I didn't try everything but I see an issue with some snippets like 

http://localhost:5564/black-forest-labs/FLUX.1-dev?inference_api=true

image

mishig25:
@gary149 https://github.com/huggingface/huggingface.js/pull/976/commits/8399207203df72fc86ae23a35110608c855f71f6 should fix the issues you've mentioned

if you are testing on moon, make sure to build hfjs again by
```
cd packages/tasks
pnpm i
pnpm build
```

image

PR #988, opened 2024-10-27
Jinja - support `rejectattr`

giladgd:
I've added support for `rejectattr` in Jinja since I encountered some GGUF models that include a chat template that uses it, like [this one](https://huggingface.co/mradermacher/Mistral-Nemo-Instruct-2407-GGUF) for example.


xenova (changes requested):
Thanks for the PR! 🎉 

Since `rejectattr` is very similar to `selectattr`, maybe we can remove code redundancy by reusing the logic?

Something like:
```js
case "selectattr": 
case "rejectattr": 
{
const select = filterName === "selectattr";

// ... same logic

// Filter the array using the test function
const filtered = (operand.value as ObjectValue[]).filter((item) => {
    const a = item.value.get(attr.value);
    const result = a ? testFunction(a, value) : false;
    return select ? result : !result;
});

}
```

the following should do the same (less readable though)
```js
const filtered = (operand.value as ObjectValue[]).filter(item => {
    const a = item.value.get(attr.value);
    return select === Boolean(a && testFunction(a, value));
});
```

giladgd:
I've made the changes, and the tests seem to pass

xenova (approved):
A perfect PR! Thanks a ton! ✅ 

Will put out a new release shortly too (just needing to finish one more thing)

julien-c:
thanks for the contrib, @giladgd!

PR #997, opened 2024-10-30
Fix formatting & linting

xenova:
For some reason, CI was green before merging #995, but missed the formatting and linting step.

This PR should fix it 👍


xenova:
Looks to be same as https://github.com/huggingface/huggingface.js/pull/990.

PR #999, opened 2024-10-30
feat(hub): adding revision property to modelInfo, datasetInfo, spaceInfo

axel7083:
When getting modelInfo, datasetInfo, spaceInfo, we might need to get the details for a specific revision.

This PR adds an optional `revision` property to those methods

### DIFF
--- a/packages/hub/src/lib/dataset-info.ts
+++ b/packages/hub/src/lib/dataset-info.ts
@@ -14,6 +14,10 @@ export async function datasetInfo<
 		hubUrl?: string;
 		additionalFields?: T[];
 		/**
+		 * An optional Git revision id which can be a branch name, a tag, or a commit hash.
+		 */
+		revision?: string;
+		/**
 		 * Custom fetch function to use instead of the default one, for example to use a proxy or edit headers.
 		 */
 		fetch?: typeof fetch;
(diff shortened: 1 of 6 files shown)


packages/hub/src/lib/dataset-info.ts
```diff
@@ -27,7 +31,7 @@ export async function datasetInfo<
 	]).toString();
 
 	const response = await (params.fetch || fetch)(
-		`${params?.hubUrl || HUB_URL}/api/datasets/${params.name}?${search.toString()}`,
+		`${params?.hubUrl || HUB_URL}/api/datasets/${params.name}${params.revision ? `/revision/${params.revision}` : ''}?${search.toString()}`,
```
coyotte508, line 34: Alternative:

```suggestion
		`${params?.hubUrl || HUB_URL}/api/datasets/${params.name}/revision/${encodeURLComponent(params.revision ?? "HEAD")}?${search.toString()}`,
```

coyotte508 (commented):
Looks good!

My only change request is to use `encodeURLComponent` on the revision, eg if it's a branch name with a `/` in it

packages/hub/src/lib/dataset-info.ts
axel7083: I applied the change! Thanks for the solution

coyotte508 (approved):

coyotte508:
Thanks!

PR #1004, opened 2024-10-30
💥 Switch gguf & tasks interdependency

coyotte508:
cc @ngxson too (you can add yourself to CODEOWNERS for gguf maybe :) )

Move the functions needed by `@huggingface/tasks` directly into `@huggingface/tasks`.

Now `@hugginface/gguf` depends on `@huggingface/tasks` and not the other way around.

The move is minimal, more could be moved. (Or a fancy solution with shared folders, but fancy solutions can become complicated esp with publishing source code along with dist. Or we could just duplicate the code, and maybe add CI check to ensure consistency)


julien-c (approved):
ok for me (or... merge the two packages)

ngxson (approved):
LGTM! (sorry for the slow response)

PR #1009, opened 2024-11-01
add @ngxson to codeowners for tasks&gguf

julien-c:


ngxson (approved):
LGTM, thanks!

Btw would mind to fix the `CODEOWNERS file contains errors` too?

image

julien-c:
well none of them are at HF but i think it's fine, cc @coyotte508

PR #1017, opened 2024-11-05
Update pipelines.ts to add Optical Music Recognition as subtask

JuanCarlosMartinezSevilla:
As I plan to publish a dataset specific for Optical Music Recognition, I would like to have this subtask to tag properly the dataset.

### DIFF
--- a/packages/tasks/src/pipelines.ts
+++ b/packages/tasks/src/pipelines.ts
@@ -426,6 +426,10 @@ export const PIPELINE_DATA = {
 				type: "image-captioning",
 				name: "Image Captioning",
 			},
+			{
+				type: "optical-music-recognition",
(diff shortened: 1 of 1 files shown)


merveenoyan:
@JuanCarlosMartinezSevilla this task is way too specific, can you explain what is the input and what is the output? at HF our tasks are often shaped according to i/o

JuanCarlosMartinezSevilla:
For this task @merveenoyan, the input is a music score image (like a document page) and the output is the sequence of tokens that represent the music symbols. This tasks is as Optical Character Recognition or Handwritten Text Recognition for text but with music documents.

merveenoyan:
yes that still falls under the image-to-text then, we can add the pipeline tag to be that and add this task under tags: in the model card

PR #1018, opened 2024-11-08
Track oasis .safetensors instead of .pt

julian-q:
Hello! We recently updated our model (https://huggingface.co/Etched/oasis-500m/tree/main) to use the safetensors format instead of pickle. We want to make the .safetensors the default and remove the .pt, and so this PR allows download counting on the .safetensors instead.


Wauplin (approved):
That's good news!

Issue #1023, opened 2024-11-13
bug(@huggingface/hub): `fileDownloadInfo` return an etag for LFS file which seems weird


axel7083:
I was experimenting with the `fileDownloadInfo` method from the `@huggingface/hub` package, trying to add caching capabilities (related https://github.com/huggingface/huggingface.js/pull/908).

https://github.com/huggingface/huggingface.js/blob/400ea899491d493f3e5d1b202ce1a765f7164695/packages/hub/src/lib/file-download-info.ts#L18

I was comparing the behaviour of the python library [github.com/huggingface/huggingface_hub](https://github.com/huggingface/huggingface_hub), as they have a [get_hf_file_metadata](https://huggingface.co/docs/huggingface_hub/en/package_reference/file_download#huggingface_hub.get_hf_file_metadata) function doing something similar (might not be the same goal, but they have some clear similarities).

To explain my problem, I have some doubt with the way that the js function is getting the fileDownloadInfo, and specifically the etag. Let's compare.

## JS vs PY request

First difference, the JS function is using the `GET` method for the request.

https://github.com/huggingface/huggingface.js/blob/400ea899491d493f3e5d1b202ce1a765f7164695/packages/hub/src/lib/file-download-info.ts#L50-L58

In comparison, the python library is using the `HEAD` method for the same url. See ref on [huggingface/huggingface_hub#src/huggingface_hub/file_download.py#L1297](https://github.com/huggingface/huggingface_hub/blob/c547c839dbbe0163e3ca422d017daad7c7f9361f/src/huggingface_hub/file_download.py#L1297)

We can note the following difference between the JS and PY requests

| options | JS | PY |
| --- | --- | --- |
| follow redirect | :heavy_check_mark: | :red_square: |
| method | `GET` | `HEAD` |
| headers | `Range: "bytes=0-0",` | `Accept-Encoding: "identity"` |

## Etag value comparison 

What are the difference of the requests, let's take the same file and compare the output of `fileDownloadInfo` (JS) and `get_hf_file_metadata` (PY)

```typescript
const info = await fileDownloadInfo({
	repo: {
		name: "bert-base-uncased",
		type: "model",
	},
	path: "tf_model.h5",
	revision: "dd4bc8b21efa05ec961e3efc4ee5e3832a3679c7",
});
assert.strictEqual(info?.etag, '"41a0e56472bad33498744818c8b1ef2c-64"');
```

```py
[...]

PR #1024, opened 2024-11-13
fix: make `fileDownloadInfo` use the repo `etag` and not the CDN one

axel7083:
Full explanation on https://github.com/huggingface/huggingface.js/issues/1023

> I opened the pull request, because I had some time to look at it, but I would understand that the `fileDownloadInfo` do not have the same result expectation than `get_hf_file_metadata`. Maybe we are expecting to have the CDN etag when using `fileDownloadInfo`, but then should a dedicated method `getFileMetadata` be created ?

## Related issues

Fixes https://github.com/huggingface/huggingface.js/issues/1023

## Testing

- [x] unit tests has been added

**Manually**

You can try to use the `fileDownloadInfo` for a specific file (LFS), and compare it with the blobs created by the python cli tool.

### DIFF
--- a/packages/hub/src/lib/file-download-info.ts
+++ b/packages/hub/src/lib/file-download-info.ts
@@ -4,14 +4,58 @@ import type { CredentialsParams, RepoDesignation } from "../types/public";
 import { checkCredentials } from "../utils/checkCredentials";
 import { toRepoId } from "../utils/toRepoId";
 
+const HUGGINGFACE_HEADER_X_REPO_COMMIT = "X-Repo-Commit"
+const HUGGINGFACE_HEADER_X_LINKED_ETAG = "X-Linked-Etag"
+const HUGGINGFACE_HEADER_X_LINKED_SIZE = "X-Linked-Size"
+
 export interface FileDownloadInfoOutput {
 	size: number;
 	etag: string;
+	commitHash: string | null;
 	/**
 	 * In case of LFS file, link to download directly from cloud provider
 	 */
 	downloadLink: string | null;
 }
+
+/**
+ * Useful when we want to follow a redirection to a renamed repository without following redirection to a CDN.
+ * If a Location header is `/hello` we should follow the relative direct
+ * However we may have full url redirect, on the same origin, we need to properly compare the origin then.
+ * @param params
+ */
+async function followSameOriginRedirect(params: {
+	url: string,
+	method: string,
+	headers: Record<string, string>,
+	/**
+	 * Custom fetch function to use instead of the default one, for example to use a proxy or edit headers.
+	 */
+	fetch?: typeof fetch;
+}): Promise<Response> {
+	const resp = await (params.fetch ?? fetch)(params.url, {
+		method: params.method,
+		headers: params.headers,
+		// prevent automatic redirect
+		redirect: 'manual',
+	});
+
+	const location: string | null = resp.headers.get('Location');
+	if(!location) return resp;
+
+	// new URL('http://foo/bar', 'http://example.com/hello').href == http://foo/bar
+	// new URL('/bar', 'http://example.com/hello').href == http://example.com/bar
+	const nURL = new URL(location, params.url);
+	// ensure origin are matching
+	if(new URL(params.url).origin !== nURL.origin)
+		return resp;
+
+	return followSameOriginRedirect({
+		...params,
+		url: nURL.href,
+	});
+}
+
 /**
  * @returns null when the file doesn't exist
  */
@@ -47,13 +91,16 @@ export async function fileDownloadInfo(
 		}/${encodeURIComponent(params.revision ?? "main")}/${params.path}` +
 		(params.noContentDisposition ? "?noContentDisposition=1" : "");
 
-	const resp = await (params.fetch ?? fetch)(url, {
-		method: "GET",
+	//
+	const resp = await followSameOriginRedirect({
+		url: url,
+		method: "HEAD",
 		headers: {
 			...(params.credentials && {
 				Authorization: `Bearer ${accessToken}`,
+				// prevent any compression => we want to know the real size of the file
+				'Accept-Encoding': 'identity',
 			}),
-			Range: "bytes=0-0",
 		},
 	});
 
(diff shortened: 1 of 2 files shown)


packages/hub/src/lib/file-download-info.ts
```diff
@@ -4,14 +4,58 @@ import type { CredentialsParams, RepoDesignation } from "../types/public";
...
+	/**
+	 * Custom fetch function to use instead of the default one, for example to use a proxy or edit headers.
+	 */
+	fetch?: typeof fetch;
+}): Promise<Response> {
+	const resp = await (params.fetch ?? fetch)(params.url, {
+		method: params.method,
+		headers: params.headers,
+		// prevent automatic redirect
+		redirect: 'manual',
```
coyotte508, line 40: This doesn't work in the browser

coyotte508 (commented):
There's an endpoint `/api/:repoType(models|spaces|datasets)/:namespace?/:repo/paths-info/:rev` called with POST

With body `{paths: string[], expand: boolean}` (expand adds security status & last commit info)

I think it would be more helpful here. Not sure if we want to keep the same `fileDownloadInfo` function or create a new one. A new one, probably, since the etag has a precise meaning, and in a browser context there's no way to make etag works other than the way it currently does, I think.

Hub-side, we could maybe add download URLs to the response (maybe change `expand: boolean` to a list of strings, cc @Wauplin - with current behavior of `expand` kepts for backward compat) - do you think it would be needed?

packages/hub/src/lib/file-download-info.ts
axel7083: I tried the previous version, the etag is never included, as there is a redirect, therefore the headers are empty.

image

Meaning, that today the function is already not working in the browser ?

coyotte508: ah a CDN misconfiguration probably

axel7083: I tried the following function from the devtools console of https://huggingface.co/google-bert/bert-base-uncased

image

See the fetch function

```js
(await fetch("https://huggingface.co/bert-base-uncased/resolve/dd4bc8b21efa05ec961e3efc4ee5e3832a3679c7/tf_model.h5", {
    method: 'GET',
    headers: {'Range': 'byte=0-0'}
}));
```

I confirm I am getting the right value in Node

image

coyotte508: yes our CDN CORS headers are probably misconfigured, missing `Access-Control-Expose-Headers`, we'll fix it.

(Did you see my [other comment](https://github.com/huggingface/huggingface.js/pull/1024#pullrequestreview-2434410169)?)

axel7083:
> There's an endpoint `/api/:repoType(models|spaces|datasets)/:namespace?/:repo/paths-info/:rev` called with POST

Since the HEAD request is providing all the information needed (`etag`, `commitHash`, `size`) , and as you mentioned, currently it does not work in the browser because the CDN is probably missconfigured[^1], since here I am trying to get the same result as the `get_hf_file_metadata`, to me it would make sense to use the same HEAD request :thinking: 

> Not sure if we want to keep the same fileDownloadInfo function or create a new one.

I am open to create a new one, if this is not a problem to have two different method with some similar behaviour. 

> Hub-side, we could maybe add download URLs to the response

POST request a bigger than a HEAD, just in term of response size, I have a preference for getting the info from the headers.

[^1]: https://github.com/huggingface/huggingface.js/pull/1024#discussion_r1841164931

Wauplin:
> Hub-side, we could maybe add download URLs to the response (maybe change expand: boolean to a list of strings, cc @Wauplin @hanouticelina - with current behavior of expand kepts for backward compat) - do you think it would be helpful/important?

What would you see as the list of strings? A list of fields like `lastCommit`, "securityFileStatus", "downloadUrl", etc. and return only the required fields?
I don't think if it'll be super-important but if it can save some server-compute why not. If we go this path, it would be good to have the same behavior for the `expand` parameter in `/api/:repoType(models|spaces|datasets)/:namespace?/:repo/paths-info/:rev` as well for consistency. When `expand` is passed with a list of optional fields, it would be nice if the default fields are also returned (path, size, blob_id, etc.).

coyotte508:
> Since the HEAD request is providing all the information needed (`etag`, `commitHash`, `size`) , and as you mentioned, currently it does not work in the browser because the CDN is probably missconfigured[1](#user-content-fn-1-ef1b82eae4294910e4483be405a9d0b8), since here I am trying to get the same result as the `get_hf_file_metadata`, to me it would make sense to use the same HEAD request 🤔

There's another edge case. When you do a HEAD request, depending on where you make the request (eg inside a space on huggingface), you get a S3 url instead of a Cloudfront one.

And S3 has a very annoying bug: a presigned URL for HEAD request is not valid for a GET call (and vice-versa). So if in the backend my node.js process gets the HEAD url, and passes it to the frontend, and the frontend tries to download that file, it will error.

(There's also the problem when making calls from the frontend, for some of our customers which have a direct link to HF, they would get a S3 url not a cloudfront one)

With python it's solved by stopping the redirects before getting to the S3-presigned URL - by using a intermediate URL. It could be possible with NodeJS but it would mean `redirect: manual` and different browser/nodejs behavior here as well.

We've had a lot of headaches with this :sweat_smile:

Hence why the additional endpoint - good thing about the additional endpoint, you could query several files at once, instead of making a request for each one.

> What would you see as the list of strings? A list of fields like lastCommit, "securityFileStatus", "downloadUrl", etc. and return only the required fields?

Yes those 3 would be optional, and the other fields would always be provided. The problem with `lastCommit` is it can fail if requesting many files in a repo with a lot of commits, with `securityFileStatus` it makes an extra API calll, and `downloadUrl` uses some CPU resources too (not that much, but not been added/needed so far).

But for a first version we could use the endpoint as is, and add the download url if there's a need in the future.

coyotte508:
@axel7083 do you need the etag or the download url as well too? If only the etag, the `paths-info` endpoint is already suited.

The JS lib in browser faces issues that the python lib doesn't, due to CORS restrictions, trying to make it behave the same won't work, we tried (and that's why we fell back to a GET request with 0-0 range).

axel7083:
> The JS lib in browser faces issues that the python lib doesn't, due to CORS restrictions, trying to make it behave the same won't work, we tried (and that's why we fell back to a GET request with 0-0 range).

After testing it, I can confirm that this is returning the information that I need. The download link is not a problem, as the `downloadFile` do the redirect for us.

axel7083:
Moving to draft while I do more work/experiment on it. Thanks you very much for all the details. Really appreciate the feedback and reactivity :+1:

axel7083:
Closing in favour of https://github.com/huggingface/huggingface.js/pull/1031

Wauplin:
@coyotte508 I just thought about a use case for `expand: string[]`. In `datasets`, they use `/tree` to list files with their last commit. They do that we `expand=True` but don't use the security info at all.

Issue #1001, opened 2024-10-30
Suffix all imports with `.js`


coyotte508:
So that we can run code in node after a basic `npx tsc`, without module compatibility problems

cc @martin-gorner for viz

martin-gorner:
source: Node v18+ docs: https://nodejs.org/api/esm.html#mandatory-file-extensions

"Mandatory file extensions
A file extension must be provided when using the import keyword to resolve relative or absolute specifiers. Directory indexes (e.g. './startup/index.js') must also be fully specified."

martin-gorner:
I also noticed we use import statements without an extension throughout the codebase, in huggingface.js and internal projects as well. Maybe there is a rationale for that. If so, node can be made to work with extensionless imports using the "extensionless" package:
`npm install extensionless ` and then `node --import=extensionless/register test.js `

coyotte508:
We only did it for tasks, ideally we do it for all modules I think

coyotte508:
actually with TS 5.7 it may not be needed : https://devblogs.microsoft.com/typescript/announcing-typescript-5-7/#path-rewriting-for-relative-paths

🤔

PR #1026, opened 2024-11-14
✨ Switch to nodenext resolution

coyotte508:
Seems necessary for https://github.com/huggingface/huggingface.js/pull/1025 - also handle https://github.com/huggingface/huggingface.js/issues/1001 (only for tasks at the moment) cc @martin-gorner 

also so that https://arethetypeswrong.github.io/?p=@huggingface/tasks is :heavy_check_mark: cc @Pierrci  - in addition to the .cts exports

And finally export snippets types (https://github.com/huggingface-internal/moon-landing/pull/10998/files#r1841375306)

Types pass: https://arethetypeswrong.github.io/?p=%40huggingface%2Ftasks%400.13.0-test5 !

### DIFF
--- a/packages/tasks/scripts/generate-cts.ts
+++ b/packages/tasks/scripts/generate-cts.ts
@@ -1,3 +1,4 @@
+// Just copy over the already generated .d.ts and .d.ts.map files to .d.cts and .d.cts.map files
 import { readdirSync, readFileSync, statSync, writeFileSync } from "node:fs";
 import { join } from "node:path";
 
--- a/packages/gguf/src/gguf.spec.ts
+++ b/packages/gguf/src/gguf.spec.ts
@@ -33,7 +33,7 @@ describe("gguf", () => {
 			const arrayBuf = await res.arrayBuffer();
 			fs.writeFileSync(".cache/model.gguf", Buffer.from(arrayBuf));
 		}
-	});
+	}, 30_000);
 
 	it("should parse a llama2 7b", async () => {
 		const { metadata, tensorInfos } = await gguf(URL_LLAMA);
--- a/packages/tasks/package.json
+++ b/packages/tasks/package.json
@@ -1,7 +1,7 @@
 {
 	"name": "@huggingface/tasks",
 	"packageManager": "pnpm@8.10.5",
-	"version": "0.12.30",
+	"version": "0.13.0",
 	"description": "List of ML tasks for huggingface.co/tasks",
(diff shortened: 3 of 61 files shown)


packages/tasks/scripts/generate-cts.ts
coyotte508, line 1: This was the only way I found to have at the same time:

- .d.cts and .d.ts generation
- "go to source" still working

It's also possible to upgrade `tsup` and use `--dts` (instead of tsc), but then the "go to source" won't work.

Pierrci (commented):
raaaah you didn't let me the time lol

packages/tasks/scripts/generate-cts.ts
coyotte508: I don't think you need to edit the package.json, you can run tsc with `--module commonjs` maybe.

But anyway I think the types are the same, only the js flies would be different. And with your solution we still need to rename / edit the .ts and .map files.

Thinking about it, we should not use `tsup` at all for the sourcemaps to work correctly. and in that case we do indeed need to run `tsc` twice. :thinking: 

But it was already broken, problaby

coyotte508:
> raaaah you didn't let me the time lol

yes sorry didn't want more merge conflicts with all the imports :see_no_evil:

packages/tasks/scripts/generate-cts.ts
Pierrci: cc @andrewbranch too on this btw, in case he has some insights for us by any chance on the best way to emit both ESM and CJS types through `tsc` directly (while transpiling the js itself through `tsup`) 🙏

Pierrci: > I don't think you need to edit the package.json, you can run tsc with --module commonjs maybe.

well, the thing is, it's not the same thing (from the link shared in my previous message):
> CommonJS emit is similar to [--module commonjs](https://www.typescriptlang.org/docs/handbook/modules/reference.html#commonjs), but dynamic import() calls are not transformed. Emit here is shown with esModuleInterop enabled:

coyotte508: we don't have dynamic import calls anyway in `@huggingface/tasks` :nerd_face: 

(to keep in mind for `@huggingface/hub` though)

andrewbranch: https://github.com/isaacs/tshy automates this in a safe way

Pierrci: tysm for the pointer, we'll check it out!

PR #1029, opened 2024-11-14
⚗️ Use `tshy` instead of `tsup` for releasing tasks

coyotte508:
Follow up to https://github.com/huggingface/huggingface.js/pull/1026

Benefit is that exports are better defined, and source-mapping works.

Also, move the generation scripts from `@huggingace/tasks` to `@huggingface/tasks-gen` cc @Wauplin @SBrandeis 

So that the dev dependencies are not tacked on `@huggingface/tasks`. Some package managers, *cough* yarn *cough* like to download them when importing a package, even if they're irrelevant.


packages/tasks/package.json
```diff
@@ -7,33 +7,36 @@
...
-		"prepublishOnly": "pnpm run inference-codegen && git diff --name-only --exit-code src && pnpm run build",
-		"build": "tsup src/index.ts --format cjs,esm --clean && tsc --emitDeclarationOnly --declaration && pnpm run generate-cts",
-		"watch:export": "tsup src/index.ts --format cjs,esm --watch",
-		"watch:types": "tsc --emitDeclarationOnly --declaration --watch",
-		"watch": "npm-run-all --parallel watch:export watch:types",
+		"prepublishOnly": "pnpm --filter tasks-gen inference-codegen && git diff --name-only --exit-code src && pnpm run build",
+		"build": "tshy",
+		"watch:cjs": "tsc --declaration --outdir dist/commonjs --module commonjs --watch",
+		"watch:esm": "tsc --declaration --outdir dist/esm --watch",
+		"watch": "npm-run-all --parallel watch:esm watch:cjs",
```
coyotte508, line 36: cc @krampstudio , hopefully it still works

packages/tasks-gen/scripts/inference-codegen.ts
```diff
@@ -172,8 +170,10 @@ async function postProcessOutput(path2generated: string, outputSpec: Record<stri
 	return;
 }
 
-const rootDir = rootDirFinder();
+const rootDir = path.join(rootDirFinder(), "..", "tasks");
```
coyotte508, line 173: Honestly we could just do `const rootDir = "../tasks"` since the script is always executed in the `/packages/tasks-gen` context

packages/tasks/package.json
```diff
@@ -48,9 +51,10 @@
...
 	"license": "MIT",
-	"devDependencies": {
-		"@types/node": "^20.11.5",
-		"quicktype-core": "https://github.com/huggingface/quicktype/raw/pack-18.0.17/packages/quicktype-core/quicktype-core-18.0.17.tgz",
-		"type-fest": "^3.13.1"
+	"tshy": {
+		"exports": {
+			"./package.json": "./package.json",
+			".": "./src/index.ts"
+		}
```
Pierrci, line 58: that's the default, right?

packages/tasks/package.json
```diff
@@ -7,33 +7,36 @@
...
+			"import": {
+				"types": "./dist/esm/index.d.ts",
+				"default": "./dist/esm/index.js"
+			},
+			"require": {
+				"types": "./dist/commonjs/index.d.ts",
+				"default": "./dist/commonjs/index.js"
+			}
 		}
 	},
```
Pierrci, line 26: that's the result of having run `tshy` locally?

packages/tasks/package.json
```diff
@@ -7,33 +7,36 @@
 	"publishConfig": {
 		"access": "public"
 	},
-	"main": "./dist/index.cjs",
-	"module": "./dist/index.js",
-	"types": "./dist/index.d.ts",
+	"sources": "./src/index.ts",
+	"main": "./dist/commonjs/index.js",
+	"module": "./dist/esm/index.js",
+	"types": "./dist/commonjs/index.d.ts",
```
Pierrci, line 13: iiuc from https://github.com/isaacs/tshy#old-style-exports-top-level-main-module-types, we should set `tshy.module` and `tshy.main` to `false` (I don't think we care about node < 12?)

packages/tasks/.tshy/build.json
Pierrci, line 1: were all the files in `.tshy/` generated by `tshy`?

Pierrci (commented):
didn't try it (obviously), but lgtm 😄

packages/tasks/package.json
coyotte508: it set itself like that when running `pnpm build`, I didn't add it :)

packages/tasks/package.json
coyotte508: yes, it did that itself

packages/tasks/.tshy/build.json
coyotte508: yes

packages/tasks/package.json
```diff
@@ -7,33 +7,33 @@
 	"publishConfig": {
 		"access": "public"
 	},
-	"main": "./dist/index.cjs",
-	"module": "./dist/index.js",
-	"types": "./dist/index.d.ts",
+	"sources": "./src/index.ts",
```
Pierrci, line 10: I didn't know about this one lol

Pierrci:
interesting CI errors haha ^^'

coyotte508:
> interesting CI errors haha ^^'

well yea, going bakc to adding main/module/types fields

Pierrci:
> Warning: relying on top-level main/types will likely cause incorrect types to be loaded in some scenarios.
> 
> Use with extreme caution. It's almost always better to not define top-level main and types fields if you are shipping a hybrid module. Users will need to update their module and moduleResolution tsconfigs appropriately. That is a good thing, and will save them future headaches.

but yeah for now probably simpler :)

coyotte508:
> > Warning: relying on top-level main/types will likely cause incorrect types to be loaded in some scenarios.
> > Use with extreme caution. It's almost always better to not define top-level main and types fields if you are shipping a hybrid module. Users will need to update their module and moduleResolution tsconfigs appropriately. That is a good thing, and will save them future headaches.
> 
> but yeah for now probably simpler :)

I could remove the fields, but then we have to update the moon codebase ASAP :)

Pierrci (approved):

PR #1031, opened 2024-11-15
feat(hub): adding pathsInfo function

axel7083:
## Description

Following discussion https://github.com/huggingface/huggingface.js/pull/1024 and incompatibility of using the `HEAD` request to get the same etag as the python library is using for populating the cache directory.

This PR add the `pathsInfo` function that return the paths information including the LFS oid (or etag) if the file is a LFS pointer.

As suggested by  @coyotte508 in https://github.com/huggingface/huggingface.js/pull/1024#pullrequestreview-2434410169

## Related issues

Fixes https://github.com/huggingface/huggingface.js/issues/1023 (provide an alternative method to `fileDownloadInfo`.

## Tests

- [x] unit tests has been added

### DIFF
--- a/packages/hub/src/lib/index.ts
+++ b/packages/hub/src/lib/index.ts
@@ -19,6 +19,7 @@ export * from "./model-info";
 export * from "./oauth-handle-redirect";
 export * from "./oauth-login-url";
 export * from "./parse-safetensors-metadata";
+export * from "./paths-info";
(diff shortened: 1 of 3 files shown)


coyotte508 (commented):
Thank you!

You should export the function in `lib/index.ts`

coyotte508 (approved):

PR #1035, opened 2024-11-15
export type ModelDataMinimal

mishig25:
export type ModelDataMinimal because it is needed in moon


julien-c (commented):
this is a bit weird bc ModelData is originally in moon-landing, so it's weird/cyclical to re-import a partial copy of it 

I'd rather redefine it in moon-landing in some way

mishig25:
yeah indeed, that would be better

coyotte508:
JFI it's already exported as part of https://github.com/huggingface/huggingface.js/pull/1026

via `snippets.ModelDataMinimal`

PR #1036, opened 2024-11-18
export type InferenceSnippet from @hfjs/tasks

mishig25:
export type InferenceSnippet from @hfjs/tasks


coyotte508:
yes, you would need to import `snippets` directly (before this PR)

PR #1040, opened 2024-11-18
Remove "words" key from DQA output

Rocketknight1:
cc @Wauplin - this matches `transformers` because `transformers` never outputs a `words` key. I'll make another PR to `transformers` to remove the key from the docstrings as well.


Wauplin (approved):
Thanks for spotting it @Rocketknight1 !

PR #1043, opened 2024-11-19
Tasks: add `function_to_apply`'s docstring for classification tasks

hanouticelina:
This PR adds missing parameter description for `function_to_apply`. this documentation is required since huggingface_hub's `InferenceClient` documentation is automatically generated from the specs defined here (See [huggingface_hub/.github/workflows/update-inference-types.yaml](https://github.com/huggingface/huggingface_hub/blob/main/.github/workflows/update-inference-types.yaml) for more details). 

Note: the description was not technically missing since it's already available in `ClassificationOutputTransform`'s description, but it wasn't being properly inferred for the `function_to_apply` property.


Wauplin (approved):
Thank you!

PR #1045, opened 2024-11-21
[Tasks] remove duplicate description in task parameter objects

hanouticelina:
this PR removes redundant description in task parameter classes definition while keeping it in the parameters field reference. this avoids duplicate documentation, see: [huggingface/huggingface_hub/pull/2664#discussion](https://github.com/huggingface/huggingface_hub/pull/2664#discussion_r1851713362) for more context.

Note : `packages/tasks/src/tasks/{task_name/inference.ts` files are auto-generated.


Wauplin (approved):
Thanks for cleaning this up!

PR #1048, opened 2024-11-25
update peft code snippet

### LINKED ISSUE #1047: Update peft code snippet
The peft code snippet loads a `PeftConfig` parameter which is not used anywhere, I think it would be better to remove that since the entire script functions without it.
cc @BenjaminBossan if you have any inputs about this.
<hr>

image
<hr>

image

not-lain:
this pr fixes #1047

### DIFF
--- a/packages/tasks/src/model-libraries-snippets.ts
+++ b/packages/tasks/src/model-libraries-snippets.ts
@@ -933,10 +933,9 @@ export const peft = (model: ModelData): string[] => {
 	}
 
 	return [
-		`from peft import PeftModel, PeftConfig
+		`from peft import PeftModel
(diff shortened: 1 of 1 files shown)


Wauplin:
_I think_ that the first intent was to showcase what are the different components that can be loaded using `peft`. This is also what's done for `transformers` models. Example:

image

`AutoProcessor` is not always required here, right?

cc @LysandreJik @julien-c if you have more context for when these snippets have been introduced? (see convo in https://github.com/huggingface/huggingface.js/issues/1047 as well)

LysandreJik (approved):
Agree that the change is cleaner! As @not-lain says, the config is a component of the model, it's already done within the model instantiation call

PR #1052, opened 2024-12-02
:technologist:Optimized and tested the 'object_fonts.ts' code

miaowumiaomiaowu:
I optimized the code in the `inject_fonts.ts` file to make it **more aesthetically pleasing and efficient**.

I also wrote test code in the `inject_fonts.test.ts` for the `inject_fonts.ts` file using jest.
The tests include:
- **should inject Source Sans Pro font**: Tests whether the "Source Sans Pro" font is correctly injected, with a runtime of 15 milliseconds.
- **should inject IBM Plex Mono font**: Tests whether the "IBM Plex Mono" font is correctly injected, with a runtime of 3 milliseconds.
- **should not inject duplicate fonts**: Tests whether duplicate fonts are avoided during injection, with a runtime of 2 milliseconds.
The test results are as follows:
image

Before running the tests with Jest, you need to install the required dependencies:
`pnpm add -D ts-jest jest-environment-jsdom`

Additionally, I have added the necessary configurations in the file, so you can simply run `pnpm test` to execute the tests.


julien-c:
Hi! What issue does this solve? Do you have a screen recording or something?

package-lock.json
coyotte508, line 1: do not use npm install, use pnpm

miaowumiaomiaowu:
> Hi! What issue does this solve? Do you have a screen recording or something?

coyotte508:
This PR has a 20k lines diff for non-necessary changes.

PR #1060, opened 2024-12-03
local-apps: update llama.cpp snippet to use cmake

ngxson:
Ref: https://github.com/ggerganov/llama.cpp/pull/10514

The recommended way to build llama.cpp is by using `cmake` command.


Vaibhavs10 (approved):
Thanks!

PR #1061, opened 2024-12-04
Add Sana .pth files into download counting in huggingface repo

lawrence-cj:


packages/tasks/src/model-libraries.ts
```diff
@@ -787,6 +787,13 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		filter: true,
 		countDownloads: `path_extension:"sentis"`,
 	},
+	"sana": {
```
Vaibhavs10, line 790:
```suggestion
	"Sana": {
```

Since the tags in the repo are w/ capital `S`: https://huggingface.co/models?other=Sana

packages/tasks/src/model-libraries-snippets.ts
```diff
@@ -982,6 +982,35 @@ IWorker engine = WorkerFactory.CreateWorker(BackendType.GPUCompute, model);
 `,
 ];
 
+export const sana = (model: ModelData): string[] =>	[
```
Vaibhavs10, line 985: would it be possible to reduce the size of the snippet if possible? the idea behind snippets is that it should be the minimum amount of code required to get an inference.

perhaps we can remove the `generator` bits and assume that the model is loaded to `cuda` by default.

Vaibhavs10 (commented):
Thanks for the PR @lawrence-cj left 2 quick suggestions!

packages/tasks/src/model-libraries-snippets.ts
lawrence-cj: Of course. Let me fix it.

lawrence-cj: Done

lawrence-cj:
Cool!

packages/tasks/src/model-libraries-snippets.ts
```diff
@@ -982,6 +982,28 @@ IWorker engine = WorkerFactory.CreateWorker(BackendType.GPUCompute, model);
...
+		`
+# Install from https://github.com/NVlabs/Sana
+
+# Load the model and infer image from text
```
Wauplin, line 989:
```suggestion
# Load the model and infer image from text
```

No need to link github repo in snippet. A button will be shown just above the snippet for the user to click on. See example for seed-story repo:
image

packages/tasks/src/model-libraries.ts
```diff
@@ -787,6 +787,13 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
...
+		prettyLabel: "Sana",
+		repoName: "Sana",
+		repoUrl: "https://github.com/NVlabs/Sana",
+		countDownloads: `path_extension:"pth" OR path_extension:"json"`,
```
Wauplin, line 794: .pth file each time the pipeline is instantiated. Is that the case or not? And is the config.json ever loaded from script?

1. If the script only loads the .pth file, then let's count only on .pth:
```suggestion
		countDownloads: `path_extension:"pth"`,
```

2. If the script loads the config file as well, then it's even better and we should count these calls and only these calls. config.json is already tracked by default and that's why some count are already listed on the Hub (see https://huggingface.co/Efficient-Large-Model/Sana_1600M_512px). In that case, no need to add a countDownloads rule.

```suggestion
```


I let you decide what's best for Sana library.

Wauplin (approved):
Hey there, thanks for the PR and the back and forth with @Vaibhavs10  :) I left 2 comments and then we should be good to go.

packages/tasks/src/model-libraries.ts
lawrence-cj: Actually, we didn't have a config.json file until yesterday. We want to recover it, so we add pth here. Is this reasonable?

Wauplin (approved):
Looks good!

packages/tasks/src/model-libraries.ts
Wauplin: ok, makes sense to track pth files then :+1:

packages/tasks/src/model-libraries.ts
Wauplin, line 790:
```suggestion
	sana: {
```

packages/tasks/src/model-libraries-snippets.ts
```diff
@@ -982,6 +982,26 @@ IWorker engine = WorkerFactory.CreateWorker(BackendType.GPUCompute, model);
 `,
 ];
 
+export const sana = (model: ModelData): string[] =>	[
```
Wauplin, line 985:
```suggestion
export const sana = (model: ModelData): string[] => [
```

packages/tasks/src/model-libraries-snippets.ts
```diff
@@ -982,6 +982,26 @@ IWorker engine = WorkerFactory.CreateWorker(BackendType.GPUCompute, model);
 `,
 ];
 
+export const sana = (model: ModelData): string[] => [
+		`
```
Wauplin, line 986:
```suggestion
	`
```

Wauplin:
Lint is finally passing :smile: 
I'm merging. Expect 2-3 days before getting it deployed to production.

lawrence-cj:
OHH, Great to know. Thanks a lot for your support and quick apply @Wauplin @Vaibhavs10

lawrence-cj:
Hi experts. Sorry for interrupting. The downloading count does not recover in our collection. For example, https://huggingface.co/Efficient-Large-Model/Sana_1600M_1024px
Could you pls help to check the problem?
Cc: @Wauplin @Vaibhavs10

Wauplin:
Hey @lawrence-cj I can confirm to you that this PR has been deployed to production. I can now see 7400+ downloads for https://huggingface.co/Efficient-Large-Model/Sana_1600M_1024px for instance. Is that more in line with what you were expecting?

lawrence-cj:
Oh, thank you so much for it @Wauplin. Actually, I would have expected more as we have 70K downloads of [our component models](https://huggingface.co/mit-han-lab/dc-ae-f32c32-sana-1.0). But that's ok, I'm ok if that's the backend data.

PR #1066, opened 2024-12-06
Add Birder as a library to model-libraries.ts

hassonofer:
Allow correct download stats.

### DIFF
--- a/packages/tasks/src/model-libraries.ts
+++ b/packages/tasks/src/model-libraries.ts
@@ -114,6 +114,13 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		filter: false,
 		countDownloads: `path_extension:"npz"`,
 	},
+	birder: {
+		prettyLabel: "Birder",
+		repoName: "Birder",
+		repoUrl: "https://gitlab.com/birder/birder",
+		filter: false,
+		countDownloads: `path_extension:"pt"`,
+	},
 	birefnet: {
 		prettyLabel: "BiRefNet",
 		repoName: "BiRefNet",


packages/tasks/src/model-libraries.ts
```diff
@@ -114,6 +114,13 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		filter: false,
 		countDownloads: `path_extension:"npz"`,
 	},
+	birder: {
```
Vaibhavs10, line 117: Thanks, just to double confirm are these the only two model checkpoints supported by the tag: https://huggingface.co/models?other=birder

packages/tasks/src/model-libraries.ts
```diff
@@ -114,6 +114,13 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
...
+		prettyLabel: "Birder",
+		repoName: "Birder",
+		repoUrl: "https://gitlab.com/birder/birder",
+		filter: false,
+		countDownloads: `path_extension:"pt"`,
```
Vaibhavs10, line 122: Out of curiousity, wouldn't the `config.json` not work as well? I see that the downloads already show up for: https://huggingface.co/birder-project/davit_tiny_il-all

Vaibhavs10 (commented):
Thanks for the PR @hassonofer

packages/tasks/src/model-libraries.ts
hassonofer: Yeah, I'm still testing before uploading more...
I have around 30 more models with support of more geographical regions.

packages/tasks/src/model-libraries.ts
hassonofer: In the Birder library, there are a few ways to load a model (https://gitlab.com/birder/birder/-/blob/main/docs/internal/api.md?ref_type=heads)

The load with config is not a popular one, usually only the weights are being accessed.

hassonofer:
Thank you @Vaibhavs10

packages/tasks/src/model-libraries.ts
julien-c: did you see by the way @sashavor? https://gitlab.com/birder/birder

packages/tasks/src/model-libraries.ts
Vaibhavs10: Nice! Lovely! Let us know if you need any help.

Vaibhavs10 (approved):
This looks good to me to merge. Let's wait for one more approval before merging.

packages/tasks/src/model-libraries.ts
hassonofer: Thanks a lot :)

I'm good for now, documentation is really good.

PR #1072, opened 2024-12-13
✨ GenAIScript local app

pelikhan:
GenAIScript is a local app/tool (https://microsoft.github.io/genaiscript/) that can run HF models through Ollama or the HF inference engine.

This change adds a "local-apps" entry.


Vaibhavs10 (commented):
cc: @ngxson for the ollama quant logic

coyotte508:
`parseGGUFQuantLabel` is already used in llamaccp and ollama snippets so I think it's ok in this PR

PR #1073, opened 2024-12-13
Update hardware.ts - Add 4 Intel Arc GPUs

qnixsynapse:
Added 4 details of Intel Arc GPUs including the recently launched battlemage GPUs. 

I did it via mobile. Please check for accuracies before merging. Thanks.

Cc: @Vaibhavs10

### DIFF
--- a/packages/tasks/src/hardware.ts
+++ b/packages/tasks/src/hardware.ts
@@ -303,6 +303,24 @@ export const SKUS = {
 				memory: [16],
 			},
 		},
+		INTEL: {
+			"Arc A750": {
+				tflops: 34.41,
+				memory: [8],
+			},
+			"Arc A770": {
+				tflops: 39.32,
+				memory: [16],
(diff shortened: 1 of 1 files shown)


packages/tasks/src/hardware.ts
```diff
@@ -307,6 +307,24 @@ export const SKUS = {
...
 			},
 		},
+		INTEL: {
+			"Arc A750": {
+				tflops: 34.41,
+				memory: [8],
+			},
+			"Arc A770": {
+				tflops: 39.32,
+				memory: [16],
```
Vaibhavs10, line 317:
```suggestion
				memory: [8, 16],
```

Vaibhavs10 (approved):
Thanks for the PR @qnixsynapse 

Verified the numbers through here: https://en.wikipedia.org/wiki/Intel_Arc

PR #1076, opened 2024-12-14
gguf: calculate tensor data offset

drazdra:
adding tensors offset within file. offsets stored in metadata are relative to this offset and it's absent in metadata. with this field it's possible to actually access layers in gguf. without of it - offsets shown are useless.

### DIFF
--- a/packages/gguf/src/gguf.ts
+++ b/packages/gguf/src/gguf.ts
@@ -10,8 +10,8 @@ export { parseGGUFQuantLabel, GGUF_QUANT_RE, GGUF_QUANT_RE_GLOBAL } from "@huggi
 
 export const RE_GGUF_FILE = /\.gguf$/;
 export const RE_GGUF_SHARD_FILE = /^(?<prefix>.*?)-(?<shard>\d{5})-of-(?<total>\d{5})\.gguf$/;
-const GGUF_DEFAULT_ALIGNMENT = 32; //defined in ggml.h
-const GGML_PAD = (x: number, n: number) => (x + n - 1) & ~(n - 1); //defined in ggml.h
+const GGUF_DEFAULT_ALIGNMENT = 32; // defined in ggml.h
+const GGML_PAD = (x: number, n: number) => (x + n - 1) & ~(n - 1); // defined in ggml.h
 const PARALLEL_DOWNLOADS = 20;
 
 export interface GgufShardFileInfo {
--- a/packages/gguf/src/gguf.ts
+++ b/packages/gguf/src/gguf.ts
@@ -10,6 +10,8 @@ export { parseGGUFQuantLabel, GGUF_QUANT_RE, GGUF_QUANT_RE_GLOBAL } from "@huggi
 
 export const RE_GGUF_FILE = /\.gguf$/;
 export const RE_GGUF_SHARD_FILE = /^(?<prefix>.*?)-(?<shard>\d{5})-of-(?<total>\d{5})\.gguf$/;
+const GGUF_DEFAULT_ALIGNMENT = 32; //defined in ggml.h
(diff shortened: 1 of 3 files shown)


mishig25:
@drazdra happy to merge after:
1. linting issues are solved
2. `tensorsOffset` is added to https://github.com/huggingface/huggingface.js/blob/main/packages/gguf/src/gguf.spec.ts

julien-c:
does this look valid @ngxson @Vaibhavs10?

mishig25:
> with this field it's possible to actually access layers in gguf. without of it - offsets shown are useless.

@ngxson is this correct?

drazdra:
sorry about errors, i don't know TS, in my case i just patched the ready file in dist with JS and used it as is.

ngxson:
This will introduce a breaking change for anyone who are using this `tensorOffset` field. I'd suggest adding a new field for it, for example `tensorOffsetInFile`. I'll make a PR for that.

packages/gguf/src/gguf.ts
```diff
@@ -384,14 +384,17 @@ export async function gguf(
 		});
 	}
 
+  	const alignment		=metadata['general.alignment']??32;
+	const tensorsOffset	=offset+(alignment-(offset%alignment));
```
ngxson, line 388: This calculation is not correct because the current `offset` is the offset of the current tensor metadata, not the offset of whole metadata

drazdra: current offset is taken after tensor metadata ends, that's where starts actual tensors. also, i did check the value received with the same value reported by gguf tool from llama.cpp, it matched. 
https://github.com/ggerganov/ggml/blob/master/docs/gguf.md

so i don't really understand why you believe it's wrong :).

ngxson: ok sorry I misread the code. I thought that here you're setting it for each tensor. It's correct to use the `offset` only after we're done reading all data.

ngxson:
I changed the naming to `tensorDataOffset` instead of `tensorsOffset`, because it will be more aligned to `gguf_get_data_offset()`. Here is the equivalent cpp code:

```cpp
int main(int argc, const char ** argv) {
    struct ggml_context * ctx_meta = NULL;
    struct gguf_init_params gguf_params = {
        /*.no_alloc = */ true,
        /*.ctx      = */ &ctx_meta,
    };
    auto * ctx_gguf = gguf_init_from_file("/Users/ngxson/Downloads/gguf_test_big_metadata.gguf", gguf_params);
    printf("gguf_get_data_offset: %zu\n", gguf_get_data_offset(ctx_gguf));
}
```

Note sure why node test failed, I'll have a look later (I'm using a slow network)

ngxson:
Should be ok now, matched the output from cpp code earlier.

Pinging @mishig25 for an extra review.

packages/gguf/src/gguf.ts
```diff
@@ -10,6 +10,8 @@ export { parseGGUFQuantLabel, GGUF_QUANT_RE, GGUF_QUANT_RE_GLOBAL } from "@huggi
 
 export const RE_GGUF_FILE = /\.gguf$/;
 export const RE_GGUF_SHARD_FILE = /^(?<prefix>.*?)-(?<shard>\d{5})-of-(?<total>\d{5})\.gguf$/;
+const GGUF_DEFAULT_ALIGNMENT = 32; //defined in ggml.h
+const GGML_PAD = (x: number, n: number) => (x + n - 1) & ~(n - 1); //defined in ggml.h
```
julien-c, line 14:
```suggestion
const GGUF_DEFAULT_ALIGNMENT = 32; // defined in ggml.h
const GGML_PAD = (x: number, n: number) => (x + n - 1) & ~(n - 1); // defined in ggml.h
```

julien-c (approved):

PR #1089, opened 2025-01-07
Add cosmos library

pcuenca:
This enables downloads [once these models are assigned the `cosmos` library](https://huggingface.co/collections/nvidia/cosmos-6751e884dc10e013a0a0d8e6). Snippets can come later.

### DIFF
--- a/packages/tasks/src/model-libraries.ts
+++ b/packages/tasks/src/model-libraries.ts
@@ -170,7 +170,6 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		prettyLabel: "Cosmos",
 		repoName: "Cosmos",
 		repoUrl: "https://github.com/NVIDIA/Cosmos",
-		filter: false,
 		countDownloads: `path:"config.json" OR path_extension:"pt"`,
 	},
 	"cxr-foundation": {
--- a/packages/tasks/src/model-libraries.ts
+++ b/packages/tasks/src/model-libraries.ts
@@ -166,6 +166,13 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		filter: false,
 		countDownloads: `path:"adapter_config.json"`,
 	},
+	cosmos: {
+		prettyLabel: "Cosmos",
+		repoName: "Cosmos",
+		repoUrl: "https://github.com/NVIDIA/Cosmos",
+		filter: false,
+		countDownloads: `path:"config.json" OR path_extension:"pt"`,
+	},
 	"cxr-foundation": {
 		prettyLabel: "CXR Foundation",
 		repoName: "cxr-foundation",


packages/tasks/src/model-libraries.ts
```diff
@@ -166,6 +166,13 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		filter: false,
 		countDownloads: `path:"adapter_config.json"`,
 	},
+	cosmos: {
+		prettyLabel: "Cosmos",
+		repoName: "Cosmos",
+		repoUrl: "https://github.com/NVIDIA/Cosmos",
+		filter: false,
```
pcuenca, line 173:
```suggestion
```

julien-c (approved):

Vaibhavs10 (approved):
Nice! Thanks! Should be much more easier to have snippets once/ if in transformers.

Haoxiang-Wang:
Thank you @pcuenca !

I am the Cosmos team member who is in charge of our model uploading to Hugging Face Hub. I noticed most of our Cosmos-1.0 models have no download statistics counted yet (see the screenshot below), which concerns us.

@Vaibhavs10 @julien-c  Please merge this MR as soon as possible, and I will tag all Cosmos-1.0 models with the `cosmos` library. I think this can turn on the download counter, right?

cosmos-downloads

Vaibhavs10 (commented):
Hi @Haoxiang-Wang - It should merge today, note after the PR is merged and after a deployment. It'd take 24-48 hours for download counters to come in effect.

and, yes, please do update the `library_name` to `cosmos` 🤗

packages/tasks/src/model-libraries.ts
```diff
@@ -166,6 +166,12 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		filter: false,
 		countDownloads: `path:"adapter_config.json"`,
 	},
+	cosmos: {
```
Wauplin, line 169: Unrelated to this PR but are these models supposed to be usable with Nemo ? Since they are tagged as "nemo", there is a "use this model" snippet in all of them but with empty snippets. This is due to the fact that the nemo code snippet generator is incomplete (see here https://github.com/huggingface/huggingface.js/blob/504fcb4fd8d9774327770a3b949d9812823138e0/packages/tasks/src/model-libraries-snippets.ts#L1082 and https://github.com/huggingface/huggingface.js/blob/504fcb4fd8d9774327770a3b949d9812823138e0/packages/tasks/src/model-libraries-snippets.ts#L959). @Haoxiang-Wang @pcuenca  would it make sense to open a separate PR to complete the snippets?

pcuenca:
Hi @Wauplin, we took a quick look with @ariG23498 to create a Cosmos-based snippet, but it was not immediately obvious so we decided to tackle it later. Happy to revisit! I believe that the compatibility with nemo in some of these models comes from the existence of a `nemo` subfolder, such as in [this model](https://huggingface.co/nvidia/Cosmos-1.0-Autoregressive-4B/tree/main). If that's the case, there's a [similar discussion in a related PR](https://github.com/huggingface/huggingface.js/pull/1090#issuecomment-2579972538) about potentially considering whether to split those repos.

Other models, however, [do not contain a `nemo` folder but still have the tag](https://huggingface.co/nvidia/Cosmos-1.0-Autoregressive-5B-Video2World/tree/main).

Additional guidance from @Haoxiang-Wang would be super helpful here.

PR #661, opened 2024-05-10
Hardware inventory

julien-c:
I hesitated putting this one in `@huggingface/tasks`, or creating a new `@huggingface/hardware`.

What do you think?

The idea is the community to be able to contribute to that list.

### What I picked for now ⤵️

Because I'm lazy – and because it's somewhat linked to https://github.com/huggingface/huggingface.js/pull/659 – i've added it to `tasks`.

PR #1095, opened 2025-01-11
Add AMD Ryzen Threadripper 7000 series

melroy89:
- Adding AMD Ryzen Threadripper 7000 series to the drop-down selection option.
- Use 10 TFLOPS for this Threadripper generation.
- Update comment to explain we use FP32 results for CPUs

---

And it seems there might be no FP16 or FP32 measurements done yet on either cpu-monkey.com or techpowerup.com regarding the Threadripper 7000 serie CPUs.

And are like 7 or 8 CPUs in this 7000 series, so I would pick a AMD Ryzen Threadripper 7970X or AMD Ryzen Threadripper 7980X as a generic CPU for the whole 7000 serie.  
Then again, I have no clue what the tflops would be.

I personally own an AMD Ryzen Threadripper 7960X. Could I do some benchmark myself? If so, which tool or calculation (stress) test do you advice to give us a good indication?

### DIFF
--- a/packages/tasks/src/hardware.ts
+++ b/packages/tasks/src/hardware.ts
@@ -14,7 +14,7 @@ export const TFLOPS_THRESHOLD_EU_AI_ACT_MODEL_TRAINING_TOTAL = 10 ** 13;
 
 export interface HardwareSpec {
 	/**
-	 * Approximate value, in FP16 whenever possible.
+	 * Approximate value, in FP16 whenever possible of GPUs and FP32 for CPUs.
 	 * This is only approximate/theoretical and shouldn't be taken too seriously.
 	 * Currently the CPU values are from cpu-monkey.com
 	 * while the GPU values are from techpowerup.com
@@ -417,7 +417,7 @@ export const SKUS = {
 				tflops: 0.6,
 			},
 			"Ryzen Zen 4 7000 (Threadripper)": {
-				tflops: ?,
+				tflops: 10.0,
 			},
 			"Ryzen Zen4 7000 (Ryzen 9)": {
 				tflops: 0.56,
--- a/packages/tasks/src/hardware.ts
+++ b/packages/tasks/src/hardware.ts
@@ -14,7 +14,7 @@ export const TFLOPS_THRESHOLD_EU_AI_ACT_MODEL_TRAINING_TOTAL = 10 ** 13;
 
 export interface HardwareSpec {
 	/**
-	 * Approximate value, in FP16 whenever possible of GPUs and FP32 for CPUs.
+	 * Approximate value, in FP16 whenever possible for GPUs and FP32 for CPUs.
 	 * This is only approximate/theoretical and shouldn't be taken too seriously.
 	 * Currently the CPU values are from cpu-monkey.com
 	 * while the GPU values are from techpowerup.com
--- a/packages/tasks/src/hardware.ts
+++ b/packages/tasks/src/hardware.ts
@@ -416,6 +416,9 @@ export const SKUS = {
 			"EPYC 1st Generation (Naples)": {
 				tflops: 0.6,
 			},
+			"Ryzen Zen 4 7000 (Threadripper)": {
+				tflops: ?,
+			},
 			"Ryzen Zen4 7000 (Ryzen 9)": {
 				tflops: 0.56,
 			},


pcuenca:
Hi @melroy89! In general, we try to use representative numbers for the way the hardware is typically used or benchmarked, which usually means fp16 for GPU and mostly fp32 for CPUs. For CPU families, [we talked in this discussion](https://github.com/huggingface/huggingface.js/pull/661#issuecomment-2110391440) about using numbers between the min and the max.

But it does seem complicated in this case, I could not find much information either. I saw 190,950 MOps/Sec of ["floating path math" here](https://www.cpubenchmark.net/cpu.php?cpu=AMD+Ryzen+Threadripper+7960X&id=5792), but I don't know if that benchmark is comparable to the specs published in techpowerup and other places. I hope you get luckier and can find some info somewhere :)

melroy89:
I literally own this threadripper CPU, so if you let me know what benchmark to run, I know at least the `min` value of fp32.

pcuenca:
> I literally own this threadripper CPU, so if you let me know what benchmark to run, I know at least the `min` value of fp32.

Hi @melroy89, we don't run benchmarks, we retrieve numbers reported by trustworthy resources. I believe in most cases they are theoretical figures calculated taken into account the frequency, number of cores, number of floating ops per clock cycle, etc., but I haven't looked at the exact process yet.

melroy89:
Well normally you get these numbers by running benchmarks, just like your "trusted resources" are using benchmarks to get the numbers.

https://www.tweaktown.com/news/93928/amds-new-ryzen-threadripper-pro-7995wx-cpu-has-more-fp32-perf-than-xbox-series-and-ps5/index.html

Here I found a number of the top-tier Threadripper 7000 in TFLOPs of FP32:

> The AIDA64 GPGPU benchmark sees the Ryzen Threadripper PRO 7995WX processor spitting out 12.16 TFLOPs of FP32 compute performance

As you can see, here they use a benchmark tool called "AIDA64 GPGPU benchmark" under Windows.

----

Read more: https://www.tweaktown.com/news/93928/amds-new-ryzen-threadripper-pro-7995wx-cpu-has-more-fp32-perf-than-xbox-series-and-ps5/index.html

melroy89:
For now I think the best number I come up with is around 10 TFLOPS actually for FP32 on Threadripper 7000 series. So I'm below the high-end CPU results of 12.16 TFLOPS.

pcuenca:
Hi @melroy89, sorry if I caused confusion, I meant that in many cases these aren't really measured performance numbers, but theoretical ones based on specs (see below). I'm happy to accept the figures you got from your real-life benchmark, and grateful that you took the time to run it!

----

As for how specs can translate to teraflops, consider a [3090 GPU](https://www.techpowerup.com/gpu-specs/geforce-rtx-3090.c3622). From the specs sheet we see that it has a "boost" clock frequency of 1695 Mhz and 10496 parallel "shading" units, or cuda cores. According to [this page](https://en.wikipedia.org/wiki/Floating_point_operations_per_second), the Ampere cards can run 2 FP32 operations per clock cycle using the PTX instruction set, so we get `1.695 * 10496 * 2 = 35.58 tflops`, same as what appears in the techpowerup sheet. I don't know why the fp16 performance is the same.

Consider now a CPU, the [Intel® Core™ i9-7980XE Extreme Edition](https://www.intel.com/content/www/us/en/products/sku/126699/intel-core-i97980xe-extreme-edition-processor-24-75m-cache-up-to-4-20-ghz/specifications.html). It's "Max Turbo Frequency" is 4.20 GHz, and it has 18 parallel cores. It has two AVX-512 fused-multiply-add (FMA) units that work on 512-bit numbers. The two units run in parallel, but each one takes two cycles to compute an FMA instruction (see the [throughput number for Skylake-X](https://uops.info/html-instr/VFMADD231PD_ZMM_ZMM_ZMM.html#SKX), which is 0.5). AVX-512 is capable of computing one 512-bit operation, or 16 (512/32) "packed" fp32 operations. Putting it together, we have a theoretical performance of `4.20 * 18 * 512 * 2 * 0.5 / 32 = 1.2 tflops`, which is close to the 1.3 tflops [reported here](https://www.intel.com/content/dam/develop/external/us/en/documents/pdf/raw-compute-power-of-new-i9-processor-based-enables-megatasking.pdf) (at 4.3 GHz instead of 4.2).

So essentially these numbers come from the manufacturers and not from real-world benchmarks that use more operations than just FMA, reflect overhead, thermal throttling, etc. But I have no idea how to track down these numbers for the threadrippers and be sure that the calculations are correct, so happy to accept your measures!

pcuenca (approved):

packages/tasks/src/hardware.ts
```diff
@@ -14,7 +14,7 @@ export const TFLOPS_THRESHOLD_EU_AI_ACT_MODEL_TRAINING_TOTAL = 10 ** 13;
 
 export interface HardwareSpec {
 	/**
-	 * Approximate value, in FP16 whenever possible.
+	 * Approximate value, in FP16 whenever possible of GPUs and FP32 for CPUs.
```
julien-c, line 17:
```suggestion
	 * Approximate value, in FP16 whenever possible for GPUs and FP32 for CPUs.
```

julien-c (approved):

melroy89:
Fixed typos

PR #1098, opened 2025-01-13
Update Tasks

merveenoyan:
Update models, datasets, Spaces and to content where necessary

### DIFF
--- a/packages/tasks/src/tasks/audio-to-audio/data.ts
+++ b/packages/tasks/src/tasks/audio-to-audio/data.ts
@@ -39,8 +39,8 @@ const taskData: TaskDataCustom = {
 	],
 	models: [
 		{
-			description: "A solid model of audio source separation.",
-			id: "speechbrain/sepformer-wham",
+			description: "A robust model to compress speech.",
+			id: "stabilityai/stable-codec-speech-16k",
 		},
 		{
 			description: "A speech enhancement model.",
--- a/packages/tasks/src/tasks/audio-to-audio/data.ts
+++ b/packages/tasks/src/tasks/audio-to-audio/data.ts
@@ -39,10 +39,6 @@ const taskData: TaskDataCustom = {
 	],
 	models: [
 		{
-			description: "A robust model to compress speech.",
-			id: "stabilityai/stable-codec-speech-16k",
-		},
-		{
 			description: "A speech enhancement model.",
 			id: "ResembleAI/resemble-enhance",
 		},
--- a/packages/tasks/src/tasks/text-generation/data.ts
+++ b/packages/tasks/src/tasks/text-generation/data.ts
@@ -3,25 +3,25 @@ import type { TaskDataCustom } from "../index.js";
 const taskData: TaskDataCustom = {
 	datasets: [
 		{
-			description: "A large multilingual dataset of text crawled from the web.",
-			id: "mc4",
+			description: "Multilingual dataset used to evaluate text generation models.",
+			id: "CohereForAI/Global-MMLU",
 		},
 		{
 			description:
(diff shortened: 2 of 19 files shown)


packages/tasks/src/tasks/text-generation/data.ts
```diff
@@ -3,25 +3,25 @@ import type { TaskDataCustom } from "../index.js";
...
 		{
-			description: "An instruction dataset with preference ratings on responses.",
-			id: "openbmb/UltraFeedback",
+			description: "A multilingual instruction dataset with preference ratings on responses.",
+			id: "allenai/llama-3.1-tulu-3-8b-preference-mixture",
 		},
 		{
 			description: "A large synthetic dataset for alignment of text generation models.",
-			id: "argilla/magpie-ultra-v0.1",
+			id: "HuggingFaceTB/smoltalk",
```
Vaibhavs10, line 24: Maybe we can use FineMath here: https://huggingface.co/datasets/HuggingFaceTB/finemath

packages/tasks/src/tasks/text-to-video/data.ts
```diff
@@ -68,15 +68,15 @@ const taskData: TaskDataCustom = {
...
 		{
 			description: "A strong model for consistent video generation.",
-			id: "rain1011/pyramid-flow-sd3",
+			id: "tencent/HunyuanVideo",
 		},
 		{
-			description: "A robust model for text-to-video generation.",
-			id: "VideoCrafter/VideoCrafter2",
+			description: "A text-to-video model with high fidelity motion and strong prompt adherence.",
+			id: "genmo/mochi-1-preview",
```
Vaibhavs10, line 75:
```suggestion
			id: "Lightricks/LTX-Video",
```

Quite powerful and the community loves it, Mochi-1 is a bit too VRAM hungry

Vaibhavs10 (commented):
NIceee! Left some suggestions!

packages/tasks/src/tasks/text-to-video/data.ts
merveenoyan: this page is for text-to-video, that model is added to image-to-video already

packages/tasks/src/tasks/audio-to-audio/data.ts
```diff
@@ -39,8 +39,8 @@ const taskData: TaskDataCustom = {
 	],
 	models: [
 		{
-			description: "A solid model of audio source separation.",
-			id: "speechbrain/sepformer-wham",
+			description: "A robust model to compress speech.",
+			id: "stabilityai/stable-codec-speech-16k",
```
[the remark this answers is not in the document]
merveenoyan: I also thought the same but it was under audio-to-audio tag so I rolled with it 😄

packages/tasks/src/tasks/text-to-video/data.ts
Vaibhavs10: But, LTX-Video is Text to Video: https://huggingface.co/Lightricks/LTX-Video 😅

Vaibhavs10 (approved):
Thanks for iterating! <3

Vaibhavs10 (commented):
Oh and make sure to fix the lint error before merge 🙏

packages/tasks/src/tasks/text-to-video/data.ts
merveenoyan: most of the models that are image-to-video are actually image+text-to-video, so most of the time the differentiator here is the fact that it can take in an image as initial frame or video as few initial frames. I have no strong opinion here but I find it confusing it redirects to a model tagged as image-to-video

julien-c:
Note: in the coming weeks, IMO we can update the task pages to show some examples on how to run inference on Third-party Inference providers as well (from both JS or Python) cc @SBrandeis

PR #1102, opened 2025-01-14
add more Replicate models

zeke:
This is a work in progress PR to expand the list of Replicate models.

Every model in this list has over 10,000 runs on Replicate.

## To Do

- [x] Sort this out: https://github.com/huggingface/huggingface.js/pull/1106

### DIFF
--- a/packages/inference/src/providers/replicate.ts
+++ b/packages/inference/src/providers/replicate.ts
@@ -24,51 +24,74 @@ export const REPLICATE_MODEL_IDS: Record<ModelId, ReplicateId> = {
   "black-forest-labs/FLUX.1-Fill-dev": "black-forest-labs/flux-fill-dev",
   "black-forest-labs/FLUX.1-Redux-dev": "black-forest-labs/flux-redux-dev",
   "black-forest-labs/FLUX.1-schnell": "black-forest-labs/flux-schnell",
-  "ByteDance/Hyper-SD": "bytedance/hyper-flux-16step",
+  "ByteDance/Hyper-SD":
+    "bytedance/hyper-flux-16step:382cf8959fb0f0d665b26e7e80b8d6dc3faaef1510f14ce017e8c732bb3d1eb7",
   "ByteDance/SDXL-Lightning":
     "bytedance/sdxl-lightning-4step:5599ed30703defd1d160a25a63321b4dec97101d98b4674bcc56e41f62f35637",
-  "dataautogpt3/ProteusV0.3": "datacte/proteus-v0.3",
-  "Efficient-Large-Model/Sana_1600M_1024px": "nvidia/sana",
-  "fal/AuraFlow": "fofr/aura-flow",
-  "KwaiVGI/LivePortrait": "fofr/expression-editor",
+  "dataautogpt3/ProteusV0.3":
+    "datacte/proteus-v0.3:b28b79d725c8548b173b6a19ff9bffd16b9b80df5b18b8dc5cb9e1ee471bfa48",
+  "Efficient-Large-Model/Sana_1600M_1024px":
+    "nvidia/sana:c6b5d2b7459910fec94432e9e1203c3cdce92d6db20f714f1355747990b52fa6",
+  "fal/AuraFlow":
+    "fofr/aura-flow:ae5ab66a7d1ca7ee44cf8c50265d3bafdef23734d03d66063d1c8fcf82f0c17b",
+  "KwaiVGI/LivePortrait":
+    "fofr/expression-editor:bf913bc90e1c44ba288ba3942a538693b72e8cc7df576f3beebe56adc0a92b86",
   "playgroundai/playground-v2.5-1024px-aesthetic":
-    "playgroundai/playground-v2.5-1024px-aesthetic",
-  "SG161222/Realistic_Vision_V5.1_noVAE": "lucataco/realistic-vision-v5.1",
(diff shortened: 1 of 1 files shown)


packages/inference/src/providers/replicate.ts
```diff
@@ -13,9 +13,55 @@ type ReplicateId = string;
  * 	-H "Authorization: Bearer $REPLICATE_API_TOKEN" \
  * 	'https://api.replicate.com/v1/models'
  * ```
+ * 
+ * See also https://github.com/replicate/all-the-public-replicate-models
```
julien-c, line 17: 👍

packages/inference/src/providers/replicate.ts
```diff
@@ -13,9 +13,55 @@ type ReplicateId = string;
...
+	"SG161222/Realistic_Vision_V5.1_noVAE": "lucataco/realistic-vision-v5.1",
+	"SG161222/RealVisXL_V3.0": "fofr/realvisxl-v3-multi-controlnet-lora",
+	"stabilityai/stable-diffusion-3.5-large-turbo": "stability-ai/stable-diffusion-3.5-large-turbo",
+	"stabilityai/stable-diffusion-3.5-large": "stability-ai/stable-diffusion-3.5-large",
+	"stabilityai/stable-diffusion-3.5-medium": "stability-ai/stable-diffusion-3.5-medium",
+	"stabilityai/stable-diffusion-xl-base-1.0": "stability-ai/sdxl",
+	"XLabs-AI/flux-controlnet-collections": "xlabs-ai/flux-dev-controlnet",
+
+	/** image tools */
+	"Carve/tracer_b7": "lucataco/remove-bg",
```
julien-c, line 43: this HF model page doesn't have an identified "pipeline type" but we could open a PR to add one (image-to-image?)

BTW our list of tasks is here: https://huggingface.co/tasks

packages/inference/src/providers/replicate.ts
```diff
@@ -13,9 +13,55 @@ type ReplicateId = string;
...
+	"Falconsai/nsfw_image_detection": "falcons-ai/nsfw_image_detection",
```
julien-c, line 44: this HF model page is identified as image-classification

packages/inference/src/providers/replicate.ts
```diff
@@ -13,9 +13,55 @@ type ReplicateId = string;
...
+
+	/** language */
+	"ibm-granite/granite-3.0-2b-instruct": "ibm-granite/granite-3.0-2b-instruct",
+	"ibm-granite/granite-3.0-8b-instruct": "ibm-granite/granite-3.0-8b-instruct",
```
julien-c, line 48: those HF model pages are identified as "text-generation pipeline" so i think the Replicate APIs could work

packages/inference/src/providers/replicate.ts
```diff
@@ -13,9 +13,55 @@ type ReplicateId = string;
...
+
+	/** vision */
+	"allenai/Molmo-7B-D-0924": "zsxkib/molmo-7b",
```
julien-c, line 51: [Image-Text-to-Text](https://huggingface.co/tasks/image-text-to-text)

julien-c (commented):
awesome ❤️ 
one thing to take into account is that each HF model page is (automatically or manually) identified as one `pipeline type` aka. `task type`, and then the API schema is conditioned by this pipeline type

I haven't looked in depth, but some of the Replicate endpoints might very well not have an API that's conforming to the HF pipeline type. In those cases we should exclude them from the list because the widget won't work.

cc @SBrandeis for more details:)

packages/inference/src/providers/replicate.ts
zeke: image-to-image makes sense to me. 👍🏼

SBrandeis:
Cool stuff @zeke !

Now that #1106 is merged to the main branch, would you mind rebasing this please?

packages/inference/src/providers/replicate.ts
```diff
@@ -13,9 +13,57 @@ type ReplicateId = string;
...
+	"genmo/mochi-1-preview": "genmoai/mochi-1",
+	"Lightricks/LTX-Video": "lightricks/ltx-video",
+	"rain1011/pyramid-flow-sd3": "zsxkib/pyramid-flow",
+	"tencent/HunyuanVideo": "tencent/hunyuan-video",
+
+	/** audio */
+	"hkchengrex/MMAudio/tree/main": "zsxkib/mmaudio",
+	"ICTNLP/Llama-3.1-8B-Omni": "ictnlp/llama-omni",
+
+	/** speech */
```
julien-c, line 66:
```suggestion
	/** automatic-speech-recognition */
```

SBrandeis:
Thank you @zeke , very exciting to see more models in the mapping !!

Following @julien-c 's initial review comment: can you make sure that all the added models' APIs conform the pipeline tag from huggingface? (it's listed in the model info API, example: https://huggingface.co/api/models/black-forest-labs/FLUX.1-dev?expand[]=pipeline_tag)
Otherwise, users won't be able to use Replicate through our Inference Widget on the model pages.

zeke:
The model list in this PR is too amibitious, so I'm gonna close it. Too many modalities to try to support all at once.

Instead we can approach this incrementally. Here's a new PR that adds a handful more models, with tests: https://github.com/huggingface/huggingface.js/pull/1130

PR #1104, opened 2025-01-14
Add fal.ai inference endpoints

gokayfem:
Adds fal.ai inference endpoints


julien-c:
Hey @gokayfem in https://github.com/huggingface/huggingface.js/pull/1109 @SBrandeis has switched to a dict of mappings, indexed by the HF model's `task type` (which in turn, conditions the API schema we expect)

Can you rebase your PR on top of main? Also happy to jump on a quick call if it's easier!

SBrandeis (approved):
Looks good!

julien-c (approved):

PR #1107, opened 2025-01-15
[Inference] Factor makeRequestOptions logic

SBrandeis:
- Update the `makeRequestOptions` function to make the logic (hopefully) more readable; especially since #1077 
- Stop support of model URLs
- Update tapes


packages/inference/src/lib/makeRequestOptions.ts
```diff
@@ -31,62 +30,46 @@ export async function makeRequestOptions(
...
+	if (forceTask && provider !== "hf-inference") {
+		throw new Error(`Cannot use forceTask with a third-party provider.`);
 	}
 
-	if (!model && tasks && taskHint) {
-		const taskInfo = tasks[taskHint];
-		if (taskInfo) {
-			model = taskInfo.models[0].id;
+	let model: string;
+	if (!maybeModel) {
```
coyotte508, line 47: Maybe add something like

```ts
if (maybeModel && URL.parse(maybeModel)) {
  throw new TypeError("model URLs are no longer supported, use endpointUrl instead")
}
```

coyotte508 (approved):

julien-c (approved):

PR #1111, opened 2025-01-17
Add `@huggingface/ollama-utils`

ngxson:
With the ollama compatibility layer already been up and running on HF hub for a while now, we want to open-source part of the integration for (1) provide more transparency, and (2) for encouraging the community to contribute.

Therefore, we decided to publish `@huggingface/ollama-utils`, a package containing tools that power this integration.

For now, the only tool that we provide is `chat-template.ts`, a tool that takes a parsed GGUF config from `@huggingface/gguf` as input, then returns the ollama Go template.

# chat-template.ts

This module expose one single `convertGGUFTemplateToOllama` function. It works by trying (by order) these mechanisms:
1. Try to find a matched template already exist on ollama hub. To make this possible, we regularly dump a list of public templates from ollama.com using `scripts/generate-automap.ts`
2. If no exact match is find, we still check against the list of official templates, but now only match a list of special tokens. These tokens are extracted using regex `RE_SPECIAL_TOKEN`
3. If we still can't find a match, we search inside `CUSTOM_TEMPLATE_MAPPING`, a list of hand-picked template mapping
4. If all above still failed, we try parsing jinja template and try converting it to Go template. This works almost all of the time, except if the template is not supported by `@huggingface/jinja`. See `convertJinjaToGoTemplate`


packages/ollama-utils/README.md
```diff
@@ -0,0 +1,11 @@
+# `@huggingface/ollama-utils`
+
+Various utilities for maintaining Ollama compatibility with models on Hugging Face hub.
+
+Documentation: https://huggingface.co/docs/hub/en/ollama
+
+## Chat template converter
+
+```ts
+const
```
ngxson, line 10: TODO: complete this @ngxson

packages/ollama-utils/tsup.config.ts
coyotte508, line 1: If you feel up to it you can use `tshy` instead, like in `@hugginface/tasks`.

It has more accurate exports / better compatibility with all that includes the package (and as long as you don't have different code in browser/node it should be easy to configure)

coyotte508 (approved):

coyotte508:
btw the github action to publish should be added (with optionnally a check to check the `@huggingface/jinja` is up to date, like `@huggingface/hub` does with `@huggingface/tasks`)

julien-c (approved):
neat!

Vaibhavs10 (commented):
don't know enough typescript to review, but two questions/comments from myside:
1. maybe in a follow-up PR we should add a scraper and something to update the mapping?
2. via comments/ README/ docs we should clarify how users can  add their own exception as well

packages/ollama-utils/tsup.config.ts
ngxson: I can't get around the error with this line:

```
src/chat-template.ts:1:43 - error TS1479: The current file is a CommonJS module whose imports will produce 'require' calls; however, the referenced file is an ECMAScript module and cannot be imported with 'require'. Consider writing a dynamic 'import("@huggingface/jinja")' call instead.
  To convert this file to an ECMAScript module, change its file extension to '.mts' or create a local package.json file with `{ "type": "module" }`.

1 import { Template as JinjaTemplate } from "@huggingface/jinja";
                                            ~~~~~~~~~~~~~~~~~~~~
```

Do you have any ideas @coyotte508 ?

Vaibhavs10 (approved):
yay! conceptually lgtm!

packages/ollama-utils/README.md
```diff
@@ -0,0 +1,56 @@
...
+   */
+} else {
+  console.error("Conversion is not successful");
+}
+```
+
+## How can I add a custom template?
+
+Most templates will be converted automatically. You can debug the output template using:
+- This space to retrieve the converted template: https://huggingface.co/spaces/ngxson/debug_ollama_manifest
```
Vaibhavs10, line 45: maybe we can do both in the same space?

packages/ollama-utils/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+Please only add a new template only when the conversion above is not successful. Cases that are acceptable to add a custom handler:
+- The covnerted template is wrong
+- The Jinja template is not compatible with `@huggingface/jinja`
+- The Jinja template is not "linear"
+
+To add a new custom handler:
+1. Edit the list of `CUSTOM_TEMPLATE_MAPPING` inside `chat-template.ts`
+2. Add a new test case in `chat-template.spec.ts`
+3. hen push your change into new PR.
```
Vaibhavs10, line 56:
```suggestion
3. Push your change into new PR.
```

packages/ollama-utils/README.md
```diff
@@ -0,0 +1,56 @@
...
+```
+
+## How can I add a custom template?
+
+Most templates will be converted automatically. You can debug the output template using:
+- This space to retrieve the converted template: https://huggingface.co/spaces/ngxson/debug_ollama_manifest
+- And this space to apply the Go template into a list of messages: https://huggingface.co/spaces/ngxson/ollama_template_test
+
+Please only add a new template only when the conversion above is not successful. Cases that are acceptable to add a custom handler:
+- The covnerted template is wrong
```
Vaibhavs10, line 49:
```suggestion
- The converted template is wrong
```

packages/ollama-utils/README.md
```diff
@@ -0,0 +1,56 @@
...
+- The Jinja template is not compatible with `@huggingface/jinja`
+- The Jinja template is not "linear"
```
Vaibhavs10, line 51: maybe add a quick sentence to explain what `linear` means

packages/ollama-utils/README.md
ngxson: I couldn't figure out how to embed my script into gradio space, but will have a look later

(The template debugging space is static btw)

packages/ollama-utils/tsup.config.ts
xenova: Strange, the exports should operate correctly
image

Maybe I need to restructure it like transformers.js does it [here](https://github.com/huggingface/transformers.js/blob/6f43f244e04522545d3d939589c761fdaff057d4/package.json#L8-L23).

xenova: Do any other hf.js libs use tshy?

Vaibhavs10 (commented):
merge + comms - today?

packages/ollama-utils/README.md
```diff
@@ -0,0 +1,56 @@
+# `@huggingface/ollama-utils`
+
+Various utilities for maintaining Ollama compatibility with models on Hugging Face hub.
+
+Documentation: https://huggingface.co/docs/hub/en/ollama
+
```
pcuenca, line 6:
```suggestion
```

pcuenca (commented):
Looks good, suggesting a few nits

packages/ollama-utils/tsup.config.ts
ngxson: @xenova FYI I took `tasks` as my example and follow its implementation

PR #1116, opened 2025-01-20
[Inference] Generate binary types as `Blob` and use the generated types in the inference client

SBrandeis:
# TL;DR

- Use Typescript interfaces generated from `@huggingface/tasks` JSON schemas to type the input types of the Inference client
- try to maintain as much backwards-compatibility as possible with the existing inference client API

### DIFF
--- a/packages/inference/src/tasks/audio/textToSpeech.ts
+++ b/packages/inference/src/tasks/audio/textToSpeech.ts
@@ -1,15 +1,9 @@
+import type { TextToSpeechInput, TextToSpeechOutput } from "@huggingface/tasks";
 import { InferenceOutputError } from "../../lib/InferenceOutputError";
 import type { BaseArgs, Options } from "../../types";
 import { request } from "../custom/request";
 
-export type TextToSpeechArgs = BaseArgs & {
-	/**
-	 * The text to generate an audio from
-	 */
-	inputs: string;
-};
-
-export type TextToSpeechOutput = Blob;
+type TextToSpeechArgs = BaseArgs & TextToSpeechInput;
 
 /**
  * This task synthesize an audio of a voice pronouncing a given text.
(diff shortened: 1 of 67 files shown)


packages/inference/src/tasks/audio/textToSpeech.ts
```diff
@@ -18,26 +13,26 @@ interface OutputUrlTextToSpeechGeneration {
  * Recommended model: espnet/kan-bayashi_ljspeech_vits
  */
 export async function textToSpeech(args: TextToSpeechArgs, options?: Options): Promise<TextToSpeechOutput> {
-	const res = await request<TextToSpeechOutput | OutputUrlTextToSpeechGeneration>(args, {
+	const res = await request<Blob | OutputUrlTextToSpeechGeneration>(args, {
 		...options,
 		taskHint: "text-to-speech",
 	});
+	console.log(res)
```
julien-c, line 20:
```suggestion
```

packages/tasks/src/tasks/text-to-audio/spec/output.json
```diff
@@ -6,12 +6,13 @@
 	"type": "object",
 	"properties": {
 		"audio": {
-			"description": "The generated audio waveform."
+			"description": "The generated audio waveform.",
+			"comment": "type=binary"
```
SBrandeis, line 10: cc @Wauplin @hanouticelina 

I used this trick to mark a parameter as "binary" (which translates, in the generated Typescript code, to declaring it to be of type `Blob`)

Generation code: https://github.com/huggingface/huggingface.js/pull/1116/files#diff-3a9454ed7c359e2e0522703d9acf4559f6b10fdd67236692bdd4422dd9c8f472R97-R182

julien-c (approved):
haven't dived in detail but the compatibility plan LGTM @SBrandeis

coyotte508 (approved):

PR #1120, opened 2025-01-20
Add image-feature-extraction task to transformers and timm lib->task mapping

rwightman:
There's no pipeline snippet for models w/ the image-feature-extraction task, probably because this mapping is missing?

cc @pcuenca re slack comment

### DIFF
--- a/packages/tasks/src/library-to-tasks.ts
+++ b/packages/tasks/src/library-to-tasks.ts
@@ -43,7 +43,7 @@ export const LIBRARY_TASK_MAPPING: Partial<Record<ModelLibraryKey, PipelineType[
 		"text2text-generation",
 	],
 	stanza: ["token-classification"],
-	timm: ["image-classification"],
+	timm: ["image-classification", "image-feature-extraction"],
 	transformers: [
 		"audio-classification",
 		"automatic-speech-recognition",
@@ -52,6 +52,7 @@ export const LIBRARY_TASK_MAPPING: Partial<Record<ModelLibraryKey, PipelineType[
 		"feature-extraction",
 		"fill-mask",
 		"image-classification",
+		"image-feature-extraction",
 		"image-segmentation",
 		"image-to-image",
 		"image-to-text",


rwightman:
See e.g.
https://huggingface.co/facebook/dinov2-large?library=transformers
https://huggingface.co/timm/vit_large_patch14_dinov2.lvd142m?library=transformers

pcuenca (commented):
The intent here is to enable the pipeline snippet in the "Use this model" button, and I think it'll work for that purpose. However, I'm not sure whether this will attempt to show an inference widget as well, along with the curl / js snippets. What do you think @SBrandeis?

The image-feature-extraction task is supported for inference [here](https://github.com/huggingface-internal/api-inference/blob/main/master-rs/src/data/hf_types.rs#L850), but not [here](https://github.com/huggingface-internal/api-inference/blob/b055f0c771ed60e4e0f786c096eccf317e0ef471/shard/app/model.py#L210). The following `curl` command hence fails with `{"error":"image-feature-extraction is not a valid pipeline"}`

```bash
curl https://api-inference.huggingface.co/models/facebook/dinov2-large \ 
        -X POST \
        --data-binary '@example.jpg' \
        -H "Authorization: Bearer $TOKEN"
```

Also cc @merveenoyan @NielsRogge

Vaibhavs10 (commented):
for snippets we'd need to add the task here as well AFAIR: https://github.com/huggingface/huggingface.js/tree/44e9e02cee0a3cb9d6e89b329f8c2d60b458e480/packages/tasks/src/snippets

pcuenca:
> for snippets we'd need to add the task here as well AFAIR: 

That's for inference snippets / widgets IIUC. For "Use this model", would the current state of the PR work with no impact elsewhere? (i.e., no widget enabled)

julien-c:
"Use this model" snippets are defined in https://github.com/huggingface/huggingface.js/blob/44e9e02cee0a3cb9d6e89b329f8c2d60b458e480/packages/tasks/src/model-libraries.ts AFAIK

pcuenca:
> "Use this model" snippets are defined in https://github.com/huggingface/huggingface.js/blob/44e9e02cee0a3cb9d6e89b329f8c2d60b458e480/packages/tasks/src/model-libraries.ts AFAIK

Yes! The issue is that `image-feature-extraction` models don't have a `pipeline` snippet because of [this](https://github.com/huggingface/huggingface.js/blob/44e9e02cee0a3cb9d6e89b329f8c2d60b458e480/packages/tasks/src/model-libraries-snippets.ts#L878).

i.e, this is an `image-feature-extraction` model (AutoModel, but no pipeline snippet)
Screenshot 2025-01-21 at 14 56 23

vs an `image-classification` one (AutoModel + pipeline snippets):
Screenshot 2025-01-21 at 14 56 45


My concern is whether updating `LIBRARY_TASK_MAPPING` would have any side effects (i.e., widgets being enabled).

julien-c:
worst case we merge and we'll check in prod:)

merveenoyan (approved):
I think this PR should work. I'm a bit late to party, sorry. 

I think adding model-libraries.ts will not matter because it's mostly for third party libraries and already defined for transformers. There are two issues here: 
1. transformers `image-feature-extraction` models don't have pipeline snippets. transformers models are maintained separately AFAIK
them having `pipeline_tag` as `image-feature-extraction` won't matter either e.g. [this one](https://huggingface.co/nomic-ai/nomic-embed-vision-v1.5) has the tag and also have the model class properly but doesn't have snippet mapped (this PR should solve that). 

I have overlooked it, thanks for fixing. I will add other vision tasks in a separate PR.

2. timm models likewise doesn't have timmwrapper to load with transformers, which I'm surprised because they're inferred from [here](https://huggingface.co/datasets/huggingface/transformers-metadata/blob/main/pipeline_tags.json#L1118) and `TimmWrapper` exists there. @rwightman opened another PR for it it seems

merveenoyan:
enabled the same for VLMs with recently merged `image-text-to-text` pipeline [here](https://github.com/huggingface/huggingface.js/pull/1126)

pcuenca:
Update: it worked as expected: https://huggingface.co/facebook/dinov2-large?library=transformers now has pipeline and model usage snippets.

PR #1121, opened 2025-01-21
[Inference] Better types for HfInference

SBrandeis:
(unfinished, but I would love some inputs)

cc @coyotte508 

Before this PR, Typescript was unable to determine actual types for the inputs of methods on the `HfInference` client 

`HfInference.textToImage` would lack typings for `parameters` for example

This PR ensures that all tasks that are exported in `src/tasks/index` are re-exported by the client `HfInference`, and introduces a helper function that keeps every task's types along the way

### DIFF
--- a/packages/inference/package.json
+++ b/packages/inference/package.json
@@ -55,8 +55,7 @@
 		"@huggingface/tasks": "workspace:^"
 	},
 	"devDependencies": {
-		"@types/node": "18.13.0",
-		"type-fest": "^3.13.1"
+		"@types/node": "18.13.0"
 	},
 	"resolutions": {}
 }
--- a/packages/inference/pnpm-lock.yaml
+++ b/packages/inference/pnpm-lock.yaml
@@ -13,17 +13,9 @@ devDependencies:
   '@types/node':
     specifier: 18.13.0
     version: 18.13.0
-  type-fest:
-    specifier: ^3.13.1
-    version: 3.13.1
 
 packages:
 
   /@types/node@18.13.0:
     resolution: {integrity: sha512-gC3TazRzGoOnoKAhUx+Q0t8S9Tzs74z7m0ipwGpSqQrleP14hKxP4/JUeEQcD3W1/aIpnWl8pHowI7WokuZpXg==}
-    dev: true
-
-  /type-fest@3.13.1:
-    resolution: {integrity: sha512-tLq3bSNx+xSpwvAJnzrK0Ep5CLNWjvFTOp71URMaAEWBfRb9nnJiBoUe0tF8bI4ZFO3omgBR6NvnbzVUT3Ly4g==}
-    engines: {node: '>=14.16'}
     dev: true
--- a/packages/inference/src/utils/typedEntries.ts
+++ b/packages/inference/src/utils/typedEntries.ts
@@ -1,5 +1,0 @@
-import type { Entries } from "type-fest";
-
-export function typedEntries<T extends { [s: string]: T[keyof T] } | ArrayLike<T[keyof T]>>(obj: T): Entries<T> {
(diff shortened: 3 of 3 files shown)


packages/inference/test/HfInference.spec.ts
```diff
@@ -589,18 +589,18 @@ describe.concurrent("HfInference", () => {
...
+			// 		await hf.request({
+			// 			model: "openai-community/gpt2",
+			// 			inputs: "one plus two equals",
+			// 		})
+			// 	).toMatchObject([
+			// 		{
+			// 			generated_text: expect.any(String),
+			// 		},
+			// 	]);
+			// });
```
SBrandeis, line 603: Still need to find a way around `request` and `streamingRequest`... They use a generic type `T` which messes thing up I think

packages/inference/src/HfInference.ts
```diff
@@ -21,22 +21,83 @@ type TaskWithNoAccessTokenNoEndpointUrl = {
 	) => ReturnType<Task[key]>;
 };
 
-export class HfInference {
-	private readonly accessToken: string;
-	private readonly defaultOptions: Options;
+export class HfInference implements Task {
```
SBrandeis, line 24:
```typescript
export class HfInference implements Task {
```

That bit makes sure `HfInference` defines a method for every task (`textToImage`, `chatCompletion`, etc)

packages/inference/src/HfInference.ts
```diff
@@ -21,22 +21,83 @@ type TaskWithNoAccessTokenNoEndpointUrl = {
...
+	tabularRegression: typeof tasks.tabularRegression;
+	tabularClassification: typeof tasks.tabularClassification;
+	textGenerationStream: typeof tasks.textGenerationStream;
+	chatCompletionStream: typeof tasks.chatCompletionStream;
+
+	static mapInferenceFn<TOut, TArgs>(instance: HfInference, func: (...args: [TArgs, Options?]) => TOut) {
+		return function (...[args, options]: Parameters<(...args: [TArgs, Options?]) => TOut>): TOut {
+			return func({ ...args, accessToken: instance.accessToken }, { ...instance.defaultOptions, ...(options ?? {}) });
+		};
+	}
```
SBrandeis, line 63: This helper functions adds some default arguments when calling `HfInference.<task>()` while keeping the exact type for inputs and outputs

packages/inference/src/tasks/index.ts
```diff
@@ -40,3 +42,5 @@ export * from "./multimodal/visualQuestionAnswering";
 // Tabular tasks
 export * from "./tabular/tabularRegression";
 export * from "./tabular/tabularClassification";
+
+export const speechToText = automaticSpeechRecognition;
```
SBrandeis, line 46: (unrelated) aliasing `automaticSpeechRecognition` as `speechToText`

julien-c: did it exist before or that's new?

julien-c: (why not i guess, though we should try to push the "official" task names otherwise it's hard to standardize one way of doing things)

SBrandeis: > (why not i guess, though we should try to push the "official" task names otherwise it's hard to standardize one way of doing things)

yeah, good point

coyotte508 (commented):
Do you want to edit `HfInferenceEndpoint` too in the same file?

Also note the script `packages/inference/scripts/generate-dts.ts` used for generating definition files, you may need to change it now

(and we could switch from tsup to tshy removing the need for the script entirely maybe)

julien-c:
i suspect this might need to be closed now

PR #1129, opened 2025-01-22
[Inference] Update snippets

SBrandeis:


packages/agents/pnpm-lock.yaml
julien-c, line 1: not sure if intended

packages/tasks/src/snippets/js.ts
```diff
@@ -78,9 +93,11 @@ for await (const chunk of stream) {
 	}  
 }`,
 				},
-				{
-					client: "openai",
-					content: `import { OpenAI } from "openai";
+				...(provider === "hf-inference"
+					? [
```
julien-c, line 97: hmm my brain is friend but i think we can still support this for external providers (it's the same API)

I can do it if you want @SBrandeis

packages/tasks/src/inference-providers.ts
julien-c, line 1: not sure what we should do with this... just duplicate in two places for now? i guess it's fine

julien-c (commented):
i would maybe add one example of how a third-party provider can define their own snippet for one particular task

packages/tasks/src/inference-providers.ts
SBrandeis: I was planning to have `inference` depend on `task` for this

packages/agents/pnpm-lock.yaml
SBrandeis: No - automatic linting I think

packages/tasks/src/snippets/js.ts
SBrandeis: Ah yes, feel free, to give it a try!

packages/tasks/src/inference-providers.ts
julien-c: makes it a bit more coupled i.E for now we could just tell providers to contribute to /inference

SBrandeis: ok!

packages/tasks/src/snippets/js.ts
julien-c: 8aafe2562642db47cc2c8e38b6f2deadf711c81b

packages/tasks/src/snippets/js.ts
```diff
@@ -141,36 +189,68 @@ console.log(chatCompletion.choices[0].message);`,
...
-	)}, "parameters": {"candidate_labels": ["refund", "legal", "faq"]}}).then((response) => {
-	console.log(JSON.stringify(response));
-});`,
+export const snippetTextToImage = (
+	model: ModelDataMinimal,
+	accessToken: string,
+	provider: InferenceProvider
+): InferenceSnippet[] => {
+	return [
+		{
```
SBrandeis, line 231: Should we keep this?
If yes, let's add install instructions to other snippets

packages/tasks/src/index.ts
```diff
@@ -58,3 +58,5 @@ export type { LocalApp, LocalAppKey, LocalAppSnippet } from "./local-apps.js";
 
 export { DATASET_LIBRARIES_UI_ELEMENTS } from "./dataset-libraries.js";
 export type { DatasetLibraryUiElement, DatasetLibraryKey } from "./dataset-libraries.js";
+
+export * from "./inference-providers.js";
```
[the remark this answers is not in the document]
SBrandeis: Ok did not see the import from `tasks-gen`

Wauplin (commented):
:+1: (only checked Python snippets)

packages/agents/pnpm-lock.yaml
```diff
@@ -7,7 +7,7 @@ settings:
 dependencies:
   '@huggingface/inference':
     specifier: ^2.6.1
-    version: link:../inference
+    version: 2.8.1
```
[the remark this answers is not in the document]
julien-c: it wasn't me, but yes i think extraneous

packages/tasks/src/snippets/js.ts
julien-c: yes fine to remove imo

PR #1140, opened 2025-01-24
add gtx 1060

nroggendorff:
why wasn't this here before?


julien-c (approved):
well it's a collector item at this point 😁

pcuenca (approved):

PR #1145, opened 2025-01-28
[together provider] Update Meta Llama 3 8B

pcuenca:
Taken from https://api.together.ai/models/meta-llama/Meta-Llama-3-8B-Instruct-Turbo


hanouticelina (approved):

julien-c (approved):
lgtm, but let's make sure we ping the Together team on slack so they're aware

PR #1149, opened 2025-01-28
fix(inference/fal-ai): b64 response format argument

badayvedat:
`fal-ai` endpoints are using `sync_mode: true` argument to receive the result as a **base64 data uri**
here is an example input/output with the parameter: https://fal.ai/models/fal-ai/flux/dev?share=c1793354-3994-44ef-a50b-31ef2ee9c90f


julien-c:
launching the CI to see if it passes the existing `fal` tests

SBrandeis:
@badayvedat can you update the tapes.json file, please? It's used to replay HTTP responses in the CI.
You can do so by running test with `VCR_MODE=cache`:

```
VCR_MODE=cache pnpm run test
```

badayvedat:
thanks @SBrandeis ! updated them
fyi the new response objects are base64 images - which increased the file size of tapes.json significantly -

SBrandeis (approved):
Nice, thank you @badayvedat !

PR #1150, opened 2025-01-28
add user-agent header to inference requests

zeke:
This PR updates the inference client to include a user-agent header in outgoing API requests.

The value is derived from the `name` and `version` of the inference package, e.g. `@huggingface/inference@3.1.3`

This will allow third-party inference providers to identify API requests that originate from HF's new inference integration.

See https://huggingface.co/blog/inference-providers

cc @julien-c @SBrandeis

### DIFF
--- a/packages/inference/src/lib/makeRequestOptions.ts
+++ b/packages/inference/src/lib/makeRequestOptions.ts
@@ -7,6 +7,7 @@ import { TOGETHER_API_BASE_URL, TOGETHER_SUPPORTED_MODEL_IDS } from "../provider
 import type { InferenceProvider } from "../types";
 import type { InferenceTask, Options, RequestArgs } from "../types";
 import { isUrl } from "./isUrl";
+import { version as packageVersion, name as packageName } from "../../package.json" assert { type: "json" };
 
 const HF_HUB_INFERENCE_PROXY_TEMPLATE = `${HF_HUB_URL}/api/inference-proxy/{{PROVIDER}}`;
 
@@ -88,6 +89,9 @@ export async function makeRequestOptions(
 		headers["Authorization"] =
 			provider === "fal-ai" && authMethod === "provider-key" ? `Key ${accessToken}` : `Bearer ${accessToken}`;
 	}
+
+	// e.g. @huggingface/inference@3.1.3
+	headers["User-Agent"] = `${packageName}@${packageVersion}`;
 
 	const binary = "data" in args && !!args.data;
 


zeke:
Your requests will always originate from a server proxy, right? (i.e., not a browser?)

I heard from my teammate @aron recently that there's a [Chrome bug](https://issues.chromium.org/issues/40450316] that prevents setting the user agent... Just want to be sure that's not going to block this.

coyotte508:
Hi @zeke , thanks for the PR!

Can you use `navigator.userAgent` (if `typeof navigator !== "undefined"`) to append our user-agent to it and not replace?

Also when targetting `huggingface.co`, we'll probably add a `X-UserAgent` header + CORS authorizations for it to bypass the chrome bug, eg the lib calls our proxy, and our proxy sets the `UserAgent` header to the value of `X-UserAgent` when calling the backend service

packages/inference/src/lib/makeRequestOptions.ts
```diff
@@ -89,6 +90,9 @@ export async function makeRequestOptions(
 			provider === "fal-ai" && authMethod === "provider-key" ? `Key ${accessToken}` : `Bearer ${accessToken}`;
 	}
 
+	// e.g. @huggingface/inference@3.1.3
+	headers["User-Agent"] = `${packageName}@${packageVersion}`;
```
coyotte508, line 94:
```suggestion
	headers["User-Agent"] = `${packageName}/${packageVersion}`;
```

To harmonize with `huggingface_hub` cc @Wauplin @hanouticelina

julien-c: for the package you can just use hardcoded `huggingface.js/{version_number}` IMO (potentially this would apply to all calls from all hf.js subpackages, no need to special-case inference)

coyotte508: we have different versions for different sub-pacakges though :(

maybe `huggingface.js;@huggingface/{subPackage}/{version}`

julien-c (approved):
<3

coyotte508:
Published `@huggingface/inference@3.1.4-test` to test it & make sure it works fine

coyotte508 (approved):
Just in case @Wauplin or @SBrandeis want to take a look before merging

packages/inference/src/lib/makeRequestOptions.ts
```diff
@@ -89,6 +90,12 @@ export async function makeRequestOptions(
 			provider === "fal-ai" && authMethod === "provider-key" ? `Key ${accessToken}` : `Bearer ${accessToken}`;
 	}
 
+	// e.g. @huggingface/inference@3.1.3
```
SBrandeis, line 93:
```suggestion
	// e.g. @huggingface/inference/3.1.3
```

SBrandeis (approved):
LGTM! Thank you

zeke:
Thanks helping get this over the line so quickly, everyone. 🙌🏼

PR #1155, opened 2025-01-30
Add Mistral Small 3 to list of supported Together models

Nutlope:


julien-c (approved):
🔥 

cc @AK391 for viz

PR #1164, opened 2025-02-03
Add SDXL to fal supported models

o-toff:
Gave it a spin with:
```ts
import fs from "fs";

const hf = new HfInference("hf_xxxx");

async function callFalAiModel() {
	try {
		const result = await hf.textToImage({
			provider: "fal-ai",
			model: "stabilityai/stable-diffusion-xl-base-1.0",
			inputs: "A cat in the shape of a hug smiley",
		});
		console.log(result);

		const blob = new Blob([result], { type: "image/jpeg" });
		const arrayBuffer = await blob.arrayBuffer();
		const buffer = Buffer.from(arrayBuffer);

		fs.writeFileSync("output.jpg", buffer);
		console.log("Image saved to output.jpg");
	} catch (error) {
		console.error("Error calling fal-ai/fast-sdxl model:", error);
	}
}

callFalAiModel();
```

got
output


julien-c (approved):
nice!

SBrandeis (approved):

julien-c:
thx a lot!

PR #1167, opened 2025-02-03
[Tasks] update automatic speech recognition output specs

hanouticelina:
It should have been `timestamp` instead of `timestamps` (see transformers pipeline output [here](https://github.com/huggingface/transformers/blob/9d2056f12b66e64978f78a2dcb023f65b2be2108/src/transformers/pipelines/automatic_speech_recognition.py#L648)). this also aligns with fal-ai output [specs](https://fal.ai/models/fal-ai/whisper/api#type-WhisperChunk) of the model which is currently supported.


julien-c (approved):

Wauplin (approved):
Thanks! 

For completeness, I've checked OpenAI's speech-to-text API ([here](https://platform.openai.com/docs/api-reference/audio/createTranscription)) but format is completely different so let's not try to align.

PR #1190, opened 2025-02-06
:sparkles: Integrate Nebius AI Studio into HuggingFace inference

kitcat-dev:
Added support for [the inference service from Nebius](https://nebius.com/services/studio-inference-service).

I have added tests similar to others in `HfInference.spec.ts` file. They were passing just yesterday – before hardcode-models were removed. I'll have to double-check when our models will be dynamically mapped with your partners API.


packages/inference/src/lib/makeRequestOptions.ts
```diff
@@ -144,6 +145,7 @@ export async function makeRequestOptions(
 			: JSON.stringify({
 					...otherArgs,
 					...(chatCompletion || provider === "together" ? { model } : undefined),
+					...(provider === "nebius" ? { model } : {}),
```
julien-c, line 148: i'll add to the same line above for simplicity

PR #1192, opened 2025-02-07
Inference: remove the hf-only options now that the API is warm-only

julien-c:


coyotte508 (approved):
cc @beurkinger for IE widget

SBrandeis (approved):

PR #1193, opened 2025-02-07
Black Forest Labs Image Models

saksham36:


packages/inference/src/types.ts
```diff
@@ -51,6 +51,7 @@ export const INFERENCE_PROVIDERS = [
 	"replicate",
 	"sambanova",
 	"together",
+	"blackforestlabs-ai",
```
julien-c, line 54:
```suggestion
	"black-forest-labs",
```

I think i would use the same id as your org (https://huggingface.co/black-forest-labs) wdyt?

SBrandeis:
Let's add some tests!
You can find examples here:
https://github.com/huggingface/huggingface.js/blob/f85f5f7e0f07306b958eb8a9c6c00df68942071b/packages/inference/test/HfInference.spec.ts#L776-L861

You will need to override the `hf model id -> black forest labs id` mapping similarily to what is done here:
https://github.com/huggingface/huggingface.js/blob/f85f5f7e0f07306b958eb8a9c6c00df68942071b/packages/inference/test/HfInference.spec.ts#L1072-L1076

julien-c:
@saksham36 let us know if any help is needed!

packages/inference/src/lib/makeRequestOptions.ts
```diff
@@ -85,8 +86,13 @@ export async function makeRequestOptions(
...
 	if (accessToken) {
-		headers["Authorization"] =
-			provider === "fal-ai" && authMethod === "provider-key" ? `Key ${accessToken}` : `Bearer ${accessToken}`;
+		if (provider === "fal-ai" && authMethod === "provider-key") {
+			headers["Authorization"] = `Key ${accessToken}`;
+		} else if (provider === "black-forest-labs" && authMethod === "provider-key") {
+			headers["X-Key"] = accessToken;
+		} else {
+			headers["Authorization"] = `Bearer ${accessToken}`;
+		}
```
julien-c, line 95: 👍

packages/inference/src/tasks/cv/textToImage.ts
```diff
@@ -72,3 +83,33 @@ export async function textToImage(args: TextToImageArgs, options?: Options): Pro
 	}
 	return res;
 }
+
+async function pollBflResponse(url: string): Promise<Blob> {
+	const urlObj = new URL(url);
+	for (let step = 0; step < 5; step++) {
+		await delay(1000);
+		console.debug(`Polling Black Forest Labs API for the result... ${step + 1}/5`);
+		urlObj.searchParams.set("attempt", step.toString(10));
```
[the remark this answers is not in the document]
SBrandeis: We need to differentiate the requests for the VCR (otherwise, only the first call gets cached and we can't retrieve the result)

julien-c (approved):
Thanks @saksham36 for starting the PR and thanks @SBrandeis for pushing it over the finish line! let's go!!

PR #1197, opened 2025-02-12
Update hardware.ts

Archsage1:
Add RTX40 Laptop GPU


packages/tasks/src/hardware.ts
```diff
@@ -140,6 +140,26 @@ export const SKUS = {
 				tflops: 22.06,
 				memory: [8, 16],
 			},
+			"RTX 4090 LAPTOP": {
```
pcuenca, line 143: I _think_ the product name is `RTX 4090 Mobile`, is it not? I found 32.98 fp16 flops for this card [here](https://www.techpowerup.com/gpu-specs/geforce-rtx-4090-mobile.c3949).

Archsage1: Name source from nvidia homepage, but TFLOPS from other site.
The information about TFLOPS is located elsewhere, so it might need correction.

packages/tasks/src/hardware.ts
```diff
@@ -204,10 +204,26 @@ export const SKUS = {
...
+			},
+			"RTX 4080 Laptop": {
+				tflops: 24.72,
+				memory: [12],
+			},
+			"RTX 4070 Laptop": {
+				tflops: 15.62,
+				memory: [8],
+			},
+  		"RTX 4060 Laptop": {
```
pcuenca, line 219:
```suggestion
			"RTX 4060 Laptop": {
```

PR #1199, opened 2025-02-12
[inference] Kill the ability to override a model's task

julien-c:
#### Excerpt from [internal convo](https://huggingface.slack.com/archives/C021329G2LT/p1739370292210139):

 I was wondering, do you know by any chance if, in the old hf-inference-api, the ability to override the pipeline when calling the API between a feature-extraction  and a sentence-similarity  model, is still used?

I am thinking to kill this feature from the client libraries (it's simpler to always assume 1 model === 1 task) and i figured you might know if we still use this anywhere


coyotte508 (approved):

julien-c:
(i think the CI failure is transient/unrelated)

SBrandeis (approved):

PR #1201, opened 2025-02-13
[inference] Partial revert of #1199, cleaner implem

julien-c:


packages/inference/src/lib/makeRequestOptions.ts
```diff
@@ -243,6 +243,10 @@ function makeUrl(params: {
 		}
 		default: {
 			const baseUrl = HF_HUB_INFERENCE_PROXY_TEMPLATE.replaceAll("{{PROVIDER}}", "hf-inference");
+			if (params.taskHint && ["feature-extraction", "sentence-similarity"].includes(params.taskHint)) {
+				/// when deployed on hf-inference, those two tasks are automatically compatible with one another.
+				return `${baseUrl}/pipeline/${params.taskHint}/${params.model}`;
```
julien-c, line 248: would be cleaner if it was `/{model}/pipeline/{task}`, probably (also simpler to route)

coyotte508 (approved):

Wauplin (approved):
Let's do that :smile: I made the same PR for huggingface_hub: https://github.com/huggingface/huggingface_hub/pull/2861

PR #1202, opened 2025-02-14
Add Cohere provider

alexrs-cohere:
### What
Adds Cohere as an inference provider.

### Test Plan
Added new tests for Cohere both with and without streaming.

### What Should Reviewers Focus On?
Is the implementation correct? Anything important that I missed? Also happy to get feedback on the code, I am a bit rusty with my JS!


SBrandeis (approved):
Looks good, thank you so much!
I took the liberty to factor some cohere-specific types into the `providers/cohere` file

Test failure is unrelated to the current PR

lfayoux (approved):

packages/inference/src/tasks/nlp/chatCompletionStream.ts
```diff
@@ -1,7 +1,6 @@
 import type { BaseArgs, Options } from "../../types";
 import { streamingRequest } from "../custom/streamingRequest";
 import type { ChatCompletionInput, ChatCompletionStreamOutput } from "@huggingface/tasks";
-
```
SBrandeis, line 4:
```suggestion

```

SBrandeis (approved):
Very nice 🔥 Thank you for implementing an OpenAI-compatible API!
Looks good to me, let's merge!

Wauplin (approved):
Much cleaner this way! Thanks for making the effort on your side :hugs:

PR #1203, opened 2025-02-14
fix mapping for hyperbolic and huggingface models

connorch:
We accidentally used the incorrect model name in one of our mappings - this is a fix.  Thanks!


julien-c (approved):
LGTM, but note @connorch that this is only used in unit tests i.e. this is not the mapping used in production (which is defined directly on hf.co)

SBrandeis (approved):

SBrandeis:
@connorch do not forget to update the mapping in huggingface API: https://huggingface.co/api/partners/fireworks-ai/models

connorch:
Thanks @SBrandeis and @julien-c !

And yeah, I plan on getting to that today.

PR #1204, opened 2025-02-16
[inference] Proposal: rename `taskHint` param to `task`

julien-c:
so we merge `task` and `taskHint`


SBrandeis (approved):
sounds good
This is a breaking change since the `request` and `streamingRequest` functions are part of the public API

PR #1206, opened 2025-02-16
ollama-utils: add publish and cronjob workflows

ngxson:
In this PR:
- Add cronjob to update template every monday at 7am GMT+0 (so I'll review it after having a 🥐)
- Add workflow to publish the package to npm, need to check with @coyotte508 if we must do anything else to set it up
- Add myself to CODEOWNERS of ollama-utils, not sure if @Vaibhavs10 wants to join?

For the auto-update workflow, here is a demo:
- Output logs: https://github.com/ngxson/huggingface.js/actions/runs/13355822807/job/37298436139
- The created PR: https://github.com/ngxson/huggingface.js/pull/3


Vaibhavs10 (approved):
Conceptually LGTM!

packages/ollama-utils/scripts/generate-automap.ts
```diff
@@ -190,7 +201,13 @@ const getSpecialTokens = (tmpl: string): string[] => {
 			.map(() => workerGetTemplate())
 	);
 
+	console.log("====================================");
 	console.log("DONE");
+	console.log("Added templates for:");
+	console.log(addedModels.join("\n"));
+	console.log("Skipped these models due to error:");
+	console.log(skippedModelsDueToErr.join("\n"));
```
PoornakumarRasiraju, line 209: Can we use console.error() for better error differentiation? This would help separate errors from regular logs, making debugging and log filtering easier.

.github/workflows/ollama-template-update.yml
coyotte508, line 1: We need to be careful about injection when writing stuff to env in github actions - but I think it's fine in this case :+1:  (no user-provided ref or branch name or file name in the vars)

coyotte508 (approved):
thanks

packages/ollama-utils/scripts/generate-automap.ts
ngxson: I'm not sure why stderr sometimes come out-of-order on github CI output, so I'm sticking with `console.log` for now

.github/workflows/ollama-template-update.yml
ngxson: I acknowledge about this class of attack. In fact, that's why I always pass variable firstly via `env` and then access the environment variables in the code. So even if the user try to inject via `${{ steps.changes.outputs.NEW_BRANCH }}` for example, it will still be contained inside a string.

In other words, a bad approach will be:

```
const newBranch = "${{ steps.changes.outputs.NEW_BRANCH }}";
```

Good approach:

```
const newBranch = process.env.NEW_BRANCH;
```

ngxson: (Please note that the above is an example. Indeed, none of the variables in the script is user-controllable as you pointed out)

ngxson:
I'm getting this error:

```
Error: Unhandled error: HttpError: GitHub Actions is not permitted to create or approve pull requests.
```

https://github.com/huggingface/huggingface.js/actions/runs/13501236917/job/37719936942

@coyotte508 @julien-c Do you know how to enable this? Thanks!

julien-c:
maybe ask in #infra channel

PR #1207, opened 2025-02-17
[Inference] text-to-image: option to return a URL instead of Blob

SBrandeis:
Context (internal): https://huggingface.slack.com/archives/C07KX53FZTK/p1739562014823989?thread_ts=1739561786.364929&cid=C07KX53FZTK


Wauplin (approved):
LGTM!
I continue to like how TS typing system works :smile:

hanouticelina (approved):

PR #1214, opened 2025-02-20
Add biology tag

NielsRogge:
This PR adds a tag for "biology" models. Some relevant examples include:
- https://huggingface.co/arcinstitute/evo2_40b
- all artifacts here: https://huggingface.co/BIOMEDICA
- all artifacts here: https://huggingface.co/metagene-ai
- https://huggingface.co/chaidiscovery/chai-1
- https://huggingface.co/genbio-ai/AIDO.Protein-16B
- https://huggingface.co/facebook/esmfold_v1
- (many more)


NielsRogge:
Hi @merveenoyan and @Vaibhavs10 could you have a look?

julien-c (changes requested):
hmmm, no, in our taxonomy system this is for sure not a pipeline type

This is a domain tag, like we have in this list in the moon repo:

```ts
export const OTHER_TAGS_SUGGESTIONS = [
	"chemistry",
	"biology",
	"finance",
	"legal",
	"music",
	"art",
	"code",
	"climate",
	"medical",
	TAG_NFAA_CONTENT,
] as const;
```

but it's already there

NielsRogge:
Ok will revise the PR, my goal was just so that it shows up here at the "Other" category

Screenshot 2025-02-25 at 10 40 16

julien-c:
this will not show up here, as it is not a pipeline type

julien-c:
it could be displayed here maybe

Screenshot 2025-02-25 at 10 47 57

merveenoyan:
@julien-c can we have a domain taxonomy listed in other tab instead? we can have "legal" "biology" "art" "finance" to start with

NielsRogge:
Any reason not to include it in the "Other" category besides graph machine learning?

I don't think we have pipelines for graph ML, robotics and reinforcement learning models either?

I would prefer to use the term "task category" (as datasets does) for these filters on the left side bar, rather than "pipeline tags". Pipeline tags is a term that originated from the Transformers library (we used to have a one-on-one mapping between these tags and actual pipelines in Transformers). Right now these are more general task filters rather than actual pipelines

julien-c:
@merveenoyan that's what i suggest in my comment above no?

julien-c:
BTW note that we will probably switch to this as main interface in the future =)


models-filters-next

merveenoyan:
@julien-c hmm yes, I thought it was a temporary thing and not a full fledged thing

I would love that

merveenoyan:
@NielsRogge I kinda agree with @julien-c perhaps we could move robotics to domains and then have tasks for robotics models, e.g. vision language models are heavily used so might as well have robotics domain and image-text-to-text task tag

NielsRogge:
cc @julien-c another reason we'd need a biology tag: https://huggingface.co/datasets/bioscan-ml/CanadianInvertebrates-ML/discussions/2#6871577597d12708e79359eb

PR #1215, opened 2025-02-21
Update model-libraries.ts - add specification for WHAM

katja-hofmann:
Configuring countDownloads for wham files with ckpt file extension


katja-hofmann:
Hi team,

I see that this PR is waiting on lint - does this usually take this long (about 10 hours since I submitted the PR), or is there something wrong that I can correct on my side?

image

julien-c:
@katja-hofmann i've approved to launch CI

Wauplin (commented):
Hi @katja-hofmann , thanks for your PR! Changes look good to me.

I just want to highlight that by having a single repo https://huggingface.co/microsoft/wham for both 1.6B et 200M models, you won't be able to know the number of downloads for each model. What we usually recommend to do is to create 1 repo per model (so for instance `microsoft/wham-1.6B` and `microsoft/wham-200M`) and tag them with `library_name: wham`. This way download counts will be separated and the door will be open to add even more models on the Hub in the future (`WHAM_1.6B_v2.ckpt` for instance?).

Since the repo also contains code to execute the model, you can either decide to keep the code in a single repo (e.g. 1.6B repo has the code, while others are referring to the "main" one) or to move the codebase to a GH repo for instance. Depends on what would be the easiest for you (I personally like having code+weights on HF) 

Let me know what you think :) This topic is slightly orthogonal to this PR but always good to handle these things at the same time.

katja-hofmann:
Hi @Wauplin - thank you for your detailed comments and follow up. If you're ok with it, we'd prefer to keep the repo as is. We are not currently planning to add additional instances, and therefore prefer to keep the material in one place.

Wauplin (approved):
Ok fine for me, thanks for confirming!

PR #1216, opened 2025-02-21
wip: add HW requirements calculator

ngxson:
To be discussed with @Vaibhavs10 

Demo:

```
mem {
  "name": "hexgrad/Kokoro-82M",
  "memory": {
    "minimumGigabytes": 1.991212226,
    "recommendedGigabytes": 2.1903334486
  }
}
mem {
  "name": "microsoft/OmniParser-v2.0",
  "memory": {
    "minimumGigabytes": 1.664,
    "recommendedGigabytes": 1.8304
  }
}
mem {
  "name": "deepseek-ai/DeepSeek-R1-Distill-Qwen-1.5B",
  "memory": {
    "minimumGigabytes": 5.4170575452,
    "recommendedGigabytes": 5.95876329972
  }
}
mem {
  "name": "deepseek-ai/DeepSeek-R1-Distill-Llama-8B",
  "memory": {
    "minimumGigabytes": 20.424667624799998,
    "recommendedGigabytes": 22.467134387279998
  }
}
mem {
  "name": "NousResearch/DeepHermes-3-Llama-3-8B-Preview",
  "memory": {
    "minimumGigabytes": 20.4246676512,
    "recommendedGigabytes": 22.46713441632
  }
}
mem {
  "name": "unsloth/DeepSeek-R1-Distill-Llama-8B-unsloth-bnb-4bit",
  "memory": {
    "minimumGigabytes": 8.309023701600001,
    "recommendedGigabytes": 9.139926071760001
  }
}
```

### DIFF
--- a/packages/hub/src/lib/hardware-requirements.ts
+++ b/packages/hub/src/lib/hardware-requirements.ts
@@ -1,0 +1,72 @@
+import { ListFileEntry, listFiles } from "./list-files";
+
+export interface MemoryRequirements {
+	minimumGigabytes: number;
+	recommendedGigabytes: number;
+};
+
+export interface HardwareRequirements {
+	name: string;
+	memory: MemoryRequirements;
+};
+
+export async function getHardwareRequirements(params: {
+	/**
+	 * The model name in the format of `namespace/repo`.
+	 */
+	name: string;
+	/**
+	 * The context size in tokens, default to 2048.
+	 */
+	contextSize?: number;
+}) {
+	const files = await getFiles(params.name);
+	const hasSafetensors = files.some((file) => file.path.endsWith(".safetensors"));
+	const hasPytorch = files.some((file) => file.path.endsWith(".pth"));
+
+	// Get the total size of the model weight in bytes (we don't care about quantization scheme)
+	let totalWeightBytes = 0;
+	if (hasSafetensors) {
+		totalWeightBytes = sumFileSize(files.filter((file) => file.path.endsWith(".safetensors")));
+	} else if (hasPytorch) {
+		totalWeightBytes = sumFileSize(files.filter((file) => file.path.endsWith(".pth")));
+	}
+
+	// Calculate the memory for context window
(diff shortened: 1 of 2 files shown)


Vaibhavs10 (commented):
Nice! I like this - some quick questions:
1. How do we fetch the users hardware from their Local Apps page?
2. Should we maybe start with just LLMs to begin with?

Quite excited to make this work!

cc: @julien-c for thoughts too

packages/hub/src/lib/hardware-requirements.ts
```diff
@@ -0,0 +1,72 @@
...
+
+export async function getHardwareRequirements(params: {
+	/**
+	 * The model name in the format of `namespace/repo`.
+	 */
+	name: string;
+	/**
+	 * The context size in tokens, default to 2048.
+	 */
+	contextSize?: number;
```
julien-c, line 21: if possible `contextSize` should be taken in `tasks.ModelData` so it's clearer the data is coming from normalized parsing of HF model repos

julien-c: and tbh i'm wondering if this kinda of method should better live in `tasks` or `gguf` modules rather than here

julien-c (commented):
So, we would focus on only RAM/VRAM in the beginning, correct? I agree with this approach, but i would maybe start with just GGUF, and later extend to other formats

packages/hub/src/lib/hardware-requirements.ts
ngxson: IMO the `contextSize` should be a number that most users gonna use in practice, for example 2048 or 4096

Setting it to the max `contextSize` of the model may not be useful, as most users never have enough VRAM to run full context length anyway (especially, 128k context are more and more common nowadays)

julien-c: ok, i see your point. cc @gary149 wdyt?
maybe it's another selector in the UI then with "user-set context size"?

PR #1220, opened 2025-02-24
add Featherless as a provider

DarokCx:
First pass at trying to add Featherless.ai as a provider. 

Did some local and it was working well. I am only unsure how local tests would work without api-key.

Added some tests for chat completions, both streaming and non streaming.


julien-c:
did we want to revive this PR?

PR #1221, opened 2025-02-24
gguf: add CLI

ngxson:
Ref discussion: https://huggingface.slack.com/archives/C02CLHA19TL/p1740399079674399?thread_ts=1739968558.574099&cid=C02CLHA19TL

I'm trying with this command:

```bash
pnpm run build && npx . ~/work/models/Meta-Llama-3.1-8B-Instruct-Q4_K_M.gguf
```

Output:

```
* Dumping 36 key/value pair(s)
  Idx | Count  | Value                                                                            
  ----|--------|----------------------------------------------------------------------------------
    1 |      1 | version = 3                                                                      
    2 |      1 | tensor_count = 292                                                               
    3 |      1 | kv_count = 33                                                                    
    4 |      1 | general.architecture = "llama"                                                   
    5 |      1 | general.type = "model"                                                           
    6 |      1 | general.name = "Meta Llama 3.1 8B Instruct"                                      
    7 |      1 | general.finetune = "Instruct"                                                    
    8 |      1 | general.basename = "Meta-Llama-3.1"                                              
    9 |      1 | general.size_label = "8B"                                                        
   10 |      1 | general.license = "llama3.1"                                                     
   11 |      6 | general.tags = ["facebook","meta","pytorch","llama","llama-3","te...             
   12 |      8 | general.languages = ["en","de","fr","it","pt","hi","es","th"]                    
   13 |      1 | llama.block_count = 32                                                           
   14 |      1 | llama.context_length = 131072                                                    
   15 |      1 | llama.embedding_length = 4096                                                    
   16 |      1 | llama.feed_forward_length = 14336                                                
   17 |      1 | llama.attention.head_count = 32                                                  
   18 |      1 | llama.attention.head_count_kv = 8                                                
   19 |      1 | llama.rope.freq_base = 500000                                                    
   20 |      1 | llama.attention.layer_norm_rms_epsilon = 0.000009999999747378752                 
   21 |      1 | general.file_type = 15                                                           
   22 |      1 | llama.vocab_size = 128256                                                        
   23 |      1 | llama.rope.dimension_count = 128                                                 
   24 |      1 | tokenizer.ggml.model = "gpt2"                                                    
   25 |      1 | tokenizer.ggml.pre = "llama-bpe"                                                 
   26 | 128256 | tokenizer.ggml.tokens = ["!","\"","#","$","%","&","'","(",")","*","+",",",...    
[...]

PR #1225, opened 2025-02-25
gguf: add memory calculator to CLI

ngxson:
The logic to calculate memory usage is taken from various cpp files in the project. This can be accurate down to around 100MB

Example:

```
pnpm run build && npx . ~/work/models/Meta-Llama-3.1-8B-Instruct-Q4_K_M.gguf -c 8000

* Memory usage estimation (with 8000 tokens in context)
  Item     | Memory usage
  ---------|-------------
  K cache  |      0.52 GB
  V cache  |      0.52 GB
  Weight   |      4.91 GB
  Overhead |      0.28 GB
           |          ---
  TOTAL    |      6.24 GB
```

TODO:
- [x] Fix multi shard gguf
- [x] Do not calc for recurrence models for now (mamba, rwkv)


packages/gguf/src/gguf.ts
```diff
@@ -410,6 +410,7 @@ export async function ggufAllShards(
 		fetch?: typeof fetch;
 		additionalFetchHeaders?: Record<string, string>;
 		parallelDownloads?: number;
+		allowLocalFile?: boolean;
```
julien-c, line 413: 👍

julien-c (approved):
this is cool! 

general note, let's make sure we increase visibility on stuff like this, same for #1221

julien-c:
This should be merged (and communicated about), no?

ngxson:
Yes I'll have a look on this later today, should be pretty much merge-able at this point

ngxson:
Merging this once the CI passes

julien-c:
let's make sure to do some comms!

PR #1231, opened 2025-02-26
Also test web blob on xet file

coyotte508:
Follow up #1230 cc @Wauplin 

Since the header should be present now


Wauplin (approved):
:+1: 

(to merge once CI's green)

PR #1235, opened 2025-02-27
featureExtraction should use FeatureExtractionInput

mishig25:
Read the comment https://github.com/huggingface/huggingface.js/pull/1235/files#r1973270449

### DIFF
--- a/packages/inference/src/tasks/nlp/featureExtraction.ts
+++ b/packages/inference/src/tasks/nlp/featureExtraction.ts
@@ -1,7 +1,9 @@
 import type { FeatureExtractionInput } from "@huggingface/tasks";
 import { InferenceOutputError } from "../../lib/InferenceOutputError";
-import type { Options } from "../../types";
+import type { BaseArgs, Options } from "../../types";
 import { request } from "../custom/request";
+
+export type FeatureExtractionArgs = BaseArgs & FeatureExtractionInput;
 
 /**
  * Returned values are a multidimensional array of floats (dimension depending on if you sent a string or a list of string, and if the automatic reduction, usually mean_pooling for instance was applied for you or not. This should be explained on the model's README).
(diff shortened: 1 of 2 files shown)


packages/inference/src/tasks/nlp/featureExtraction.ts
```diff
@@ -1,17 +1,8 @@
+import type { FeatureExtractionInput } from "@huggingface/tasks";
 import { InferenceOutputError } from "../../lib/InferenceOutputError";
-import type { BaseArgs, Options } from "../../types";
+import type { Options } from "../../types";
 import { request } from "../custom/request";
 
-export type FeatureExtractionArgs = BaseArgs & {
```
mishig25, line 5: https://github.com/huggingface/huggingface.js/blob/fc79ea97031a27be67654b6fdd82f313c49945e3/packages/tasks/src/tasks/feature-extraction/inference.ts#L14-L36

coyotte508 (approved):

SBrandeis (changes requested):
The input type must extend `BaseArgs` too, see:

https://github.com/huggingface/huggingface.js/blob/5f7f4233807a41a344676c522ddf17d9a09d3a66/packages/inference/src/tasks/nlp/summarization.ts#L6-L12

julien-c:
yes good catch @SBrandeis

mishig25:
pushed https://github.com/huggingface/huggingface.js/pull/1235/commits/b6cae203e174e8d22fd11dbc13c276880b986f76

mishig25:
pushed https://github.com/huggingface/huggingface.js/pull/1235/commits/040e2d63d306457516bb2f90c7bb22126a9bee05 after more inspection

I guess more tyoe-checking is needed to add in hf.js/inference ? since TS check was all green but https://github.com/huggingface/huggingface.js/pull/1235/commits/040e2d63d306457516bb2f90c7bb22126a9bee05 is needed I think

julien-c:
please can you wait for a review from @SBrandeis when he asked for changes?

PR #1241, opened 2025-02-28
Add Hunyuan3D-2 Library

Zeqiang-Lai:


Vaibhavs10 (commented):
Hi @Zeqiang-Lai - Thanks for the PR. Is this PR for: https://huggingface.co/tencent/Hunyuan3D-2/tree/main

In general our recommendation is to have individual repos for individual models, instead of having single monorepo. This also allows for getting visibility to the individual checkpoints as well and makes it easy for downstream apps to manage them as well.

Bonus: This way you also get individual granular download counts for different types of repos.

In addition to this, if we merge this change, for each folder we'll count all the safetsensors - that would inflate the download counts for each checkpoint by 5-6x and for the repo - 25-30x (given the other checkpoints in the same repo)

Zeqiang-Lai:
Thanks vaibhavs 

We are planning to release future model individually. But we need to keep current one for backward compatibility.

The current repo contains 4 models, the paint and delight models are diffusers formatted, while the shape model is custom. 

It seems that when config.json in root is downloaded, the count increases. This means the count only count snapshot download of full model. We would like to measure the total download of these 4 models.

It is possible to do that? What's the recommended way to correct it?

packages/tasks/src/model-libraries.ts
```diff
@@ -394,6 +394,12 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		repoUrl: "https://github.com/Tencent/HunyuanDiT",
 		countDownloads: `path:"pytorch_model_ema.pt" OR path:"pytorch_model_distill.pt"`,
 	},
+	"hunyuan3d-2": {
+		prettyLabel: "Hunyuan3D-2",
+		repoName: "Hunyuan3D-2",
+		repoUrl: "https://github.com/Tencent/Hunyuan3D-2",
+		countDownloads: `path_extension:"safetensors" OR path_extension:"ckpt"`,
```
Vaibhavs10, line 401:
```suggestion
		countDownloads: `path:"model_index.json" OR path:"config.yaml"`,
```

Chatted with @Wauplin and we think that this is a good compromise as only one or the other are used and this is still quite future proof even when you split to other repos as well! Let me know if this doesn't work for you!

Zeqiang-Lai: Sure, this solution would be great! I just made an edition according to the suggesiton.

Vaibhavs10 (approved):

Vaibhavs10:
Thanks for the PR @Zeqiang-Lai and for iterating on feedback - please make sure that all future repos (even the splitted ones) have the same tag as in the PR.

The download counts should be live in 3-4 days! Cheers!

Zeqiang-Lai:
@Vaibhavs10 Hi, I am a little confused about path and path_filename, will 
```
countDownloads: `path:"model_index.json" OR path:"config.yaml"`,
```
correctly count when our model_index.json and config.yaml live in the subfolder ?.
https://huggingface.co/tencent/Hunyuan3D-2/blob/main/hunyuan3d-delight-v2-0/model_index.json

BTW, will history download counts refresh with new rules?

Zeqiang-Lai:
@Wauplin sure, just open a new pull request as in https://github.com/huggingface/huggingface.js/pull/1256

PR #1257, opened 2025-03-07
XetBlob util

coyotte508:
Python lib code: https://github.com/huggingface-internal/xetpoc_huggingface_hub/pull/14/files (cc @bpronan @Wauplin @hanouticelina)

# Todo

Add more test cases

- [x] on a larger amount of data (need to fetch several terms)
- [x] when offset_into_first_range > a chunk length
- [ ] when term.range.start > fetchinfo.range.start => created https://github.com/huggingface/huggingface.js/issues/1285
- [x] check range request when requesting range at exact chunk boundaries, or just before/just after

### DIFF
--- a/packages/hub/src/utils/XetBlob.ts
+++ b/packages/hub/src/utils/XetBlob.ts
@@ -88,7 +88,7 @@ export class XetBlob extends Blob {
 	constructor(params: XetBlobCreateOptions) {
 		super([]);
 
-		this.fetch = params.fetch ?? fetch;
+		this.fetch = params.fetch ?? fetch.bind(globalThis);
 		this.accessToken = checkCredentials(params);
 		this.repoId = toRepoId(params.repo);
 		this.hubUrl = params.hubUrl ?? HUB_URL;
@@ -163,7 +163,7 @@ export class XetBlob extends Blob {
 
 		async function* readData(reconstructionInfo: ReconstructionInfo, customFetch: typeof fetch, maxBytes: number) {
 			let totalBytesRead = 0;
-			let isFirstChunk = true;
+			let readBytesToSkip = reconstructionInfo.offset_into_first_range;
 
 			for (const term of reconstructionInfo.terms) {
 				if (totalBytesRead >= maxBytes) {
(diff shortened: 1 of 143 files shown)


hanouticelina:
cc @assafvayner too for viz

packages/hub/src/utils/XetBlob.ts
```diff
@@ -0,0 +1,439 @@
...
+							if (isFirstChunk) {
+								yield uncompressed.slice(
+									reconstructionInfo.offset_into_first_range,
+									Math.min(uncompressed.length, reconstructionInfo.offset_into_first_range + maxBytes - totalBytesRead)
+								);
+								totalBytesRead += Math.min(
+									uncompressed.length,
+									reconstructionInfo.offset_into_first_range + maxBytes - totalBytesRead
+								);
+								isFirstChunk = false;
```
assafvayner, line 282: reconstructionInfo.offset_into_first_range may be larger than 1 chunk length.

suppose each chunk uncompressed was 100 bytes, first term has 10 terms, reconstructionInfo.offset_into_first_range can be 250, so we need to skip the first 2 whole chunks (we will only know their lengths after reading the chunk header) and then the first 50 bytes of the 3rd chunk.

coyotte508: ok so the `term` can give a range with whole chunks to skip if I understand correctly

assafvayner: to be clear `offset_into_first_range` is only ever set non-zero if the client requested a range, it is guarenteed that the first byte of a file is also the first byte of the first chunk in a term.

`offset_info_first_range` is non-zero when the first byte requested by the client is in the middle of the term.

coyotte508: should be fixed

coyotte508:
This PR is ready to review! Will merge it early next week

It exposes an `__internal_XetBlob` export that can be played around with

Before more serious / production usage, at the very least https://github.com/huggingface/huggingface.js/issues/1284 should be dealt with. But it involves maintaining ranges of chunks (and intersection of ranges of chunks, know when some chunks can be discarded from cache) and crafting / mocking tests 

Still, in the mean time we can merge this & have at least some xet tests added & do some early xet tests with the published package.

PR #1261, opened 2025-03-09
add MatAnyone

not-lain:
this pr is following https://github.com/pq-yang/MatAnyone/pull/26 after the integration of MatAnyone with PyTorchModelHubMixin


Wauplin (approved):
Very cool! :fire: 

I can see that models are already correctly tagged at https://huggingface.co/models?other=matanyone so let's get this merged :)

PR #1272, opened 2025-03-12
fix: better handling of symlink (Windows,...)

### LINKED ISSUE #1268: Hub snapshot_download not compatible with Windows
It is using symlink which is rejected on most Windows configurations (requires administrator priviledges)

Should use helper function similar to 

https://github.com/huggingface/huggingface_hub/blob/fcfd14361bd03f23f82efced1aa65a7cbfa4b922/src/huggingface_hub/file_download.py#L517

jeffmaury:
Add finer handling of symlink as it is not supported on Windows.

Fixes #1268

### DIFF
--- a/packages/hub/src/utils/symlink.ts
+++ b/packages/hub/src/utils/symlink.ts
@@ -1,0 +1,123 @@
+import * as fs from "node:fs/promises";
+import * as path from "node:path";
+import { getHFHubCachePath } from "../lib";
+import * as os from "node:os";
+
+const _are_symlinks_supported_in_dir = new Map<string, boolean>();
+
+export function reset(): void {
+	_are_symlinks_supported_in_dir.clear();
+}
+
+function expandUser(path: string): string {
+	if (path.startsWith("~")) {
+		return path.replace("~", process.env.HOME ?? "");
+	}
+	return path;
+}
+
+function commonPath(paths: string[]): string {
+	if (!paths.length) return "";
+
+	const parts = paths.map((p) => p.split(path.sep));
+	const common = parts[0];
+
+	for (let i = 1; i < parts.length; i++) {
+		for (let j = 0; j < common.length; j++) {
+			if (parts[i][j] !== common[j]) {
+				common.length = j;
+				break;
+			}
+		}
+	}
+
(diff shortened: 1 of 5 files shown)


coyotte508:
Thanks for the contribution!

cc @Wauplin @hanouticelina  for viz since it's also in the python implem

packages/hub/src/utils/symlink.ts
```diff
@@ -0,0 +1,172 @@
...
+		_are_symlinks_supported_in_dir.set(cache_dir, true);
+
+		await fs.mkdir(cache_dir, { recursive: true });
+		const tmp_dir = await fs.mkdtemp(cache_dir);
+		try {
+			const src_path = path.join(tmp_dir, "dummy_src");
+			await fs.open(src_path, "w").then((fd) => fd.close());
+			const dst_path = path.join(tmp_dir, "dummy_dst");
+
+			const relative_src = path.relative(src_path, path.dirname(dst_path));
```
coyotte508, line 64:
```suggestion
			const relative_src = path.relative(path.dirname(dst_path), src_path);
```

I think python & JS arguments are not in the same order

packages/hub/src/utils/symlink.ts
```diff
@@ -0,0 +1,172 @@
...
+			if (_are_symlinks_supported_in_dir.size > MAX_CACHE_SIZE) {
+				const key = _are_symlinks_supported_in_dir.keys().next().value;
+				if (key) {
+					_are_symlinks_supported_in_dir.delete(key);
+				}
+			}
+		}
+	}
+	return _are_symlinks_supported_in_dir.get(cache_dir) ?? false;
+}
```
coyotte508, line 105: Is it possible to remove this whole function?

By doing in `createSymlink` something like:

```ts
	try {
    await symlink(dest, src);
    return;
  } catch {}

  // symlink failed, copy or rename the file
```

coyotte508 (approved):
Thanks

I think we may still want to use relative paths (from what is said in the doc comment) on windows in the symlink, but current version is already an improvement to make it work better in windows.

Approving and  will merge in a few days

axel7083 (approved):

PR #1292, opened 2025-03-17
[Inference Providers] Async calls for `text-to-video` with fal.ai

hanouticelina:
## What does this PR do?

This PR adds asynchronous polling to the fal.ai text-to-video generation. This allows running inference with models that may take > 2 min to generate results. The other motivation behind this PR is to align the Python and JS clients, the Python client has already been merged into main: https://github.com/huggingface/huggingface_hub/pull/2927

## Main Changes
- Replaced static `baseUrl` property with `makeBaseUrl()` function across all providers. This is needed to be able to customize the base url based on the task. We want to use `FAL_AI_API_BASE_URL_QUEUE` for `text-to-video` only. I'm not convinced if it's the simplest and the best way to do that.
- Added a `pollFalResponse()` for `text-to-video`(similarly to what it's done with BFL for `text-to-image`).

Any refactoring suggestions are welcome! I'm willing to spend some additional time to make provider-specific updates easier to implement and better align our two clients 🙂 

btw, I did not update the VCR tests as we've discussed that it'd be best to remove the VCR for `text-to-video`. Maybe we should remove them here? 
**EDIT**: removed the text-to-video tests in [f8a6386](https://github.com/huggingface/huggingface.js/pull/1292/commits/f8a638657973ff8ce31b19581b50ed68d4c87b93).

I've tested it locally with [tencent/HunyuanVideo](https://huggingface.co/tencent/HunyuanVideo) for which the generation takes more than 2min and it works fine:

https://github.com/user-attachments/assets/3cd38900-c4ed-4b28-ae79-8a4e724f58d1


packages/inference/src/providers/fal-ai.ts
```diff
@@ -29,11 +34,16 @@ const makeHeaders = (params: HeaderParams): Record<string, string> => {
 };
 
 const makeUrl = (params: UrlParams): string => {
-	return `${params.baseUrl}/${params.model}`;
+	const baseUrl = `${params.baseUrl}/${params.model}`;
+	if (params.authMethod !== "provider-key" && params.task === "text-to-video") {
+		return `${baseUrl}?_subdomain=queue`;
+	}
+	console.log("baseUrl", baseUrl);
```
julien-c, line 41:
```suggestion
```

packages/inference/src/providers/fal-ai.ts
```diff
@@ -14,9 +14,14 @@
  *
  * Thanks!
  */
-import type { ProviderConfig, UrlParams, HeaderParams, BodyParams } from "../types";
+import type { BodyParams, HeaderParams, InferenceTask, ProviderConfig, UrlParams } from "../types";
 
 const FAL_AI_API_BASE_URL = "https://fal.run";
+const FAL_AI_API_BASE_URL_QUEUE = "https://fal.run/queue";
```
SBrandeis, line 20: Isn't it this URL?
```suggestion
const FAL_AI_API_BASE_URL_QUEUE = "https://queue.fal.run";
```

SBrandeis (commented):
Looks good 🤩 
Let's also update the tapes

packages/inference/src/tasks/cv/textToVideo.ts
```diff
@@ -85,3 +73,44 @@ export async function textToVideo(args: TextToVideoArgs, options?: Options): Pro
...
+		!!result.video &&
+		"url" in result.video &&
+		typeof result.video.url === "string" &&
+		isUrl(result.video.url);
+	if (!isValidOutput) {
+		throw new InferenceOutputError("Expected { video: { url: string } }");
+	}
+	const urlResponse = await fetch(result.video.url);
+	return await urlResponse.blob();
+}
```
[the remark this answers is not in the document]
hanouticelina: ah yes, we should probably do the same for: https://github.com/huggingface/huggingface.js/blob/0a0960c8a41df5a82296754b9b43d328755a0cff/packages/inference/src/tasks/cv/textToImage.ts#L132

packages/inference/src/providers/fal-ai.ts
hanouticelina: my bad!

packages/inference/src/tasks/cv/textToVideo.ts
```diff
@@ -85,3 +73,44 @@ export async function textToVideo(args: TextToVideoArgs, options?: Options): Pro
...
+	let status = res.status;
+	const { url, info } = await makeRequestOptions(args, { ...options, task: "text-to-video" });
+	const baseUrl = url?.split("?")[0] || "";
+	const query = url?.includes("_subdomain=queue") ? "?_subdomain=queue" : "";
+
+	const statusUrl = `${baseUrl}/requests/${requestId}/status${query}`;
+	const resultUrl = `${baseUrl}/requests/${requestId}${query}`;
+
+	while (status !== "COMPLETED") {
+		await delay(1000);
```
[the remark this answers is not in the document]
hanouticelina: I think we are using 2 sec in the Python client, it was an arbitrary value to be honest 😬 . 
Here I used the same value as the one used for BFL. I think we can set it to half a sec? wdyt?

packages/inference/src/tasks/cv/textToVideo.ts
```diff
@@ -85,3 +73,44 @@ export async function textToVideo(args: TextToVideoArgs, options?: Options): Pro
...
+	const isValidOutput =
+		typeof result === "object" &&
+		!!result &&
+		"video" in result &&
+		typeof result.video === "object" &&
+		!!result.video &&
+		"url" in result.video &&
+		typeof result.video.url === "string" &&
+		isUrl(result.video.url);
+	if (!isValidOutput) {
```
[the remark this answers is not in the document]
hanouticelina: I can definitely take care of that in a following PR!

packages/inference/src/tasks/cv/textToVideo.ts
```diff
@@ -85,3 +73,44 @@ export async function textToVideo(args: TextToVideoArgs, options?: Options): Pro
...
+
+async function pollFalResponse(res: FalAiOutput, args: TextToVideoArgs, options?: Options): Promise<Blob> {
+	const requestId = res.request_id;
+	if (!requestId) {
+		throw new InferenceOutputError("No request ID found in the response");
+	}
+	let status = res.status;
+	const { url, info } = await makeRequestOptions(args, { ...options, task: "text-to-video" });
+	const baseUrl = url?.split("?")[0] || "";
+	const query = url?.includes("_subdomain=queue") ? "?_subdomain=queue" : "";
```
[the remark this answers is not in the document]
hanouticelina: addressed in [cf2d1ac](https://github.com/huggingface/huggingface.js/pull/1292/commits/cf2d1ac81928715fe00bcbfbf469cb0c782122bb)

packages/inference/src/tasks/cv/textToVideo.ts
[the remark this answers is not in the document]
Wauplin: > maybe we should move every provider-specific output type to their respective provider files, wdyt ?

Agree with this and in particular their output type + logic to validate and parse the response. Same as the `get_response`'s in Python. Out of scope for this PR though

packages/inference/src/tasks/cv/textToVideo.ts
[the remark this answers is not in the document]
SBrandeis: > I can definitely take care of that in a following PR!

Discussed with @coyotte508 and @julien-c , let's avoid adding external dependencies

packages/inference/src/providers/fal-ai.ts
```diff
@@ -29,12 +37,71 @@ const makeHeaders = (params: HeaderParams): Record<string, string> => {
...
 
 export const FAL_AI_CONFIG: ProviderConfig = {
-	baseUrl: FAL_AI_API_BASE_URL,
+	makeBaseUrl,
 	makeBody,
 	makeHeaders,
 	makeUrl,
 };
+
+export interface FalAiOutput {
```
Wauplin, line 54:
```suggestion
export interface FalAiQueueOutput {
```
(nit)

packages/inference/src/tasks/cv/textToVideo.ts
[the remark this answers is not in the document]
coyotte508: Note there is this PR : https://github.com/huggingface/huggingface.js/pull/385/files 

with an internal validator util (mini-zod)

packages/inference/src/tasks/cv/textToVideo.ts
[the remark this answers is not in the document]
Wauplin: no matter the value, let's align both clients :)

Wauplin (commented):
:fire:

packages/inference/src/tasks/cv/textToVideo.ts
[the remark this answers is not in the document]
hanouticelina: > no matter the value, let's align both clients :)

yes, agree. I'll update the python client and use the same smaller value. 
FYI in their client, they use 0.1s interval: [fal-ai/fal/projects/fal_client/src/fal_client/client.py#L703](https://github.com/fal-ai/fal/blob/75fe22f19cf61c7b6488d919d9a8c4bcb3433b42/projects/fal_client/src/fal_client/client.py#L703). I guess we can use the same value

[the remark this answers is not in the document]
Wauplin: let's just be careful on the rate limits (expecting 600req/min)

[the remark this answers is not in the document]
hanouticelina: yes i was about to update my comment: let's use 0.5s instead given that `text-to-video` generation takes more time than other tasks like text-to-image.

packages/inference/src/providers/fal-ai.ts
```diff
@@ -29,12 +37,85 @@ const makeHeaders = (params: HeaderParams): Record<string, string> => {
...
+	// from the response as it might be different from the mapped model in `url`
+	const modelId = new URL(res.response_url).pathname;
+	const queryParams = parsedUrl.search;
+
+	const statusUrl = `${baseUrl}${modelId}/status${queryParams}`;
+	const resultUrl = `${baseUrl}${modelId}${queryParams}`;
+
+	while (status !== "COMPLETED") {
+		await delay(500);
+		const statusResponse = await fetch(statusUrl, { headers: headers });
```
SBrandeis, line 86:
```suggestion
		const statusResponse = await fetch(statusUrl, { headers });
```

packages/inference/src/providers/fal-ai.ts
```diff
@@ -29,12 +37,85 @@ const makeHeaders = (params: HeaderParams): Record<string, string> => {
...
+			throw new InferenceOutputError("Failed to fetch response status from fal-ai API");
+		}
+		try {
+			status = (await statusResponse.json()).status;
+		} catch (error) {
+			throw new InferenceOutputError("Failed to parse status response from fal-ai API");
+		}
+	}
+
+	const resultResponse = await fetch(resultUrl, { headers: headers });
```
SBrandeis, line 98:
```suggestion
	const resultResponse = await fetch(resultUrl, { headers });
```

packages/inference/src/providers/fal-ai.ts
```diff
@@ -29,12 +37,85 @@ const makeHeaders = (params: HeaderParams): Record<string, string> => {
...
+	let result;
```
SBrandeis, line 99:
```suggestion
	let result: unknown;
```

packages/inference/src/providers/fal-ai.ts
```diff
@@ -29,12 +37,85 @@ const makeHeaders = (params: HeaderParams): Record<string, string> => {
...
+		typeof result.video === "object" &&
+		!!result.video &&
+		"url" in result.video &&
+		typeof result.video.url === "string" &&
+		isUrl(result.video.url);
+	if (!isValidOutput) {
+		throw new InferenceOutputError(
+			"Expected { video: { url: string } } result format, got instead: " + JSON.stringify(result)
+		);
+	}
```
SBrandeis, line 118: Note: Moving the condition inside the `if` test will trigger Typescript type inference
```
	if (typeof result === "object" &&
		!!result &&
		"video" in result &&
		typeof result.video === "object" &&
		!!result.video &&
		"url" in result.video &&
		typeof result.video.url === "string" &&
		isUrl(result.video.url) {
			/// result has the correct type here
	} else {
		throw new InferenceOutputError(
			"Expected { video: { url: string } } result format, got instead: " + JSON.stringify(result)
		);
	}
```

packages/inference/src/types.ts
```diff
@@ -107,6 +107,7 @@ export interface HeaderParams {
 }
 
 export interface UrlParams {
+	authMethod: "none" | "hf-token" | "credentials-include" | "provider-key";
```
SBrandeis, line 110: Let's extract the authMethod type in a dedicated type

packages/inference/src/tasks/cv/textToVideo.ts
```diff
@@ -5,17 +5,13 @@ import { omit } from "../../utils/omit";
 import { isUrl } from "../../lib/isUrl";
 import { InferenceOutputError } from "../../lib/InferenceOutputError";
 import { typedInclude } from "../../utils/typedInclude";
+import { makeRequestOptions } from "../../lib/makeRequestOptions";
+import { pollFalResponse, FalAiQueueOutput } from "../../providers/fal-ai";
```
SBrandeis, line 9:
```suggestion
import { pollFalResponse, type FalAiQueueOutput } from "../../providers/fal-ai";
```

packages/inference/src/tasks/cv/textToVideo.ts
```diff
@@ -39,25 +35,13 @@ export async function textToVideo(args: TextToVideoArgs, options?: Options): Pro
...
-			!!res.video &&
-			"url" in res.video &&
-			typeof res.video.url === "string" &&
-			isUrl(res.video.url);
-		if (!isValidOutput) {
-			throw new InferenceOutputError("Expected { video: { url: string } }");
-		}
-		const urlResponse = await fetch((res as FalAiOutput).video.url);
-		return await urlResponse.blob();
+		const { url, info } = await makeRequestOptions(args, { ...options, task: "text-to-video" });
```
SBrandeis, line 43: That works, but it feels a bit weird to have to call `makeRequestOptions` again here

I think we can have a simpler / better API that does not require to mentally map why we call `makeRequestOptions` with those parameters.

I'm wondering whether `request` should return an optional `Promise` that we would just `await` here? - example usage syntax:

```typescript
const res = await request<FalAiQueueOutput | ReplicateOutput | NovitaOutput>(payload, {
	...options,
	task: "text-to-video",
});
if (res.poll) {
	const blob: Blob = await res.poll;
}
```

Anyways - I'm OK with addressing that in a subsequent PR

hanouticelina: yes totally agree, it feels a bit wrong to call `makeRequestOptions` again just after calling `request`. 
I'm currently working on a refactoring PR, I'll rework this part as well

SBrandeis (approved):
Looks good! Thank you!!

kefranabg:
That means we can allow any `text-to-video` model that supports fal-ai to use the inference widget on the hub right? 🙂

SBrandeis:
Yes @kefranabg !

PR #1305, opened 2025-03-21
Download using xet

coyotte508:
cc @hanouticelina @Wauplin @bpronan @assafvayner 

`downloadFile` has a `xet: true` optional param, that when set will download the file with the xet protocol if possible.

## Breaking changes

- `downloadFile` returns a `Blob`
- `fileDownloadInfo`'s return format is changed:
	- `downloadLink` => `url`, and it's present every time
	-  optional `xet` prop for xet files

## Concerns

https://huggingface.co/spaces/coyotte508/hub-xet uses quite a bit of CPU when downloading a full xet file, maybe we need to introduce multi-threading (with workers?) or optimize perf

Especially since we use it to parse safetensors data now cc @Kakulukian 

## Performance work

Before releasing out of experimental, some things need to be addressed & tested on different engines (node, browser):

- CPU usage, eg how fast it is to de-chunk 10GB of local data. Can it be significantly improved with moving some of the code to WASM.
- Stream backpressure, how much RAM does get used, especially due to different handling of Web streams, and if we can make it consistent across
- Http calls are sequentially made, to save RAM, but it can hurt in high-ping high-bandwidth situations. Not a problem if we download multiple files in parallel, but when downloading one large file, we probably don't care about reading data on the wire and can just promise-queue a bunch of http calls and use some RAM. We can probably change the behavior of the lib depending on the downloaded file's size.

Being able to upload xet files would be nice too, experimentally as well (eg for hub-ci tests)


packages/hub/src/lib/file-download-info.ts
```diff
@@ -84,9 +92,24 @@ export async function fileDownloadInfo(
...
+	}
+
 	return {
 		etag,
 		size,
-		downloadLink: new URL(resp.url).hostname !== new URL(hubUrl).hostname ? resp.url : null,
+		xet: xetInfo,
+		// Cannot use resp.url in case it's a S3 url and the user adds an Authorization header to it.
+		url:
+			new URL(resp.url).hostname === new URL(hubUrl).hostname || resp.headers.get("X-Cache")?.endsWith(" cloudfront")
```
coyotte508, line 111: @XciD @rtrompier can you add `X-Cache` and `Etag` to `Access-Control-Expose-Headers` on our cloudfronts?

```
Access-Control-Allow-Origin: *
Access-Control-Expose-Headers: X-Cache, ETag
```

cc @Michellehbn for viz

assafvayner:
> uses quite a bit of CPU when downloading a full xet file, maybe we need to introduce multi-threading (with workers?) or optimize perf

We (myself and @seanses) are also looking at introducing web-workers for multithreading (and general interface for processing uploads) wasm xet-core. Promising stuff so far I think.

coyotte508:
Merging this one, and major version!

julien-c:
> Being able to upload xet files would be nice too, experimentally as well

but for uploads from the Web we will/would use the Wasm version, is that correct? not the pure JS version?

coyotte508:
> > Being able to upload xet files would be nice too, experimentally as well
> 
> but for uploads from the Web we will/would use the Wasm version, is that correct? not the pure JS version?

wasm for chunking

julien-c:
^yes

PR #1308, opened 2025-03-21
Add non-mocked test for symlink

### LINKED ISSUE #1306: createSymlink is deleting the wrong file before perfoming operations
https://github.com/huggingface/huggingface.js/blob/e0de0081545a9936309588fcfde37ab73701d998/packages/hub/src/utils/symlink.ts#L45

coyotte508:
Alternative to https://github.com/huggingface/huggingface.js/pull/1307

Fix #1306 

I was just too confused by the src/dst params, nodej's official docs could be clearer too.

Now the params are explictly named and there is a real test to check the filesystem contents.

By the way, it's a `copy` operation not a `rename` (because it was already the case, the `new_blob` param was never passed)

Also took the opportunty to use relative symlinks cc @Wauplin for better windows support

cc @jeffmaury @axel7083

### DIFF
--- a/packages/hub/src/utils/symlink.spec.ts
+++ b/packages/hub/src/utils/symlink.spec.ts
@@ -4,6 +4,8 @@ import { describe, expect, it, vi } from "vitest";
 import { createSymlink } from "./symlink";
 import { readFileSync, writeFileSync } from "node:fs";
 import { lstat, rm } from "node:fs/promises";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
 
 let failSymlink = false;
 vi.mock("node:fs/promises", async (importOriginal) => ({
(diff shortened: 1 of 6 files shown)


coyotte508:
> I just a little bit concerned because the new tests don't pass on Windows which is why createSymlink has been created and as I'm on Windows it will make my life harder to check I'm not introducing regression

Ok, I added a mocked test on linux, where I just fail the `symlink` function to behave like it does on windows. 

And in the other tests, instead of `		expect(stats.isSymbolicLink()).toBe(true);`, I added `		expect(stats.isSymbolicLink()).toBe(process.platform !== "win32");`

Hopefully the tests will now work for you on windows with those changes (and linux at the same time)

packages/hub/src/utils/symlink.spec.ts
```diff
@@ -1,63 +1,89 @@
...
+import { lstat, rm } from "node:fs/promises";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+
+let failSymlink = false;
+vi.mock("node:fs/promises", async (importOriginal) => ({
+	...(await importOriginal<typeof import("node:fs/promises")>()),
+	symlink: async (...args: any[]) => {
+		if (failSymlink) {
+			failSymlink = false;
```
jeffmaury, line 15: Not sure to get why we are changing the flag here

jeffmaury (approved):
LGTM. But there are changes in XetBlob that seem unrelated

packages/hub/src/utils/symlink.spec.ts
coyotte508: It's to reset the function once it's called.  (so it doesn't impact other tests)

Could be done more elegantly

PR #1314, opened 2025-03-25
[Inference] `request()` returns a request context to avoid redundant makeRequestOptions calls

hanouticelina:
Related to @SBrandeis's comment here https://github.com/huggingface/huggingface.js/pull/1292#discussion_r2010219210. This PR addresses the original concern about redundant `makeRequestOptions` calls introduced in #1292.The solution implemented here updates the `request` function to return both the response and a _request context_ when needed, allowing provider-specific polling code to reuse this context without redundant calls to `makeRequestOptions`. 

This differs from the initial suggestion in the comment as each provider implements polling differently with different parameters / response formats. Making a generic `.poll` property would require mixing provider-specific logic into the core request function (we don't want that, right? 😄 ). 
In the end, we want to keep provider-specific logic isolated in their respective provider files (PR coming today to push that further!).


coyotte508 (commented):
This PR adds different return types, not sure the complication is worth it (unless it saves network calls)

Wauplin:
FWIW in the Python client we've deprecated the "custom requests" method to make it internal-only and gain some flexibility -which would prove useful here-.

SBrandeis (commented):
Nice!
I would be in favor of removing the `request` method from the public API as suggested by @Wauplin 
Or at least, don't make the return type conditioned by a parameter (always return `ResponseWrapper<T>`)

SBrandeis (approved):

coyotte508:
just to reiterate a bit, it's totally fine to call a "light" sync function twice to simplify things

hanouticelina:
I added a deprecation for custom requests through `request` and `streamingRequest` to be removed from the public API (I was not sure if I can just remove them directly or go through a deprecation cycle). I also made the inner request method to always return `ResponseWrapper<T>`.
Let's wait for @SBrandeis final review to see if it's still worth merging this PR or just keep the two call to `makeRequestOptions` as suggested by @coyotte508 (I'm okay with both solutions).

coyotte508:
Instead of changing the return type of `request`, you can add an extra param that is either a value to be filled or a listener.

Then `request`can fill it will the value of `makeRequestOptions`

It's really not pretty, but it works, we can doc-comment it as internal-only.

(otherwise, maybe there's a way to refactor this better)

packages/inference/README.md
```diff
@@ -572,31 +572,6 @@ await hf.tabularClassification({
...
-for await (const output of hf.streamingRequest({
-  model: 'my-custom-model',
-  inputs: 'hello world',
-  parameters: {
-    custom_param: 'some magic',
-  }
-})) {
-  ...
-}
-```
```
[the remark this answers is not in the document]
julien-c: wdym? what are the use cases for this? (and advantages over just calling fetch)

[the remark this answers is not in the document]
julien-c: And BTW we've already removed it from huggingface_hub and we try to have better Python<>TS convergence now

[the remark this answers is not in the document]
coyotte508: Being able to use `@huggingface/inference` for custom models / endpoints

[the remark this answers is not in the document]
hanouticelina: btw you can also use @huggingface/inference for custom endpoints like this:
```js
const hf = new InferenceClient(env.HF_TOKEN);
const ep = hf.endpoint("https://j3z5luu0ooo76jnl.us-east-1.aws.endpoints.huggingface.cloud/v1/");
const res = ep.chatCompletionStream({
	model: "dummy",
	messages: [
		{
			role: "user",
			content: "What is the capital of France?",
		},
	],
	max_tokens: 500,
});
```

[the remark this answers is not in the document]
julien-c: using @huggingface/inference for custom models (not related to any task) I don't think provides any value compared to `fetch`

For custom endpoints, it is supported as @hanouticelina mentions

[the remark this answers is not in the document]
coyotte508: ok

packages/inference/src/tasks/custom/request.ts
```diff
@@ -13,35 +14,6 @@ export async function request<T>(
...
-		const message = contentType?.startsWith("text/plain;") ? await response.text() : undefined;
-		throw new Error(message ?? "An error occurred while fetching the blob");
-	}
-
-	if (response.headers.get("Content-Type")?.startsWith("application/json")) {
-		return await response.json();
-	}
-
-	return (await response.blob()) as T;
+	const result = await innerRequest<T>(args, options);
```
SBrandeis, line 17: Let's log some warning message as well?

```suggestion
	console.warn("The request method is deprecated and will be removed in a future version of huggingface.js. Use specific task functions instead.")
	const result = await innerRequest<T>(args, options);
```

packages/inference/src/tasks/custom/streamingRequest.ts
```diff
@@ -15,86 +13,5 @@ export async function* streamingRequest<T>(
...
-					}
-					yield data as T;
-				}
-			}
-			events = [];
-		}
-	} finally {
-		reader.releaseLock();
-	}
+	yield* innerStreamingRequest(args, options);
```
SBrandeis, line 16:
```suggestion
    console.warn("The streamingRequest method is deprecated and will be removed in a future version of huggingface.js. Use specific task functions instead.")  
	yield* innerStreamingRequest(args, options);
```

packages/inference/src/tasks/audio/audioClassification.ts
```diff
@@ -16,10 +16,12 @@ export async function audioClassification(
...
-	const res = await request<AudioClassificationOutput>(payload, {
-		...options,
-		task: "audio-classification",
-	});
+	const res = (
+		await innerRequest<AudioClassificationOutput>(payload, {
+			...options,
+			task: "audio-classification",
+		})
+	).data;
```
SBrandeis, line 24: This syntax feels more "idiomatic"
```suggestion
	const { data: res } = await innerRequest<AudioClassificationOutput>(payload, {
		...options,
		task: "audio-classification",
	});
```

packages/inference/src/tasks/custom/request.ts
```diff
@@ -1,8 +1,9 @@
 import type { InferenceTask, Options, RequestArgs } from "../../types";
-import { makeRequestOptions } from "../../lib/makeRequestOptions";
+import { innerRequest } from "../../utils/request";
 
 /**
  * Primitive to make custom calls to the inference provider
+ * @deprecated Use specific task functions instead. This function will be removed in a future version.
```
SBrandeis, line 6: :ok_hand: nice

packages/inference/src/tasks/cv/textToVideo.ts
```diff
@@ -35,37 +34,42 @@ export async function textToVideo(args: TextToVideoArgs, options?: Options): Pro
 		args.provider === "fal-ai" || args.provider === "replicate" || args.provider === "novita"
 			? { ...omit(args, ["inputs", "parameters"]), ...args.parameters, prompt: args.inputs }
 			: args;
-	const res = await request<FalAiQueueOutput | ReplicateOutput | NovitaOutput>(payload, {
+	const response = await innerRequest<FalAiQueueOutput | ReplicateOutput | NovitaOutput>(payload, {
 		...options,
 		task: "text-to-video",
 	});
+
+	const { data, requestContext } = response;
```
SBrandeis, line 42:
```suggestion
	const { data, requestContext } = await innerRequest<FalAiQueueOutput | ReplicateOutput | NovitaOutput>(payload, {
		...options,
		task: "text-to-video",
	});
```

SBrandeis (approved):
Super nice - thanks for keeping the diff very low !!

I have some suggestion regarding syntax / code style, otherwise LGTM

packages/inference/src/tasks/cv/textToVideo.ts
SBrandeis: No need for an intermediate var here

packages/inference/src/tasks/audio/audioClassification.ts
hanouticelina: Applied this suggestion in all tasks here [68fda16](https://github.com/huggingface/huggingface.js/pull/1314/commits/68fda16ccffdf5bb91a5cf8800a05a026e90dde0)

PR #1322, opened 2025-03-31
Do not make any extra API call when endpoint URL is directly specified

### LINKED ISSUE #1321: @huggingface/inference should not call /api/tasks when endpoint url is provided
cc @hanouticelina @Wauplin @SBrandeis for viz

/api/tasks is only called to load default HF model, and should only be called when no model and no endpoint url is provided.

I think maybe originates in https://github.com/huggingface/huggingface.js/pull/1107 (where `model` could either be an url or a string)

See also https://github.com/huggingface-internal/moon-landing/pull/13175 cc @Kakulukian

coyotte508:
Fix #1321 (cc @Kakulukian for viz)

The fix is not perfect, a greater refactoring would need to take place (eg to remove all concept of `providerConfig` when `endpointUrl` is provided, use a separate default provider config maybe not relying on `model` being defined)

### DIFF
--- a/packages/inference/src/lib/makeRequestOptions.ts
+++ b/packages/inference/src/lib/makeRequestOptions.ts
@@ -35,6 +35,12 @@ export async function makeRequestOptions(
 	if (maybeModel && isUrl(maybeModel)) {
 		throw new Error(`Model URLs are no longer supported. Use endpointUrl instead.`);
 	}
+
+	if (args.endpointUrl) {
+		// No need to have maybeModel, or to load default model for a task
+		return makeRequestOptionsFromResolvedModel(args.endpointUrl, args, options);
+	}
+
 	if (!maybeModel && !task) {
 		throw new Error("No model provided, and no task has been specified.");
 	}
--- a/packages/inference/src/lib/makeRequestOptions.ts
+++ b/packages/inference/src/lib/makeRequestOptions.ts
@@ -38,7 +38,7 @@ export async function makeRequestOptions(
 
 	if (args.endpointUrl) {
 		// No need to have maybeModel, or to load default model for a task
-		return makeRequestOptionsFromResolvedModel(args.endpointUrl, args, options);
+		return makeRequestOptionsFromResolvedModel(maybeModel ?? args.endpointUrl, args, options);
 	}
 
 	if (!maybeModel && !task) {


Kakulukian (approved):
ty

packages/inference/src/types.ts
```diff
@@ -127,6 +124,6 @@ export interface UrlParams {
 export interface BodyParams {
 	args: Record<string, unknown>;
 	chatCompletion?: boolean;
-	model: string;
+	model?: string;
```
Wauplin, line 127: I would prefer if we keep `model: string` here since not passing a model should be considered the exception. Since `model` is undefined only when endpoint is passed, in which case "hf-inference"` is used, can we define a specific interface used only by `"hf-inference"`? Doing this, we won't have to touch any other providers' logic

packages/inference/src/providers/replicate.ts
```diff
@@ -25,7 +25,7 @@ const makeBaseUrl = (): string => {
 const makeBody = (params: BodyParams): Record<string, unknown> => {
 	return {
 		input: params.args,
-		version: params.model.includes(":") ? params.model.split(":")[1] : undefined,
+		version: params.model?.includes(":") ? params.model.split(":")[1] : undefined,
```
Wauplin, line 28: Typically here, we know `param.model` cannot be undefined

Wauplin (commented):
Thanks! Just a nit regarding the potentially undefined `model`.

> a greater refactoring would need to take place (eg to remove all concept of providerConfig when endpointUrl is provided, use a separate default provider config maybe not relying on model being defined)

What we do on the Python side is to use the `"hf-inference"` provider when an endpoint is passed. And in the hf-inference logic we take care of handling things correctly. If a user passes `provider="<any-provider-except-hf-inference>", endpoint="https://..."` we raise an exception.

coyotte508:
Going ahead with the bigger refacto then :)

coyotte508:
> And in the hf-inference logic we take care of handling things correctly. If a user passes `provider="<any-provider-except-hf-inference>", endpoint="https://..."` we raise an exception.

handled here: https://github.com/huggingface/huggingface.js/blob/0c9cbe63f55f165be9a13e78692093d89b4ea6aa/packages/inference/src/lib/makeRequestOptions.ts#L87-L89

packages/inference/src/providers/types.ts
```diff
@@ -0,0 +1,51 @@
...
+export namespace InferenceProviderTypes {
+	export type AuthMethod = "none" | "hf-token" | "credentials-include" | "provider-key";
+
+	export interface Config {
+		makeBaseUrl: MakeBaseUrl;
+		makeBody: MakeBody;
+		makeHeaders: MakeHeaders;
+		makeUrl: MakeUrl;
+		clientSideRoutingOnly?: boolean;
+	}
```
coyotte508, line 18: A few changes:

- Made those types "internal" (not exported via the package), and namespaced them with `InferenceProviderTypes` to not conflict with `InferenceProvider` which is an enum/string union
- Directly created types for the makeX functions, instead of for the args. That way it's simpler to create new provider or edit return types.

I can revert those changes if you want cc @Wauplin @SBrandeis @hanouticelina 

(and just keep the `ProviderConfigWithOptionalModel` addition)

packages/inference/src/lib/makeRequestOptions.ts
```diff
@@ -180,12 +212,19 @@ export function makeRequestOptionsFromResolvedModel(
...
+							model: input.resolvedModel,
+							task,
+							chatCompletion,
+					  })
+					: providerConfig.makeBody({
+							args: remainingArgs as Record<string, unknown>,
+							model: input.resolvedModel,
+							task,
+							chatCompletion,
+					  })
```
coyotte508, line 227: Exact same code as before but needed to help TS compiler see that we use an `undefined`  model only with `hf-inference` provider

Wauplin:
I'm a bit afraid this PR and https://github.com/huggingface/huggingface.js/pull/1315 will highly conflict :confused:

coyotte508:
> I'm a bit afraid this PR and #1315 will highly conflict 😕 cc @hanouticelina

How soon will it get merged?

This is a bug fix (and caused some issues in prod, see linked https://github.com/huggingface-internal/moon-landing/pull/13175)

coyotte508:
I can go back to the imperfect pre-refacto version if you want otherwise (as of [d79d5af](https://github.com/huggingface/huggingface.js/pull/1322/commits/d79d5af8e42435c08b3a30d050b58f9494a20cba)), it should conflict less

packages/inference/src/lib/makeRequestOptions.ts
```diff
@@ -35,6 +35,12 @@ export async function makeRequestOptions(
 	if (maybeModel && isUrl(maybeModel)) {
 		throw new Error(`Model URLs are no longer supported. Use endpointUrl instead.`);
 	}
+
+	if (args.endpointUrl) {
+		// No need to have maybeModel, or to load default model for a task
+		return makeRequestOptionsFromResolvedModel(args.endpointUrl, args, options);
```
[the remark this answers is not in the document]
hanouticelina: the param is mandatory for conversational models. in `huggingface_hub`, we check if the param is an url, if yes we send `model: "dummy"`, otherwise we send `params.model` in the payload. imo it's okay to do the same here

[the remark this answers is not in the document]
coyotte508: ok so either leave it as it (send url in that case) or do the bigger refacto (and change this last line to `params.model ?? "dummy"`)

SBrandeis (approved):

PR #1326, opened 2025-04-02
Fix list-files test following xet migration in prod

coyotte508:
cc @jsulz @bpronan for viz :) (nice to see xet migration in real time)


coyotte508:
There's also an `E-tag` change on the xet-migrated files :thinking: 

Will merge as is, will cherry-pick the `fileDownloadInfo` change from https://github.com/huggingface/huggingface.js/pull/1305 maybe as it may return the original etag

PR #1329, opened 2025-04-02
Intel core ultra 7

flavour:
https://huggingface.co/spaces/huggingface/HuggingDiscussions/discussions/31#67ed22a78f6ec2a57261ec51

### DIFF
--- a/packages/tasks/src/hardware.ts
+++ b/packages/tasks/src/hardware.ts
@@ -442,6 +442,9 @@ export const SKUS = {
 			"Xeon 2th Generation (Cascade Lake)": {
 				tflops: 0.55,
 			},
+			"Intel Core Ultra 7 265KF": {
+				tflops: 1.53,
+			},
 			"Intel Core 13th Generation (i9)": {
 				tflops: 0.85,
 			},


packages/tasks/src/hardware.ts
```diff
@@ -442,6 +442,9 @@ export const SKUS = {
 			"Xeon 2th Generation (Cascade Lake)": {
 				tflops: 0.55,
 			},
+			"Intel Core Ultra 7 265KF": {
+				tflops: 1.53,
```
coyotte508, line 446: it says 2TFlops here: https://nanoreview.net/en/cpu/intel-core-ultra-7-265k

flavour: The K has an integrated Arc GPU
KF does not

flavour: KF has an NPU, but I don't think we can easily use that yet.

coyotte508: hmm ok. cc @pcuenca @Vaibhavs10 for this PR

pcuenca (approved):
Looks good to me according to the calculations and measures you shared in the discussion. Thanks a lot!

PR #1332, opened 2025-04-02
[Inference Providers] Refactor: better typing?

SBrandeis:
cc @hanouticelina 

For demonstration purpose only - let's not merge

I think we can achieve better typing of the `getProviderHelper` if we reverse the mapping (`provider -> task` to `task -> provider`)

See the textToImage file (no more type casts)

Having to implement and extend 2 types when implementing a helper is quite inconvenient on the other hand - so not sure about this either

### DIFF
--- a/packages/inference/src/providers/providerHelper.ts
+++ b/packages/inference/src/providers/providerHelper.ts
@@ -1,7 +1,8 @@
-import type { ChatCompletionOutput, TextGenerationOutput } from "@huggingface/tasks";
+import type { ChatCompletionOutput, TextGenerationInput, TextGenerationOutput, TextToImageInput } from "@huggingface/tasks";
 import { InferenceOutputError } from "../lib/InferenceOutputError";
-import type { BodyParams, HeaderParams, UrlParams } from "../types";
+import type { BaseArgs, BodyParams, HeaderParams, UrlParams } from "../types";
 import { toArray } from "../utils/toArray";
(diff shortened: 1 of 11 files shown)


hanouticelina:
much better in terms of typing, thank you! 

one question:  It seems to me (and i might be wrong) the improved typing for `providerHelper.getResponse()` comes directly from the additional per-task Helpers (`TextToImageTaskHelper` and `TextGenerationTaskHelper`) as you suggested in the initial PR as well. if that's the case, what additional typing advantages reversing the main `PROVIDERS` mapping brings?

packages/inference/src/providers/providerHelper.ts
```diff
@@ -73,7 +74,36 @@ export abstract class TaskProviderHelper {
 	abstract preparePayload(params: BodyParams): unknown;
 }
 
-export class BaseConversationalTask extends TaskProviderHelper {
+export interface TextToImageTaskHelper {
+	getResponse(response: unknown,
```
hanouticelina, line 78: when making response as `unknown` here, is it still possible to declare the `getResponse` method in `BlackForestLabsTextToImageTask` with `response: BlackForestLabsResponse` ? 
same for other downstream classes that will implement `TextToImageTaskHelper`.

hanouticelina: got my answer: it should be good 😄 i was just overthinking it

Issue #1087, opened 2024-12-30
Question: Jinja2 package sandboxing


bdashore3:
Hi there,
I was looking into the @huggingface/jinja package as an alternative to nunjucks and was wondering if it's safe to use with user-defined templates. The main reason I'm asking this is because of a warning on the nunjucks website:

> nunjucks does not sandbox execution so it is not safe to run user-defined templates or inject user-defined content into template definitions. On the server, you can expose attack vectors for accessing sensitive data and remote code execution. On the client, you can expose cross-site scripting vulnerabilities even for precompiled templates (which can be mitigated with a strong [CSP](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy)). See [this issue](https://github.com/mozilla/nunjucks-docs/issues/17) for more information.

This is not a huge concern at the moment based on how I handle templates in my software, but I may need to make the switch at some point to something more secure.

If @xenova can answer, that would be very helpful. Thanks!

xenova:
Hi @bdashore3 👋 Indeed, `@huggingface/jinja` execution is sandboxed for this exact reason, so you are able to use it to parse user-defined templates.

Do note, however, that the library is not 100% feature compatible with the Jinja templating language, as it is a minimalistic re-implementation, designed specifically for chat templates.

Hopefully that answers your questions!

xenova:
> For example, a statement with type Set can't be interpreted as a SetStatement since the type isn't exposed from the package. I can try making a PR for this.

That would be great - thanks!

> Multiline set/endset: This should make it easier when setting variables that span multiple lines ([Documentation](https://jinja.palletsprojects.com/en/stable/templates/#block-assignments)). This will support tool templates such as [FireFunctionV2](https://huggingface.co/fireworks-ai/llama-3-firefunction-v2/blob/main/tokenizer_config.json)

I wasn't aware of this feature, but if it's important, feel free to open a feature request / PR! :) 

> Document which filters/features are supported: Every default filter provided by jinja isn't supported in this library, so it would be helpful to new users if they knew what subset of jinja is supported in this lib.

This is also a good idea, and could be a useful way to indicate to users what functionality to include in their templates (since in the past, we've seen some people use very niche behaviour of jinja that aren't necessarily supported by other jinja implementations, like `minijinja`. cc @Rocketknight1 for any ideas here.

Rocketknight1:
[...]

PR #1341, opened 2025-04-08
:sparkles: Add block assignment functionality in `jinja` package.

pweglik:
This PR implements functionality of block assignments (https://jinja.palletsprojects.com/en/stable/templates/#block-assignments) in lexer, parser and engine=
I felt encouraged by the reply here:
https://github.com/huggingface/huggingface.js/issues/1087#issuecomment-2597121569

It was necessary to parse some chat_templates e.g. https://huggingface.co/MadeAgents/Hammer2.1-1.5b/blob/main/tokenizer_config.json (nicely formatted `chat_template`: https://pastebin.com/xBmhbwLi)

Let me know if you have any feedback.


packages/jinja/src/runtime.ts
```diff
@@ -587,6 +587,8 @@ export class Interpreter {
 						);
 					case "join":
 						return new StringValue(operand.value.map((x) => x.value).join(""));
+					case "string":
+						return new StringValue(toJSON(operand));
```
pweglik, line 591: Not really needed for block assignment, but Hammer also used `tools | string` in template and the library missed `string` filter implementation

xenova:
Thanks so much! This is very useful and the PR looks clean 👍 

Maybe it's worth adding a unit test for a very simple test string that tests just this functionality? To be added in https://github.com/huggingface/huggingface.js/blob/1a63901ea07f706ed451216ad82d761ef964c743/packages/jinja/test/templates.test.js#L7

pweglik:
Good suggestion @xenova, I've added a simple test 👍

pweglik:
I've also added a simple test for `string` array filter.

xenova:
Flawless PR! Thanks a ton!

Failing lint is unrelated to this PR - could you merge main into your branch? I'll re-run CI and merge when all is green.

pweglik:
I've already `pweglik:main` rebased onto `huggingface:main`
Thanks a lot 😄

xenova (approved):
Ah yes you are right! Fixed by https://github.com/huggingface/huggingface.js/pull/1348 (& ran `pnpm format`).

Thanks again.

PR #1350, opened 2025-04-11
[Inference] top-down injection of providerHelper

SBrandeis:
cc @hanouticelina @Wauplin 

Prompted by: https://huggingface.slack.com/archives/C02EMARJ65P/p1744370608755449 (internal)

This PR reduces the number of places where we call `getProviderHelper` by enforcing it's passed as an argument by the caller when needed

Because `getProviderHelper` is fallible, calling it from different places (sometimes deep in the function call chain) can be dangerous and result in unexpected bugs.

This PR reduces this risk by reducing the number of different places where we call `getProviderHelper` , and instead pass it as an argument top-down

It also has the benefits of clearly indicating which functions depend on provider-specific logic


hanouticelina (approved):
thank you for fixing this! it's indeed much safer to control where the `getProviderHelper` function is called

julien-c (approved):
wynaut

Wauplin (commented):
ok ok :+1:

PR #1351, opened 2025-04-11
PoC: `InferenceClient` is also a `McpClient`

julien-c:
## Required reading

https://modelcontextprotocol.io/quickstart/client

TL;DR: MCP is a standard API to expose sets of Tools that can be hooked to LLMs

## Summary of how to use this

You can either use McpClient, or you can run an example Agent directly:

# Tiny Agent

We now have a tiny Agent (a while loop, really) in this PR, built on top of the MCP Client. You can run it like this:

```bash
cd packages/mcp-client
pnpm run agent
```

# `McpClient` i.e. the underlying class

```ts
const client = new McpClient({
	provider: "together",
	model: "Qwen/Qwen2.5-72B-Instruct",
	apiKey: process.env.HF_TOKEN,
});

await client.addMcpServer({
	// Filesystem "official" mcp-server with access to your Desktop
	command: "npx",
	args: ["-y", "@modelcontextprotocol/server-filesystem", join(homedir(), "Desktop")],
});
```

## Variant where we call a custom, local MCP server

```ts
await client.addMcpServer(
	"node",
	["--disable-warning=ExperimentalWarning", join(homedir(), "Desktop/hf-mcp/index.ts")],
	{
		HF_TOKEN: process.env.HF_TOKEN,
	}
);

const response = await client.processQuery(`
	find an app that generates 3D models from text,
	and also get the best paper about transformers
`);
```
#### Where to find the MCP Server used here as an example

https://gist.github.com/julien-c/0500ba922e1b38f2dc30447fb81f7dc6

(Note that you can replace it with any MCP Server, from this doc for instance: https://modelcontextprotocol.io/examples)

## Python version

Python version will be implemented in `huggingface_hub` in this PR: https://github.com/huggingface/huggingface_hub/pull/2986

Contributions are welcome!


packages/inference/package.json
```diff
@@ -55,8 +55,9 @@
 		"export-templates": "tsx scripts/export-templates.ts"
 	},
 	"dependencies": {
+		"@huggingface/jinja": "workspace:^",
 		"@huggingface/tasks": "workspace:^",
-		"@huggingface/jinja": "workspace:^"
+		"@modelcontextprotocol/sdk": "^1.9.0"
```
coyotte508, line 60: If it's not too troublesome, it would be nice to either reimplement what's needed or vendor part of the code eg as we did with `fetch-event-source`: https://github.com/huggingface/huggingface.js/blob/b033ad193002c32091e196bf4936b28450eaae0a/packages/inference/src/vendor/fetch-event-source/parse.ts

See https://www.npmjs.com/package/@modelcontextprotocol/sdk?activeTab=dependencies this adds 10 sub-dependencies including `express`

julien-c: sorry i should have mentioned, i'll publish this as another sub-package @coyotte508

grll (commented):
https://github.com/huggingface/huggingface_hub/pull/2986 few of my comments there are also relevant here. Probably major one is I think you could add SSE support quite easily. 

Also very minor but first time I see typescript with 4 indents, but why not.

coyotte508:
> Also very minor but first time I see typescript with 4 indents, but why not.

It uses tabs, you can set your tab width to whatever you want :) (2, 4 or 8)

grll:
Ah the good old tab vs space conundrum then :)

julien-c:
At this point we should let LLMs decide between tabs and spaces once and for all

grll:
> At this point we should let LLMs decide between tabs and spaces once and for all

2025-04-17-134859_886x333_scrot

Claude has spoken

packages/tasks/src/tasks/chat-completion/inference.ts
```diff
@@ -130,7 +130,8 @@ export interface ChatCompletionInputToolCall {
 	[property: string]: unknown;
 }
 export interface ChatCompletionInputFunctionDefinition {
-	arguments: unknown;
+	/// TODO: fix upstream
+	parameters: unknown;
```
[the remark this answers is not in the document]
julien-c: done in #1367 (caution, you might neeed to clean your eyes after looking at it)

.github/workflows/mcp-client-publish.yml
```diff
@@ -0,0 +1,72 @@
...
+          # Update package.json with the new version
+          node -e "const fs = require('fs'); const package = JSON.parse(fs.readFileSync('./package.json')); package.version = '$BUMPED_VERSION'; fs.writeFileSync('./package.json', JSON.stringify(package, null, '\t') + '\n');"
+          pnpm --filter doc-internal run fix-cdn-versions
+          git add ../..
+          git commit -m "🔖 @huggingface/mcp-client $BUMPED_VERSION"
+          git tag "mcp-client-v$BUMPED_VERSION"
+
+      # Add checks for dependencies if needed, similar to hub-publish.yml
+      # - name: "Check Deps are published before publishing this package"
+      #   run: pnpm -w check-deps <dependency_name>
```
coyotte508, line 53:
```suggestion
      - name: "Check Deps are published before publishing this package"
        run: pnpm -w check-deps inference && pnpm -w check-deps tasks
```

packages/mcp-client/package.json
```diff
@@ -0,0 +1,59 @@
...
+		"mcp",
+		"client",
+		"hugging",
+		"face"
+	],
+	"author": "Hugging Face",
+	"license": "MIT",
+	"devDependencies": {
+		"@types/node": "^22.14.1"
+	},
```
coyotte508, line 53: Maybe you can upgrade the root package.json with this instead

julien-c: (done)

julien-c:
Ok this is ready for review!

Note, there's now a small `Agent.ts` class, and an example agent you can run with:

```bash
cd packages/mcp-client
pnpm run agent
```

Wauplin:
Not sure I am using this the expected way but I had to tweak a bit to run `hf-mcp.ts`:

1. Added some dev dependencies to the project

```js
	"devDependencies": {
		"@huggingface/hub": "workspace:^",
		"ts-node": "^10.9.2",
		"typescript": "^5.8.3",
		"zod": "^3.24.3"
	}
```

2. Fixed a compile error  in `hf-mcp.ts` (see https://gist.github.com/julien-c/0500ba922e1b38f2dc30447fb81f7dc6?permalink_comment_id=5553466#gistcomment-5553466)
3. Added `"hf-mcp": "npx ts-node hf-mcp.ts"` as official script to package.json (this is only for easier testing for me)
4. Updated in `cli.ts` to run the command:
```js
		command: "pnpm",
		args: ["run", "hf-mcp"],
```
5. (also had to hardcode the authorized path for filesystem MCP server-filesystem)


Then with everything setup I ran `pnpm agent` with prompt

```
List packages implemented under ~/projects/huggingface.js and make a nice looking HTML page listing them shortly. No need to provide further information per package. Once you have that, publish this HTML page as a static Space named "list-huggingface.js-packages"
```

(yup, not very inspired here)

=> resulting in https://huggingface.co/spaces/Wauplin/list-huggingface.js-packages

julien-c:
@Wauplin I should have updated the PR description because `hf-mcp.ts` (an example MCP server) is quite orthogonal to the PR. I'll update the PR desc. right now.

Your agent's Generated Space looks cool though 😍 cc @gary149:

image

packages/mcp-client/src/McpClient.ts
```diff
@@ -0,0 +1,132 @@
...
+import { debug } from "./utils";
+
+type ToolName = string;
+
+export interface ChatCompletionInputMessageTool extends ChatCompletionInputMessage {
+	role: "tool";
+	tool_call_id: string;
+	content: string;
+	name?: string;
+}
```
julien-c, line 21: missing type in our spec probably @hanouticelina

hanouticelina: isn't that [ChatCompletionOutputTextMessage](https://github.com/huggingface/huggingface.js/blob/f37b0aaa079cfdcab29ff600cad54501b829316c/packages/tasks/src/tasks/chat-completion/spec/output.json#L120) ?

julien-c: well: 
- it's not in our generated types (inference.ts)
- and it's weird to me to use a output message type as an input message
wdyt?

hanouticelina: ok ok you're right, I'd rather define another type indeed, but I'd keep this type defined here instead of putting it in our specs if that's okay for you

julien-c: yes that works

packages/mcp-client/src/McpClient.ts
```diff
@@ -0,0 +1,132 @@
...
+		}
+		for (const toolCall of toolCalls) {
+			const toolName = toolCall.function.name;
+			const toolArgs = JSON.parse(toolCall.function.arguments);
+
+			const message: ChatCompletionInputMessageTool = {
+				role: "tool",
+				tool_call_id: toolCall.id,
+				content: "",
+				name: toolName,
```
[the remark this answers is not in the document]
julien-c: maybe not for the LLM inference, but i'm using it in my agent's outer loop (in cli.ts) and i think it can't hurt anyways

PR #1353, opened 2025-04-13
Add Intel Core 14th Generation (i7)

1TSnakers:
I Just added my CPU, work still needs to be done to cover intel's 14th gen CPUs.

### DIFF
--- a/packages/tasks/src/hardware.ts
+++ b/packages/tasks/src/hardware.ts
@@ -489,6 +489,9 @@ export const SKUS = {
 			"Intel Core Ultra 7 265KF": {
 				tflops: 1.53,
 			},
+			"Intel Core 14th Generation (i7)": {
+				tflops: 1.13,
+			},
 			"Intel Core 13th Generation (i9)": {
 				tflops: 0.85,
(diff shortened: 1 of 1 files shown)


packages/tasks/src/hardware.ts
```diff
@@ -489,6 +489,9 @@ export const SKUS = {
 			"Intel Core Ultra 7 265KF": {
 				tflops: 1.53,
 			},
+			"Intel Core 14th Generation (i7)": {
+				tflops: 1.13,
```
Vaibhavs10, line 493: Hi @1TSnakers - Thanks for the PR where are you getting these values for? 

I get 0.8 TFLOP (ref: https://nanoreview.net/en/cpu/intel-core-i7-14700hx)

1TSnakers:
i made some code to test it

if you want to change it, you can!

packages/tasks/src/hardware.ts
Vaibhavs10, line 493:
```suggestion
				tflops: 0.8,
```

PR #1359, opened 2025-04-15
[Inference] Make CI green on main

SBrandeis:
Best reviewed commit per commit

- Some missing code for HF Inference tests
- Run the format and lint scripts


Wauplin (commented):
Thanks :+1:

PR #1364, opened 2025-04-17
[Inference] fix feature extraction (embeddings) for sambanova

hanouticelina:
This PR fixes the widget of https://huggingface.co/intfloat/e5-mistral-7b-instruct 

i didn’t want to introduce a breaking change in this PR, but we should consider adding support for the OpenAI [Embeddings API](https://platform.openai.com/docs/api-reference/embeddings). Other providers like [Fireworks AI](https://docs.fireworks.ai/guides/querying-embeddings-models) and [Together](https://docs.together.ai/docs/embeddings-overview) also host embedding models with OpenAI-compatible endpoints. AFAIK, TEI supports the OAI Embeddings API as well.


Wauplin (approved):
:+1: 

Re: OpenAI-compatible API, agree with it yes

PR #1365, opened 2025-04-17
[tasks] `{{QUANT_TAG}}` placeholder for all snippets

krampstudio:
cc @Vaibhavs10 @ngxson


packages/tasks/src/local-apps.ts
```diff
@@ -94,14 +94,19 @@ function isMlxModel(model: ModelData) {
...
+function getQuantTag(filepath: string | undefined): string {
+	let tagName = ":{{QUANT_TAG}}";
+
 	if (filepath) {
 		const quantLabel = parseGGUFQuantLabel(filepath);
-		tagName = quantLabel ? `:${quantLabel}` : "";
+		tagName = quantLabel ? `:${quantLabel}` : tagName;
 	}
+	return tagName;
+}
```
mishig25, line 105: maybe bit better? 

```suggestion
function getQuantTag(filepath?: string): string {
  const defaultTag = ":{{QUANT_TAG}}";
  
  if (!filepath) {
    return defaultTag;
  }
  
  const quantLabel = parseGGUFQuantLabel(filepath);
  return quantLabel ? `:${quantLabel}` : defaultTag;
}
```

mishig25 (approved):

packages/tasks/src/local-apps.ts
krampstudio: see https://github.com/huggingface/huggingface.js/pull/1365/commits/b5af43a78709dda5c6862da271f778e4bbab836b

PR #1369, opened 2025-04-19
Add NVIDIA RTX 5060, RTX 5060 Ti models to hardware.ts

stelterlab:
Please add those to new NVIDIA RTX 50 series cards which were recently released so they can be shown in the users profile. ;-)

Sources used:

- https://www.techpowerup.com/gpu-specs/geforce-rtx-5060-ti.c4246
- https://www.techpowerup.com/gpu-specs/geforce-rtx-5060.c4219


packages/tasks/src/hardware.ts
```diff
@@ -152,6 +152,14 @@ export const SKUS = {
 				tflops: 43.94,
 				memory: [16],
 			},
+			"RTX 5060 Ti": {
+				tflops: 23.70, // source https://www.techpowerup.com/gpu-specs/geforce-rtx-5060-ti.c4246
+				memory: [16],
```
pcuenca, line 157:
```suggestion
				memory: [16, 8],
```

There are variants with 8 and 16 GB, I think.

pcuenca (approved):
Thank you!

PR #1375, opened 2025-04-22
[Inference] Remove inference vcr tests

hanouticelina:
same as https://github.com/huggingface/huggingface_hub/pull/3021. 
I’ve just skipped the tests for now and removed VCR related code so that we can quickly re-enable or adapt them when we introduce proper inference tests, as mentioned by @Wauplin.


Wauplin (approved):
Will require more manual testing when adding new providers but in general agree with removing the current burden induced by the tapes (I feel that we've wasted more time compared to the benefits so far).

I still think a good way to make the VCR tests (or cached tests) dynamic and maintainable would be to a either server-side tests or a cache proxy that would work as the VCRs but dynamically updated and never versioned.

SBrandeis (approved):

PR #1376, opened 2025-04-22
add-dia-snippet

Deep-unlearning:
I've open a pr to add the library for this model: https://huggingface.co/nari-labs/Dia-1.6B
here is the link to the pr: https://huggingface.co/nari-labs/Dia-1.6B/discussions/10

### DIFF
--- a/packages/tasks/src/model-libraries-snippets.ts
+++ b/packages/tasks/src/model-libraries-snippets.ts
@@ -226,6 +226,17 @@ infer = loaded_model.signatures["serving_default"]
 print(infer(inputs=tf.constant([input_tensor])))`,
 ];
 
+export const dia = (model: ModelData): string[] => [
+	`import soundfile as sf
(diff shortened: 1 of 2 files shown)


Wauplin (approved):
Nice! I see that the PR has already been merged, which is good :)

Annoyingly, 2 other unrelated models are listed under https://huggingface.co/models?other=dia because they have `dia` as a language (for https://huggingface.co/espnet/xeus) and because of a tag in this one https://huggingface.co/petrioteer/dia-convo-v1.2c... Might be good to try to open PRs to kindly ask them to remove the `dia` tags? Not blocking for a merge but would still be nice for consistency.

**EDIT:** Actually for https://huggingface.co/espnet/xeus it will really be annoying because the `espnet` library might not be recognized anymore (and it's a quite important model :confused:). Will check what we can do server-side so that languages are not taken into account for library resolution.

PR #1382, opened 2025-04-24
Fix hfjs `chatCompletionStream`  snippet

julien-c:
there's no `await` here

~~I don't remember if i need to run a script to check CI~~

EDIT: nvm found it


SBrandeis (approved):

Wauplin (commented):
:+1:

PR #1386, opened 2025-04-25
Add MAGI-1 library

jiahy0825:
This PR adds MAGI-1 library


jiahy0825:
> Hi @jiahy0825 - Thanks for the PR, couple of quick notes:
> 
> 1. I see that instead of one model per repo, there are multiple different models in the same repo: [huggingface.co/sand-ai/MAGI-1/tree/main/ckpt/magi](https://huggingface.co/sand-ai/MAGI-1/tree/main/ckpt/magi)
> 2. Since all the model repos are diffusers repos, the download tracking should automatically work if these checkpoints were split in 3 repos.
> 
> Let me know if you have any questions.

Hi, thanks a lot for your reply!

We would like to keep all these models in a single repo because our model pipeline is relatively complex — it involves multiple components like T5, VAE, and others. Having all related models organized within one repo would make it much easier for users to find and use them together.

Therefore, we would prefer to track downloads through this repo via the PR if possible.

Please let me know if you have any better suggestions. Thanks again for your help!

Vaibhavs10 (commented):
in this case I'd defer to @Wauplin if we can make an exception. In general, we do strongly recommend having separate repos for separate models!

packages/tasks/src/model-libraries.ts
```diff
@@ -508,6 +508,12 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		repoName: "mindspore",
 		repoUrl: "https://github.com/mindspore-ai/mindspore",
 	},
+	"magi-1": {
+		prettyLabel: "MAGI-1",
+		repoName: "MAGI-1",
+		repoUrl: "https://github.com/SandAI-org/MAGI-1",
+		countDownloads: `path_extension:"json"`,
```
Wauplin, line 515: Doing this will likely double-count many downloads since there are several .json files needed to instantiate a model. What we usually do is to track downloads on a single file that is essential for the model to be loaded correctly. Could https://huggingface.co/sand-ai/MAGI-1/blob/main/ckpt/vae/config.json be used for that for instance?

packages/tasks/src/model-libraries.ts
```diff
@@ -508,6 +508,12 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		repoName: "mindspore",
 		repoUrl: "https://github.com/mindspore-ai/mindspore",
 	},
+	"magi-1": {
```
Wauplin, line 511: as @Vaibhavs10 mentioned, you'll need to set `library_name: magi-1` in your model card metadata to make this work

Wauplin (commented):
Sorry for the delayed review. Agree with @Vaibhavs10 on the fact that it's usually best to split the different variants of a models in separate repos. If you don't want to reupload all t5/vae weights to each separate repo, you can update the Magi library to load the weights from individual repos except for the t5/vae backbones that would have to be fetched from the "main" MAGI-1 repo. By doing this, you'll have separate download counts for each variants (24B, 24B_distill, 24B_distill_quant, 4.5B). Also it would make them more easily discoverable on the Hub for the users. 

But note that spliting repos is only a strong recommendation, not a hard requirements. If for your needs you believe it makes more sense to keep everything together then, let's keep it like this! Only the download count rule has to be updated to count only "1" when a user instanciate the model.

packages/tasks/src/model-libraries.ts
jiahy0825: Thanks for your advice and I have changed the path to `ckpt/vae/config.json`

packages/tasks/src/model-libraries.ts
jiahy0825: Thanks for your advice and I have set `library_name: "magi-1"` in the model card. Does the appearance of the symbol in the image mean the settings are correct?

image

jiahy0825:
@Wauplin @Vaibhavs10 Thank you both very much for your help. Could I trouble you to review it once more?

packages/tasks/src/model-libraries.ts
Wauplin: yes!

PR #1398, opened 2025-04-29
handle folders in uploadFiles

coyotte508:
If we want to expose a CLI for the JS client...


coyotte508:
test failings unrelated to this PR, tests added in this PR pass

PR #1404, opened 2025-04-30
Pass an `endpointUrl` directly to `InferenceClient`

julien-c:
in the end i kept the `endpointUrl` var name so the diff is not massive.

I think keeping that name is ok.

### DIFF
--- a/packages/inference/src/InferenceClient.ts
+++ b/packages/inference/src/InferenceClient.ts
@@ -18,7 +18,12 @@ export class InferenceClient {
 	private readonly accessToken: string;
 	private readonly defaultOptions: Options;
 
-	constructor(accessToken = "", defaultOptions: Options = {}) {
+	constructor(
+		accessToken = "",
+		defaultOptions: Options & {
+			endpointUrl?: string;
+		} = {}
+	) {
 		this.accessToken = accessToken;
 		this.defaultOptions = defaultOptions;
 
@@ -27,7 +32,10 @@ export class InferenceClient {
 				enumerable: false,
 				value: (params: RequestArgs, options: Options) =>
 					// eslint-disable-next-line @typescript-eslint/no-explicit-any
-					fn({ ...params, accessToken } as any, { ...defaultOptions, ...options }),
+					fn({ endpointUrl: defaultOptions.endpointUrl, accessToken, ...params } as any, {
+						...defaultOptions,
+						...options,
+					}),
 			});
 		}
 	}
--- a/packages/inference/src/InferenceClient.ts
+++ b/packages/inference/src/InferenceClient.ts
@@ -33,13 +33,12 @@ export class InferenceClient {
 	}
 
 	/**
-	 * Returns instance of InferenceClient tied to a specified endpoint.
+	 * Returns a new instance of InferenceClient tied to a specified endpoint.
 	 *
 	 * For backward compatibility mostly.
 	 */
-	public endpoint(endpointUrl: string): this {
-		this.defaultOptions.endpointUrl = endpointUrl;
-		return this;
+	public endpoint(endpointUrl: string): InferenceClient {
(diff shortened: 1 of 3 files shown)


julien-c:
CI failure is unrelated as far as i can tell

packages/inference/src/InferenceClient.ts
```diff
@@ -40,34 +33,25 @@ export class InferenceClient {
...
-		for (const [name, fn] of Object.entries(tasks)) {
-			Object.defineProperty(this, name, {
-				enumerable: false,
-				value: (params: RequestArgs, options: Options) =>
-					// eslint-disable-next-line @typescript-eslint/no-explicit-any
-					fn({ ...params, accessToken, endpointUrl } as any, { ...defaultOptions, ...options }),
-			});
-		}
+	public endpoint(endpointUrl: string): this {
+		this.defaultOptions.endpointUrl = endpointUrl;
```
coyotte508, line 41: I don't think it should override this, it should return a new instance. 

Eg if someone has:

```ts
const hf = new InferneceClient(token);

const endpoint1 = hf.endpoint("...");
const endpoint2 = hf.endpoint("...");
```

it will break stuff.

Honestly this PR should just add the possibility to add `endpointsUrl` in the constructor of `InferenceClient`, and pass it in the `args` in that case, that 's all.

packages/inference/src/types.ts
```diff
@@ -33,6 +33,11 @@ export interface Options {
 	 * Requests can only be billed to an organization the user is a member of, and which has subscribed to Enterprise Hub.
 	 */
 	billTo?: string;
+
+	/**
+	 * Optional: override the default router.huggingface.co Inference Providers endpoint with any baseURL.
+	 */
+	endpointUrl?: string;
```
coyotte508, line 40: I don't think that works without changes to `makeRequestOptions`, `endpointUrl` is in the `Args` not in the `Options`

packages/inference/src/InferenceClient.ts
julien-c: but ok for returning a new instance, i'll make the change

julien-c: f86dbea25fb442c68e5729a70c4087cf9e8a51e7

packages/inference/src/types.ts
julien-c: good catch => 5d352ac8d86ef2846f371ee3f80995bf353af1a3

packages/inference/src/InferenceClient.ts
```diff
@@ -34,40 +32,33 @@ export class InferenceClient {
 				enumerable: false,
 				value: (params: RequestArgs, options: Options) =>
 					// eslint-disable-next-line @typescript-eslint/no-explicit-any
-					fn({ ...params, accessToken } as any, { ...defaultOptions, ...options }),
+					fn({ endpointUrl: defaultOptions.endpointUrl, accessToken, ...params } as any, {
```
julien-c, line 35: i switched the order here btw @coyotte508 so that method params take precedence over the instance's props, which is what we always want i assume?

coyotte508: sure

SBrandeis (approved):
Looks good to me!

coyotte508 (approved):

packages/inference/src/InferenceClient.ts
```diff
@@ -34,40 +32,33 @@ export class InferenceClient {
...
+						...defaultOptions,
```
coyotte508, line 36: should probably remove `endpointUrl` here though

```suggestion
						...omit(defaultOptions, ["endpointUrl"]),
```

hanouticelina (approved):
Looks good, thanks!

PR #1407, opened 2025-04-30
Change E2E from using summarization to chatCompletion

coyotte508:


e2e/deno/index.ts
```diff
@@ -15,14 +15,13 @@ if (token) {
 	const tokenInfo = await whoAmI({ credentials: { accessToken: token } });
 	console.log(tokenInfo);
 
-	const sum = await hf.summarization({
-		model: "facebook/bart-large-cnn",
-		inputs:
-			"The tower is 324 metres (1,063 ft) tall, about the same height as an 81-storey building, and the tallest structure in Paris. Its base is square, measuring 125 metres (410 ft) on each side. During its construction, the Eiffel Tower surpassed the Washington Monument to become the tallest man-made structure in the world, a title it held for 41 years until the Chrysler Building in New York City was finished in 1930.",
+	const result = await hf.chatCompletion({
+		model: "microsoft/Phi-3-mini-4k-instruct",
```
coyotte508, line 19: Used a random model supported by hf-inference which seemed small enough

coyotte508:
Merging to fix the CI, feel free to follow-up

e2e/deno/index.ts
hanouticelina: I believe the strategy for supported models by` hf-inference` is to refresh the list daily, not sure if this applies to chat completion models too — if it does, we can’t guarantee that phi-3 will remain available through hf-inference. cc @julien-c to confirm.
so maybe it's better to test with an external provider?

```suggestion
		model: "meta-llama/Llama-3.2-1B-Instruct",
		provider: "nebius"
```
(meta-llama/Llama-3.2-1B-Instruct is even smaller than phi-3-mini)

hanouticelina (commented):
thanks for fixing this! i left a comment related to the choice of the provider/model

e2e/deno/index.ts
coyotte508: yes would be nice to just have a default model and be able to remove the `model` param :sob: 

cc @julien-c

coyotte508: by the way @hanouticelina do we need to set `provider: "nebius"`? Or does the "auto" mode work?

hanouticelina: yes using the default value (`provider: "auto"`) is fine and you're right let's use that instead.
("auto" will select the first provider available for the model, sorted by the user's order in https://hf.co/settings/inference-providers)

PR #1409, opened 2025-04-30
use smaller model instead for e2es

coyotte508:
https://github.com/huggingface/huggingface.js/pull/1407#discussion_r2068601267


hanouticelina (approved):

Wauplin (approved):
:+1: 

I just pushed https://github.com/huggingface/huggingface.js/pull/1414 which should also make the test quicker with a shorter response. PRs shouldn't conflict.

PR #1410, opened 2025-04-30
InferenceClient methods have the proper types now

SBrandeis:
# TL;DR

Types of `InferenceClient` methods are now correct 🥳 

```typescript
const client = new InferenceClient();

const output = client.textToImage(
    {
        inputs: "test input",
        parameters: {
            /// Correctly typed! And auto-completion works!
        },
    },
    options
)
```

### DIFF
--- a/packages/inference/src/InferenceClient.ts
+++ b/packages/inference/src/InferenceClient.ts
@@ -1,19 +1,12 @@
 import * as tasks from "./tasks";
-import type { Options, RequestArgs } from "./types";
-import type { DistributiveOmit } from "./utils/distributive-omit";
+import type { Options } from "./types";
 import { omit } from "./utils/omit";
+import { typedEntries } from "./utils/typedEntries";
 
 /* eslint-disable @typescript-eslint/no-empty-interface */
 /* eslint-disable @typescript-eslint/no-unsafe-declaration-merging */
 
 type Task = typeof tasks;
-
-type TaskWithNoAccessToken = {
-	[key in keyof Task]: (
-		args: DistributiveOmit<Parameters<Task[key]>[0], "accessToken">,
-		options?: Parameters<Task[key]>[1]
-	) => ReturnType<Task[key]>;
-};
 
 export class InferenceClient {
 	private readonly accessToken: string;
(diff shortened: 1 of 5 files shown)


packages/inference/test/InferenceClient.spec.ts
SBrandeis, line 1: HF Inference API did not match the JSON schema spec it seems :0

packages/inference/src/InferenceClient.ts
```diff
@@ -28,15 +21,19 @@ export class InferenceClient {
...
 			Object.defineProperty(this, name, {
 				enumerable: false,
-				value: (params: RequestArgs, options: Options) =>
+				value: (params: Parameters<typeof fn>[0], options: Parameters<typeof fn>[1]) =>
 					// eslint-disable-next-line @typescript-eslint/no-explicit-any
-					fn({ endpointUrl: defaultOptions.endpointUrl, accessToken, ...params } as any, {
-						...omit(defaultOptions, ["endpointUrl"]),
-						...options,
-					}),
+					(fn as any)(
```
julien-c, line 29: is the "as any" actually necessary or could we define the type of fn somewhere?

julien-c: ah ok i see the comment below

packages/inference/src/lib/getInferenceProviderMapping.ts
```diff
@@ -112,7 +112,10 @@ export async function resolveProvider(
 			throw new Error("Specifying a model is required when provider is 'auto'");
 		}
 		const inferenceProviderMapping = await fetchInferenceProviderMappingForModel(modelId);
-		provider = Object.keys(inferenceProviderMapping)[0] as InferenceProvider;
+		provider = Object.keys(inferenceProviderMapping)[0] as InferenceProvider | undefined;
+	}
+	if (!provider) {
+		provider = "hf-inference";
```
julien-c, line 118: "auto" no?

packages/mcp-client/src/McpClient.ts
```diff
@@ -48,7 +48,7 @@ export class McpClient {
 		apiKey: string;
 	}) {
 		this.client = baseUrl ? new InferenceClient(apiKey).endpoint(baseUrl) : new InferenceClient(apiKey);
-		this.provider = provider;
+		this.provider = provider ?? "auto";
```
julien-c, line 51: it should be optional imo

julien-c (approved):

packages/inference/src/lib/getInferenceProviderMapping.ts
hanouticelina: we already falling back to "auto" line 104, if provider is undefined here it means inferenceProviderMapping is empty and I'd rather raise an error instead

Wauplin (approved):
thanks for fixing this, still impressed what TS can do for typing :)

PR #1413, opened 2025-04-30
fix(mcp): use a more generic folder to put mcp filesystem root

mfuntowicz:
Currently the MCP server attempts to access a folder under `${HOMEDIR}/Desktop` which is not portable accross different OSes.

This PR proposes to change the `Desktop` part to `.mcp` which is OS agnostic and ensure the folder exists when spawning.


Wauplin (commented):
:+1: about not defaulting to `~/Desktop` but not sure `~/.mcp` is really relevant either (it would only be able to work on new files, not existing ones).

Another possibility is to default to `~/.mcp` and make it a optional parameter of the CLI.

packages/mcp-client/cli.ts
```diff
@@ -12,12 +13,13 @@ import { version as packageVersion } from "./package.json";
...
 const PROVIDER = (process.env.PROVIDER as InferenceProvider) ?? "nebius";
 const BASE_URL = process.env.BASE_URL;
+const MCP_LOCAL_FOLDER = join(homedir(), ".mcp");
 
 const SERVERS: StdioServerParameters[] = [
 	{
 		// Filesystem "official" mcp-server with access to your Desktop
 		command: "npx",
-		args: ["-y", "@modelcontextprotocol/server-filesystem", join(homedir(), "Desktop")],
+		args: ["-y", "@modelcontextprotocol/server-filesystem", MCP_LOCAL_FOLDER],
```
julien-c, line 22: i would rather still use Desktop on mac OS and just `homedir()` on Win32, can you switch to this?

You can use `process.platform === "darwin"` to figure out 

Thanks!

julien-c:
i can finish your PR if you want @mfuntowicz? let me know

mfuntowicz:
I've put `~/.mcp` to have something which can be easily implemented without branching on plateform. 
@julien-c if we are fine with the branching, I totally think we can do something closer to the original for all platform. 


Happy to take a snab at it 👍🏻

julien-c:
yes, branching depending on OS is completely fine

julien-c (approved):

PR #1421, opened 2025-05-02
gguf : add findNearestQuantType

ngxson:
In this PR:
- Move `GGMLFileQuantizationType` to `tasks`
- Update the list of `GGMLFileQuantizationType` (NOTE: a **File** can contains multiple quants, for example Q4_K_M **File** is Q4_K + Q6_K)
- For `GGMLQuantizationType`, add TQ1_0 and TQ2_0 tenary quants
- Add `findNearestQuantType` (see below)

## findNearestQuantType

This function is useful for `/v2` registry endpoint with **text + vision models**, in case we want to pick the correspond vision model that can be paired with a text model.

The main issue is that text model can go lower than Q4, like Q3/2/1, but it is **not** the case for vision model, as vision model is quite sensitive to quantization.

On @bartowski1182 , most of vision models will have BF16, F16 and maybe Q8_0 versions. The idea is:
- If user pick BF16/F16/Q8_0 text model, we pair it to the correspond BF16/F16/Q8_0
- If user pick something else, like Q4_K_M, we find the nearest quant to be paired with. It's Q8_0 in this case

### DIFF
--- a/packages/gguf/src/gguf.spec.ts
+++ b/packages/gguf/src/gguf.spec.ts
@@ -7,6 +7,8 @@ import {
 	ggufAllShards,
 	parseGgufShardFilename,
 	parseGGUFQuantLabel,
+	GGUF_QUANT_ORDER,
+	findNearestQuantType,
 } from "./gguf";
 import fs from "node:fs";
 
@@ -46,7 +48,7 @@ describe("gguf", () => {
 			tensor_count: 291n,
 			kv_count: 19n,
 			"general.architecture": "llama",
-			"general.file_type": GGMLFileQuantizationType.MOSTLY_Q2_K,
+			"general.file_type": GGMLFileQuantizationType.Q2_K,
 			"general.name": "LLaMA v2",
 			"general.quantization_version": 2,
 			"llama.attention.head_count": 32,
@@ -105,7 +107,7 @@ describe("gguf", () => {
 			tensor_count: 291n,
 			kv_count: 24n,
 			"general.architecture": "llama",
-			"general.file_type": GGMLFileQuantizationType.MOSTLY_Q5_K_M,
+			"general.file_type": GGMLFileQuantizationType.Q5_K_M,
 			"general.name": "mistralai_mistral-7b-instruct-v0.2",
 			"general.quantization_version": 2,
 			"llama.attention.head_count": 32,
@@ -143,7 +145,7 @@ describe("gguf", () => {
 			tensor_count: 164n,
 			kv_count: 21n,
 			"general.architecture": "gemma",
-			"general.file_type": GGMLFileQuantizationType.MOSTLY_Q4_K_M,
+			"general.file_type": GGMLFileQuantizationType.Q4_K_M,
 			"general.name": "gemma-2b-it",
 			"general.quantization_version": 2,
 			"gemma.attention.head_count": 8,
@@ -180,7 +182,7 @@ describe("gguf", () => {
 			tensor_count: 197n,
 			kv_count: 23n,
 			"general.architecture": "bert",
-			"general.file_type": GGMLFileQuantizationType.MOSTLY_F16,
+			"general.file_type": GGMLFileQuantizationType.F16,
 			"general.name": "bge-small-en-v1.5",
 			"bert.attention.causal": false,
 			"bert.attention.head_count": 12,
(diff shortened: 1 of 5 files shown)


packages/tasks/src/gguf.ts
```diff
@@ -1,3 +1,155 @@
...
+	// 1-bit quantizations
+	GGMLFileQuantizationType.IQ1_S,
+	GGMLFileQuantizationType.IQ1_M,
+	GGMLFileQuantizationType.TQ1_0,
+	GGMLFileQuantizationType.TQ2_0,
+];
+
+// This function finds the nearest quantization type that is less than or equal to the given quantization type.
+// It returns undefined if no such quantization type is found.
+export function findNearestQuantType(
```
ngxson, line 112: fully disclosure: this function is writted by gemini 2.5 pro 😂

ngxson:
> Is it necessarily true that a user may want to save ~400mb on the vision part (by going to Q8_0) if they choose a smaller quant?

Hmm yeah that's a valid question, I usually use Q8_0 when developing locally because kernels for Q8_0 are significantly faster than F16.

Even without that, saving 400MB is also significant (Ofc this is quite subjective, but personally 400MB is big for a vision model 😂 )

> Though I'm guessing this may be an optional flag?

I don't think it's needed, because if you think the F16 vs Q8_0 is not diff too much, you can simply do not produce quant for it.

For example an if you have 400MB model in F16, then converting to Q8 will save 200MB, this is not worth the saving.

But let's see if people think otherwise, we can iterate on this later on.

bartowski1182:
Fair enough!

Just thinking for people who might download multiple and want to test if they see a difference, it would be good to have a CLI flag that allows you to specify which specific mmproj you want to load up

I also may be mistaken and that's *exactly* what we can already do, I haven't opened up the code for this PR yet and probably won't till I'm at a computer

packages/gguf/src/gguf.spec.ts
```diff
@@ -280,12 +282,47 @@ describe("gguf", () => {
...
+		// text = Q4_K_M
+		nearestQuant = findNearestQuantType(GGMLFileQuantizationType.Q4_K_M, visionQuants);
+		expect(nearestQuant).toEqual(GGMLFileQuantizationType.Q8_0);
+		// text = Q8_0
+		nearestQuant = findNearestQuantType(GGMLFileQuantizationType.Q8_0, visionQuants);
+		expect(nearestQuant).toEqual(GGMLFileQuantizationType.Q8_0);
+		// text = F16
+		nearestQuant = findNearestQuantType(GGMLFileQuantizationType.F16, visionQuants);
+		expect(nearestQuant).toEqual(GGMLFileQuantizationType.F16);
+	});
```
ngxson, line 327: Btw @bartowski1182 , this test case is inspired from a real world scenario where we have vision quantized to F16/BF16/Q8_0, and the text can be anything else.

Feel free to suggest other test cases if you can think of any!

packages/tasks/src/gguf.ts
```diff
@@ -1,3 +1,155 @@
...
+
+export function parseGGUFQuantLabel(fname: string): string | undefined {
+	const quantLabel = fname.toUpperCase().match(GGUF_QUANT_RE_GLOBAL)?.at(-1); // if there is multiple quant substrings in a name, we prefer the last one
+	return quantLabel;
+}
+
+// order of quantization, from biggest to smallest
+// this list must be in sync with the order in GGMLFileQuantizationType
+// the gguf.spec.ts tests are using verify if the order is correct
+export const GGUF_QUANT_ORDER: GGMLFileQuantizationType[] = [
```
gary149, line 57: btw interested in improving the quant selector here: https://huggingface.co/unsloth/Qwen3-30B-A3B-GGUF?local-app=llama.cpp

ngxson: Yeah that's a good idea. This list is already exported and ready to be used in hub UI, do you think of any other improvements ?

gary149: No I think we can start with it (maybe there's a few variants missing)

ngxson: I already sync'ed this list with latest llama.cpp code, so it should be good

ngxson:
I'm merging this PR later today (unless someone oppose)

PR #1422, opened 2025-05-02
SSE and Streaming Support POC

evalstate:
Update to enable:
 - Connection via SSE (e.g. Gradio endpoints) 
 - Connection via HTTP Streaming
 - Maintains compatibility with original STDIO config format
 - Concatenates TextContent/TextResourceContents from Results; describes ImageContent/BlobResourceContents
 - SSE Server definition contains workaround for https://github.com/modelcontextprotocol/typescript-sdk/issues/436 to pass HF_TOKEN
 - Ability to connect to "Streaming HTTP" endpoints from the command line. use `pnpm agent --url https://<host>/mcp`. You can specify multiple endpoints with `--url <url1> --url <url2>`

image


julien-c:
ah btw i've "approved" for the CI to get run

packages/mcp-client/src/McpClient.ts
```diff
@@ -23,7 +29,7 @@ export interface ChatCompletionInputMessageTool extends ChatCompletionInputMessa
 
 export class McpClient {
 	protected client: InferenceClient;
-	protected provider: InferenceProviderOrPolicy | undefined;
+	protected provider: InferenceProvider | undefined;
```
julien-c, line 32: you'll need to merge main as this was changed (very slightly) on main

packages/mcp-client/package.json
```diff
@@ -55,6 +55,7 @@
 	"dependencies": {
 		"@huggingface/inference": "workspace:^",
 		"@huggingface/tasks": "workspace:^",
-		"@modelcontextprotocol/sdk": "^1.9.0"
+		"@modelcontextprotocol/sdk": ">=1.11.2",
```
julien-c, line 58:
```suggestion
		"@modelcontextprotocol/sdk": "^1.11.2",
```
better to "pin" an exact dependency imo

packages/mcp-client/src/Agent.ts
```diff
@@ -44,7 +45,7 @@ const askQuestionTool: ChatCompletionInputTool = {
 const exitLoopTools = [taskCompletionTool, askQuestionTool];
 
 export class Agent extends McpClient {
-	private readonly servers: StdioServerParameters[];
+	private readonly servers: (ServerConfig|StdioServerParameters)[];
```
julien-c, line 48: you'll want to run the linter

julien-c: if you're using VS Code, it should pick up the formatting config automatically, otherwise:

pnpm lint and pnpm format

packages/mcp-client/src/McpClient.ts
julien-c: (meaning `InferenceProviderOrPolicy` is the new type – though i agree it's a mouthful) ie. you can just revert those changes here and below

julien-c: basically what this does is we now accept some "virtual" providers like `"auto"` (which points to your preferred provider in your HF settings that serve the model in question)

cc @Wauplin @hanouticelina from our client SDK team

packages/mcp-client/cli.ts
```diff
@@ -4,28 +4,48 @@ import { stdin, stdout } from "node:process";
...
+		.filter((url) => url.length > 0);
+
+	for (const url of urls) {
+		try {
+			SERVERS.push(urlToServerConfig(url));
+		} catch (error) {
+			console.error(`Error adding server from URL "${url}": ${error.message}`);
+		}
+	}
+}
```
julien-c, line 47: i think you can just use this: 

https://nodejs.org/api/util.html#utilparseargsconfig

julien-c: and i would still keep the ones above as default (the stdio ones) if nothing is passed

packages/mcp-client/src/types.ts
```diff
@@ -0,0 +1,30 @@
...
+	url: string | URL;
+	options?: StreamableHTTPClientTransportOptions;
+}
+
+/**
+ * Discriminated union type for different MCP server types
+ */
+export type ServerConfig =
+	| { type: "stdio"; config: StdioServerParameters }
+	| { type: "sse"; config: SSEServerConfig }
```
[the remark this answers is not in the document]
julien-c: discussed offline, let's be backward compatible with SSE for now given it's not much work

packages/mcp-client/src/utils.ts
```diff
@@ -13,3 +14,77 @@ export const ANSI = {
...
+
+	const url = new URL(urlStr);
+	const hostname = url.hostname;
+	const path = url.pathname;
+
+	let type: "streamableHttp" | "sse";
+	if (path.endsWith("/sse")) {
+		type = "sse";
+	} else if (path.endsWith("/mcp")) {
+		type = "streamableHttp";
```
julien-c, line 31: wow this is actually enforced by the spec? quite a big convention push

packages/mcp-client/src/utils.ts
```diff
@@ -13,3 +14,77 @@ export const ANSI = {
...
+	} else {
+		throw new Error(`Unsupported endpoint: ${urlStr}. URL must end with /sse or /mcp`);
+	}
+
+	// Check if we should include the token
+	const authToken = token || process.env.HF_TOKEN;
```
julien-c, line 37:
```suggestion
	const authToken = token ?? process.env.HF_TOKEN;
```

tiny nit

packages/mcp-client/src/utils.ts
```diff
@@ -13,3 +14,77 @@ export const ANSI = {
...
+					? {
+							requestInit: {
+								headers: {
+									Authorization: `Bearer ${authToken}`,
+								},
+							},
+							// workaround for https://github.com/modelcontextprotocol/typescript-sdk/issues/436
+							eventSourceInit: {
+								fetch: (url, init) => {
+									const reqHeaders = new Headers(init?.headers || {});
```
julien-c, line 77:
```suggestion
									const reqHeaders = new Headers(init?.headers ?? {});
```

packages/mcp-client/src/types.ts
```diff
@@ -0,0 +1,30 @@
...
+	| { type: "streamableHttp"; config: StreamableHTTPServerConfig };
```
julien-c, line 30:
```suggestion
	| { type: "http"; config: StreamableHTTPServerConfig };
```

i think i would just rename this to `"http"`

packages/mcp-client/cli.ts
```diff
@@ -27,17 +32,28 @@ const SERVERS: StdioServerParameters[] = [
...
-if (process.env.EXPERIMENTAL_HF_MCP_SERVER) {
-	SERVERS.push({
-		// Early version of a HF-MCP server
-		// you can download it from gist.github.com/julien-c/0500ba922e1b38f2dc30447fb81f7dc6
-		// and replace the local path below
-		command: "node",
-		args: ["--disable-warning=ExperimentalWarning", join(homedir(), "Desktop/hf-mcp/index.ts")],
-		env: {
-			HF_TOKEN: process.env.HF_TOKEN ?? "",
+// Handle --url parameters: each URL will be parsed into a ServerConfig object
```
julien-c, line 35:
```suggestion
// Handle --url parameters from command line: each URL will be parsed into a ServerConfig object
```

packages/mcp-client/pnpm-lock.yaml
julien-c, line 1: (need to push the resolved .lock too)

packages/mcp-client/src/McpClient.ts
```diff
@@ -47,20 +53,42 @@ export class McpClient {
...
+	async addMcpServer(server: ServerConfig | StdioServerParameters): Promise<void> {
+		let transport: Transport;
+		const asUrl = (url: string | URL): URL => {
+			return typeof url === "string" ? new URL(url) : url;
+		};
+
+		if (!("type" in server)) {
+			transport = new StdioClientTransport({
+				...server,
+				env: { ...server.env, PATH: process.env.PATH ?? "" },
```
julien-c, line 74: ah noo it's here that i needed to pass the env.PATH explicitly! maybe it's fixed?

packages/mcp-client/src/McpClient.ts
```diff
@@ -170,7 +198,7 @@ export class McpClient {
 			const client = this.clients.get(toolName);
 			if (client) {
 				const result = await client.callTool({ name: toolName, arguments: toolArgs, signal: opts.abortSignal });
-				toolMessage.content = (result.content as Array<{ text: string }>)[0].text;
+				toolMessage.content = ResultFormatter.format(result as CallToolResult);
```
julien-c, line 201: (nit) i take this as a "bug" in the TS MCP SDK that we need to cast this as a CallToolResult, no?

julien-c (approved):
Ok, so apart from `streamableHttp` => `http` and a handful of small nits, this looks good to me

Let's 🚢 it soon!

julien-c (approved):
marked as ready for review, give it a final look @evalstate and then we can merge it?

Issue #1430, opened 2025-05-06
Issue with image segmentation on HF inference provider


oOraph:
Example with inference on this model:
https://huggingface.co/jonathandinu/face-parsing

Error obtained:
```API Implementation Error: TypeError: Invalid output: output must be of type Array<{label:string; score:number; mask: string}>```

Issue comes from here:
https://github.com/huggingface/huggingface.js/blob/a27b5b4a0c97ceb33bc476c5ad6463da136d78c1/packages/inference/src/providers/hf-inference.ts#L306

More precisely 
```
&& typeof x.score === "number"
```

because field can be a number, null or unset

(it's even specified as such here if not mistaken:
https://github.com/huggingface/huggingface.js/blob/a27b5b4a0c97ceb33bc476c5ad6463da136d78c1/packages/tasks/src/tasks/image-segmentation/inference.ts#L65)

PR #1431, opened 2025-05-06
fix(hf-inference): image segmentation

oOraph:
related: https://github.com/huggingface/huggingface.js/issues/1430

### DIFF
--- a/packages/inference/src/providers/hf-inference.ts
+++ b/packages/inference/src/providers/hf-inference.ts
@@ -303,7 +303,12 @@ export class HFInferenceImageSegmentationTask extends HFInferenceTask implements
 	override async getResponse(response: ImageSegmentationOutput): Promise<ImageSegmentationOutput> {
 		if (
 			Array.isArray(response) &&
-			response.every((x) => typeof x.label === "string" && typeof x.mask === "string" && typeof x.score === "number")
+			response.every(
+				(x) =>
+					typeof x.label === "string" &&
+					typeof x.mask === "string" &&
+					(x.score === undefined || typeof x.score === "number")
+			)
 		) {
 			return response;
 		}


packages/inference/src/providers/hf-inference.ts
```diff
@@ -303,7 +303,10 @@ export class HFInferenceImageSegmentationTask extends HFInferenceTask implements
 	override async getResponse(response: ImageSegmentationOutput): Promise<ImageSegmentationOutput> {
 		if (
 			Array.isArray(response) &&
-			response.every((x) => typeof x.label === "string" && typeof x.mask === "string" && typeof x.score === "number")
+			response.every(
+				(x) =>
+					typeof x.label === "string" && typeof x.mask === "string" && (!("score" in x) || typeof x.score === "number")
```
julien-c, line 308:
```suggestion
					typeof x.label === "string" && typeof x.mask === "string" && (x.score === undefined || typeof x.score === "number")
```

i think you can just do this (more TS-idiomatic)

pcuenca: Can we just ignore the type checks on `score`?

oOraph: ah my bad I changed it and erased the change by mistake (push force)
-> done: removed the whole type check on score, as we prefer

packages/inference/src/providers/hf-inference.ts
```diff
@@ -301,10 +301,7 @@ export class HFInferenceImageClassificationTask extends HFInferenceTask implemen
 
 export class HFInferenceImageSegmentationTask extends HFInferenceTask implements ImageSegmentationTaskHelper {
 	override async getResponse(response: ImageSegmentationOutput): Promise<ImageSegmentationOutput> {
-		if (
-			Array.isArray(response) &&
-			response.every((x) => typeof x.label === "string" && typeof x.mask === "string" && typeof x.score === "number")
-		) {
+		if (Array.isArray(response) && response.every((x) => typeof x.label === "string" && typeof x.mask === "string")) {
```
julien-c, line 304:
```suggestion
		if (Array.isArray(response) && response.every((x) => typeof x.label === "string" && typeof x.mask === "string" && (x.score === undefined || typeof x.score === "number"))) {
```

packages/inference/src/providers/hf-inference.ts
julien-c: here the type is `score?: number;` it's a valid type, we don't want random objects or string, just an optional number

julien-c (approved):

PR #1432, opened 2025-05-06
checkbox to bypass dep checks during publish checks

coyotte508:
Eg sometimes it's not important to have up-to-date `@huggingface/tasks` requirement


coyotte508:
going to merge it and test if it displays properly in the actions, otherwise revert on main

coyotte508:
it dispalys properly 
image

PR #1433, opened 2025-05-06
upgrade to pnpm 10

coyotte508:
That way I can run `pnpm link`: https://pnpm.io/cli/link to test https://github.com/huggingface/huggingface.js/pull/1412


julien-c:
> if you installed with `corepack` like recommended

i don't think i did

PR #1434, opened 2025-05-06
update snippets for image-text-to-text models for transformers models

Vaibhavs10:


packages/tasks/src/model-libraries-snippets.ts
```diff
@@ -1118,12 +1118,32 @@ export const transformers = (model: ModelData): string[] => {
...
 		const pipelineSnippet = ["# Use a pipeline as a high-level helper", "from transformers import pipeline", ""];
+		let useMessagesSyntax = false;
 
-		if (model.tags.includes("conversational") && model.config?.tokenizer_config?.chat_template) {
-			pipelineSnippet.push("messages = [", '    {"role": "user", "content": "Who are you?"},', "]");
+		if (model.config?.tokenizer_config?.chat_template) {
+			if (model.tags.includes("image-text-to-text") && model.tags.includes("conversational")) {
+				pipelineSnippet.push(
+					"messages = [",
+					[
```
pcuenca, line 1127: I think this line became spurious?

packages/tasks/src/model-libraries-snippets.ts
```diff
@@ -1118,12 +1118,32 @@ export const transformers = (model: ModelData): string[] => {
...
+						'        "content": [',
+						'            {"type": "image", "url": "https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/p-blog/candy.JPG"},',
+						'            {"type": "text", "text": "What animal is on the candy?"}',
+						"        ]",
+						"    },",
+					].join("\n"),
+					"]"
+				);
+				useMessagesSyntax = true;
+			} else if (model.tags.includes("conversational")) {
```
pcuenca, line 1139: Don't we need `messages` here as well?

pcuenca: Perhaps we can do the flow like `if conversational and chat_template`, then check whether image-text-to-text or not, and remove the `useMessagesSyntax` var.

packages/tasks/src/model-libraries-snippets.ts
pcuenca: ah there's a join later 👍

packages/tasks/src/model-libraries-snippets.ts
Vaibhavs10: good point

packages/tasks/src/model-libraries-snippets.ts
```diff
@@ -1119,13 +1119,27 @@ export const transformers = (model: ModelData): string[] => {
 	if (model.pipeline_tag && LIBRARY_TASK_MAPPING.transformers?.includes(model.pipeline_tag)) {
 		const pipelineSnippet = ["# Use a pipeline as a high-level helper", "from transformers import pipeline", ""];
 
-		if (model.tags.includes("conversational") && model.config?.tokenizer_config?.chat_template) {
-			pipelineSnippet.push("messages = [", '    {"role": "user", "content": "Who are you?"},', "]");
+		if (model.config?.tokenizer_config?.chat_template) {
+			if (model.tags.includes("conversational")) {
```
pcuenca, line 1123: But image-text-to-text models are conversational too, no?

Vaibhavs10: yeah! I basically am a bit dense: https://github.com/huggingface/huggingface.js/pull/1434/commits/72e521231bf3a393adb25d3283dc00fd53469e89

packages/tasks/src/model-libraries-snippets.ts
```diff
@@ -1119,13 +1119,27 @@ export const transformers = (model: ModelData): string[] => {
...
+				);
+			} else if (model.tags.includes("conversational")) {
+				pipelineSnippet.push("messages = [", '    {"role": "user", "content": "Who are you?"},', "]");
+			}
 		}
 		pipelineSnippet.push(`pipe = pipeline("${model.pipeline_tag}", model="${model.id}"` + remote_code_snippet + ")");
-		if (model.tags.includes("conversational") && model.config?.tokenizer_config?.chat_template) {
-			pipelineSnippet.push("pipe(messages)");
-		}
+		pipelineSnippet.push("pipe(messages)");
```
pcuenca, line 1142: hmm, I think this line is still out of the conversational path. How about something like this? (forgive mistakes, typing in this text box)

```
		pipelineSnippet.push(`pipe = pipeline("${model.pipeline_tag}", model="${model.id}"` + remote_code_snippet + ")");

		if (model.config?.tokenizer_config?.chat_template && model.tags.includes("conversational")) {
			if (model.tags.includes("image-text-to-text")) {
				pipelineSnippet.push(
					"messages = [",
					[
						"    {",
						'        "role": "user",',
						'        "content": [',
						'            {"type": "image", "url": "https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/p-blog/candy.JPG"},',
						'            {"type": "text", "text": "What animal is on the candy?"}',
						"        ]",
						"    },",
					].join("\n"),
					"]"
				);
			} else {
				pipelineSnippet.push("messages = [", '    {"role": "user", "content": "Who are you?"},', "]");
			}
			pipelineSnippet.push("pipe(messages)");
		}

		return [pipelineSnippet.join("\n"), autoSnippet];
```

Vaibhavs10: It isn't no? https://gist.github.com/Vaibhavs10/b525c1ae4982c04fe0202c2f78ac2f03

the pipeline instantiation and the pipe(messages) is the same for both

(my JS sucks so could very well be a skill issue)

pcuenca: lol, I'm probably blind by now, but this is the way I'm seeing it:

Screenshot 2025-05-07 at 15 40 37

i.e., if there is no chat_template, there's no `messages`.

Also, if `conversational` is not present (a mistake, probably) we wouldn't have messages either, that's why I suggested to require it in the outer if too.

Wauplin: I suggested another variation based on your previous comments here: https://gist.github.com/Vaibhavs10/b525c1ae4982c04fe0202c2f78ac2f03?permalink_comment_id=5569573#gistcomment-5569573

Changes are:
- the `pipe = pipeline(...` line should be added at the beginning  => no matter the pipeline_tag we want to instantiate the pipeline
- the `pipe(messages)` line should be added only if conversational mode as @pcuenca mentioned [above](https://github.com/huggingface/huggingface.js/pull/1434#discussion_r2077681595) 
- I would not check `model.config?.tokenizer_config?.chat_template` manually but rely only on the `model.tags.includes("conversational")` condition. This makes the logic simpler and it should be equivalent since moon-landing is automatically adding `conversational` when chat template is detected (see [private link](https://github.com/huggingface-internal/moon-landing/blob/2d1dd4129a2e76d4ad4d26ba7587fa538d46dd02/server/repo_types/ModelInfo.ts#L684)) (it's even better because moon-landing checks for both tokenizer_config and processor_config and is maintained to be future-proof)

julien-c: yes^

Vaibhavs10: I'm stupid for not looking at that code block - sorry for being so dense!

pcuenca (approved):

PR #1435, opened 2025-05-06
Enable model download count for hear-pytorch

sebbaur:
https://huggingface.co/google/hear-pytorch/tree/main


sebbaur:
TODO: once submitted, update lib in model card

sebbaur:
I had misunderstood how this worked, this is not needed

PR #1437, opened 2025-05-07
Fix lint from #1421

pcuenca:
Linter tests passed in #1421, but then they fail with `Forbidden non-null assertion` on `main` after merging new PRs.


packages/tasks/src/gguf.ts
```diff
@@ -137,7 +137,7 @@ export function findNearestQuantType(
 	// This means finding the largest quantization that is smaller than or equal to the target.
 	for (const availableQuant of sortedAvailable) {
 		// We know the key exists due to the filter above.
-		const availableIndex = orderMap.get(availableQuant)!;
+		const availableIndex = orderMap.get(availableQuant) ?? 0;
```
pcuenca, line 140: I suppose it doesn't matter what number we put here cc @ngxson

coyotte508 (approved):

PR #1443, opened 2025-05-09
Local apps: add MLX LM

gary149:
Still hesitating between this (adding it as a local app) and updating the libraries snippets directly wdyt @pcuenca ? (maybe let's add it as a local app and change the library snippets to python?)

### DIFF
--- a/packages/tasks/src/local-apps.ts
+++ b/packages/tasks/src/local-apps.ts
@@ -282,7 +282,7 @@ const snippetMlxLm = (model: ModelData): LocalAppSnippet[] => {
 			content: [
 				...(model.tags.includes("conversational")
 					? ["# Interactive chat REPL", `mlx_lm.chat --model "${model.id}"`]
-					: ["# One-shot generation", `mlx_lm.generate --model "${model.id}" --prompt "Hello"`]),
+					: ["# Generate some text", `mlx_lm.generate --model "${model.id}" --prompt "Once upon a time"`]),
 			].join("\n"),
 		},
 		...(model.tags.includes("conversational")
--- a/packages/tasks/src/local-apps.ts
+++ b/packages/tasks/src/local-apps.ts
@@ -262,6 +262,43 @@ const snippetTgi = (model: ModelData): LocalAppSnippet[] => {
 	];
 };
 
+const snippetMlxLm = (model: ModelData): LocalAppSnippet[] => {
+	const openaiCurl = [
+		"# Calling the OpenAI-compatible server with curl",
+		`curl -X POST "http://localhost:8000/v1/chat/completions" \\`,
+		`   -H "Content-Type: application/json" \\`,
+		`   --data '{`,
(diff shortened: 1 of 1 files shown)


packages/tasks/src/local-apps.ts
```diff
@@ -302,6 +339,13 @@ export const LOCAL_APPS = {
 			(model.pipeline_tag === "text-generation" || model.pipeline_tag === "image-text-to-text"),
 		snippet: snippetVllm,
 	},
+	"mlx-lm": {
+		prettyLabel: "MLX LM",
+		docsUrl: "https://github.com/ml-explore/mlx-lm",
+		mainTask: "text-generation",
+		displayOnModelPage: isMlxModel,
```
julien-c, line 346: i would only display it on actually supported models (so check the pipeline_tag)

packages/tasks/src/local-apps.ts
```diff
@@ -262,6 +262,43 @@ const snippetTgi = (model: ModelData): LocalAppSnippet[] => {
...
+		`     "messages": [`,
+		`       {"role": "user", "content": "Hello"}`,
+		`     ]`,
+		`   }'`,
+	];
+
+	return [
+		{
+			title: "Generate or start a chat session",
+			setup: ["# Install MLX LM", "pip install mlx-lm"].join("\n"),
```
julien-c, line 281: or use uvx and drop the pip install?

or even better, `uv tool install mlx-lm`

packages/tasks/src/local-apps.ts
```diff
@@ -262,6 +262,43 @@ const snippetTgi = (model: ModelData): LocalAppSnippet[] => {
...
+			content: [
+				"# One-shot generation",
+				`mlx_lm.generate --model "${model.id}" --prompt "Hello"`,
```
julien-c, line 284: no need for this one in case the model is conversational, IMO

julien-c (commented):
> (maybe let's add it as a local app and change the library snippets to python?)

yes, i think that's what we should do:
- (modeling) libraries are for actual code snippets, including in Python
- local apps are for CLIs

That's consistent with what we've done for llama.cpp:
llama.cpp and node-llama-cpp are apps
whereas llama-cpp-python is a "library"

packages/tasks/src/local-apps.ts
julien-c: (note, mainTask does not check the model's task on a model page, it's only used for the settings page)

packages/tasks/src/local-apps.ts
gary149: it still work and can be useful but I removed it (it's better for the copy button to keep one snippet only)

pcuenca (approved):

pcuenca:
cc @Vaibhavs10

julien-c (approved):
nice

PR #1449, opened 2025-05-13
Fix UrlConversion Test - remove sensitivity to environment variables

evalstate:
Update to improve Environment Variable handling and ensure test is not sensitive to presence of token.


packages/mcp-client/src/utils.ts
```diff
@@ -34,7 +34,7 @@ export function urlToServerConfig(urlStr: string, token?: string): ServerConfig
 	}
 
 	// Check if we should include the token
-	const authToken = token ?? process.env.HF_TOKEN;
+	const authToken = token;
```
julien-c, line 37: or just rename the param to authToken

PR #1452, opened 2025-05-14
gguf: add custom Q_K_XL quants

ngxson:
As reported by @unslothai , these quant types are custom (not officially exist in llama.cpp). Adding these to the list so it will be recognized via registry API

After this PR is merged, we will also push the change to main code base (aka moon-landing)


danielhanchen:
Fantastic and thanks! :)

PR #1456, opened 2025-05-16
Track download stats for the ACE-Step repo.

DumoeDss:


packages/tasks/src/model-libraries.ts
```diff
@@ -1040,6 +1040,14 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		countDownloads: `path:"model_vae_fp16.pt"`,
 		snippets: snippets.threedtopia_xl,
 	},
+	acestep: {
```
Vaibhavs10, line 1043: this keyword should be present in the model metadata too here: https://huggingface.co/ACE-Step/ACE-Step-v1-3.5B

you can do so by adding this tag in the model README.

packages/tasks/src/model-libraries.ts
```diff
@@ -1040,6 +1040,14 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
...
+		prettyLabel: "ACE-Step",
+		repoName: "ACE-Step",
+		repoUrl: "https://github.com/ace-step/ACE-Step",
+		filter: false,
+		countDownloads: `path:"ace_step_transformer/diffusion_pytorch_model.safetensors"`,
```
Vaibhavs10, line 1048: perhaps easier would be that we track it via the `config.json`: https://huggingface.co/ACE-Step/ACE-Step-v1-3.5B/blob/main/config.json

Vaibhavs10 (commented):
Thanks for the PR @DumoeDss - few nits and we should be golden, I'm not sure why the library is set to `diffusers` in the repo: https://huggingface.co/ACE-Step/ACE-Step-v1-3.5B when it actually uses the acestep model library: https://github.com/ace-step/ACE-Step/blob/74121a5d8a9af2fd651ba70869e78868442f6329/infer.py#L4

let's fix that too please! 🙏

packages/tasks/src/model-libraries.ts
DumoeDss: I have added the 'acestep' tag to the model card metadata.

PR #1458, opened 2025-05-16
[Inference]  support all tasks for the auto policy

krampstudio:
For the  `auto` policy, `hf-inference` doesn't support `text-to-video` so we should bypass the mapping

see the missing snippet in https://huggingface.co/Wan-AI/Wan2.1-T2V-1.3B?language=python&inference_api=true&inference_provider=auto


SBrandeis (approved):
lgtm!
Defering to @hanouticelina for final approval

hanouticelina (approved):

PR #1459, opened 2025-05-16
[Inference Providers] feat. Refactor the NovitaTextToVideoTask using the async API.

viktor2077:
This PR refactors the `NovitaTextToVideoTask` using the async API, aiming to allow inference with models that may take over 2 minutes to generate results.

### DIFF
--- a/packages/inference/src/providers/novita.ts
+++ b/packages/inference/src/providers/novita.ts
@@ -17,6 +17,7 @@
 import { InferenceOutputError } from "../lib/InferenceOutputError";
 import { isUrl } from "../lib/isUrl";
 import type { BodyParams, UrlParams } from "../types";
+import { delay } from "../utils/delay";
 import { omit } from "../utils/omit";
 import {
 	BaseConversationalTask,
@@ -26,11 +27,11 @@ import {
 } from "./providerHelper";
 
 const NOVITA_API_BASE_URL = "https://api.novita.ai";
-export interface NovitaOutput {
-	video: {
-		video_url: string;
-	};
+
+export interface NovitaAsyncAPIOutput {
+	task_id: string;
 }
+
 export class NovitaTextGenerationTask extends BaseTextGenerationTask {
 	constructor() {
 		super("novita", NOVITA_API_BASE_URL);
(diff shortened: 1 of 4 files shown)


packages/inference/src/providers/novita.ts
```diff
@@ -50,38 +51,88 @@ export class NovitaConversationalTask extends BaseConversationalTask {
...
 		super("novita", NOVITA_API_BASE_URL);
 	}
 
-	makeRoute(params: UrlParams): string {
-		return `/v3/hf/${params.model}`;
+	override makeRoute(params: UrlParams): string {
+		if (params.authMethod !== "provider-key") {
+			return `/v3/async/${params.model}?_subdomain=queue`;
+		}
+		return `/v3/async/${params.model}`;
```
SBrandeis, line 64: Novita API doesn't have a `queue` subdomain from what I can tell from here: https://novita.ai/docs/guides/model-apis-dedicated-endpoints
```suggestion
		return `/v3/async/${params.model}`;
```

packages/inference/src/providers/novita.ts
```diff
@@ -50,38 +51,88 @@ export class NovitaConversationalTask extends BaseConversationalTask {
...
+	override preparePayload(params: BodyParams): Record<string, unknown> {
+		const { num_inference_steps, ...restParameters } = params.args.parameters as Record<string, unknown>;
 		return {
 			...omit(params.args, ["inputs", "parameters"]),
-			...(params.args.parameters as Record<string, unknown>),
+			...restParameters,
+			steps: num_inference_steps,
 			prompt: params.args.inputs,
 		};
 	}
```
SBrandeis, line 75: No need for casting if you use the proper type here
Also, `params.args.parameters` can be undefined

```suggestion
	override preparePayload(params: BodyParams<TextToVideoArgs>): Record<string, unknown> {
		const { num_inference_steps, ...restParameters } = params.args.parameters ?? {};
		return {
			...omit(params.args, ["inputs", "parameters"]),
			...restParameters,
			steps: num_inference_steps,
			prompt: params.args.inputs,
		};
	}
```

packages/inference/src/providers/novita.ts
```diff
@@ -50,38 +51,88 @@ export class NovitaConversationalTask extends BaseConversationalTask {
...
+			throw new InferenceOutputError("No task ID found in the response");
+		}
+
+		const parsedUrl = new URL(url);
+		const baseUrl = `${parsedUrl.protocol}//${parsedUrl.host}${
+			parsedUrl.host === "router.huggingface.co" ? "/novita" : ""
+		}`;
+		const queryParams = parsedUrl.search;
+		const resultUrl = `${baseUrl}/v3/async/task-result${queryParams ? queryParams + '&' : '?'}task_id=${taskId}`;
+
```
SBrandeis, line 96: Are the other query params important here?
If not, you can just pass the `task_id` in the URL:


```suggestion
		const resultUrl = `${baseUrl}/v3/async/task-result?task_id=${taskId}`;

```

SBrandeis: If other search parameters do matter, please use [the `parsedUrl.searchParams` API instead](https://developer.mozilla.org/en-US/docs/Web/API/URL/searchParams)

packages/inference/src/providers/novita.ts
```diff
@@ -50,38 +51,88 @@ export class NovitaConversationalTask extends BaseConversationalTask {
...
+		let status = '';
+		let taskResult = undefined;
```
SBrandeis, line 98: Should be typed as `unknown`
```suggestion
		let taskResult: unknown;
```

SBrandeis (commented):
Hey there - thank you for your contribution!

I have left a few comments and suggestions about typing and code style

Excited to see Novita's text-to-video service on the platform!

packages/inference/src/providers/novita.ts
viktor2077: Hi @SBrandeis, thanks for this suggestion. And I will encounter the following error if I don't cast the type explicitly:

```
Property 'num_inference_steps' does not exist on type '{}'.ts(2339)
```

viktor2077: and I'm refactoring the code to handle cases where params.args.parameters is null or undefined by using an empty object {} as the default value. This ensures the destructuring operation won't fail even if the parameters property doesn't exist:

```typescript 
const { num_inference_steps, ...restParameters } = (params.args.parameters as Record<string, unknown>) ?? {};
```

viktor2077:
> I have left a few comments and suggestions about typing and code style

Thanks @SBrandeis , these suggestions are detailed and helpful!

SBrandeis (commented):
Nice!
Apart from that typing issue, it looks good to me.
Don't forget to run `pnpm format` and `pnpm lint` in the `packages/inference` folder to conform our code quality checks.

SBrandeis (approved):
Looks good to me!
Defering to @hanouticelina for a second pair of eyes 👀 
Thanks @viktor2077

hanouticelina (approved):
Thanks @viktor2077 for the PR, and sorry again for the delay! 🤗 
I’ve tested it with direct calls to Novita and everything works as expected.
I also opened a server-side PR to allow `GET` requests through Hugging Face routing — I’ll merge this as soon as that’s deployed.

PR #1468, opened 2025-05-20
Add molecules as task

merveenoyan:
cc @NielsRogge @julien-c inspired by facebook OMol release I did a full text search "molecules" and saw 853 results in /models. I think it makes sense to add it as a task under "others" 
Also opened another PR to add the icon

### DIFF
--- a/packages/tasks/src/pipelines.ts
+++ b/packages/tasks/src/pipelines.ts
@@ -693,6 +693,12 @@ export const PIPELINE_DATA = {
 		color: "yellow",
 		hideInDatasets: true,
 	},
+	"molecules": {
+		name: "Molecules",
+		modality: "other",
+		color: "green",
+		hideInDatasets: true,
+	},
 	other: {
 		name: "Other",
 		modality: "other",
--- a/packages/tasks/src/tasks/index.ts
+++ b/packages/tasks/src/tasks/index.ts
@@ -221,6 +221,7 @@ export const TASKS_DATA: Record<PipelineType, TaskData | undefined> = {
 	"keypoint-detection": getData("keypoint-detection", keypointDetection),
 	"mask-generation": getData("mask-generation", maskGeneration),
 	"multiple-choice": undefined,
+	"molecules": undefined,
 	"object-detection": getData("object-detection", objectDetection),
 	"video-classification": getData("video-classification", videoClassification),
 	other: undefined,


packages/tasks/src/pipelines.ts
```diff
@@ -693,6 +693,12 @@ export const PIPELINE_DATA = {
 		color: "yellow",
 		hideInDatasets: true,
 	},
+	"molecules": {
+		name: "Molecules",
+		modality: "other",
+		color: "green",
+		hideInDatasets: true,
+	},
```
julien-c, line 701: maybe it's more for domains where we have biology etc?

julien-c: i.e. in there (internal codebase):

```ts
export const OTHER_TAGS_SUGGESTIONS = [
	"chemistry",
	"biology",
	"finance",
	"legal",
	"music",
	"art",
	"code",
	"climate",
	"medical",
	TAG_NFAA_CONTENT,
] as const;
```

Vaibhavs10: +1 and doesn't Chemistry cover molecules? - a bit wary of adding too niche tags

NielsRogge: Where are those domains meant to be displayed? I would prefer just a "Biology" tag next to the current "Graph machine learning" one to cover that domain (including protein folding, DNA sequence models, molecules and more).

merveenoyan: I think it's a different output type with enough number of models (although represented by text), a bit like 3D in a way.

merveenoyan: this model for instance outputs structured output

Vaibhavs10 (commented):
Thought about it a bit more pretty much all the models that you see tagged as `molecules` are also tagged w/ `Chemistry`: https://huggingface.co/search/full-text?q=molecules&type=model

I'm not sure if adding a niche task which is not tied to a specific modality like this makes much sense.

merveenoyan:
@Vaibhavs10 I feel like we should make larger domain specific filters visible on Hub, for which I have opened an issue long ago and was closed by @julien-c I think it's hard for users to write tags to link everytime they want to search for something specific. we still don't resurface commonly used tags https://github.com/huggingface-internal/moon-landing/issues/4381 

Other tab in /models is a good candidate for this imo

merveenoyan:
I'm curious how these models should still be tagged, OMol and UMA for instance still sit untagged on Hub

PR #1470, opened 2025-05-20
Tiny agents custom endpoint

julien-c:


packages/tiny-agents/README.md
```diff
@@ -42,7 +41,7 @@ touch my-agent/agent.json
 ```json
 {
 	"model": "Qwen/Qwen2.5-72B-Instruct", // model id
-	"provider": "nebius", // or you can also use a local endpoint base url with `endpointUrl`
+	"provider": "nebius", // or you can also use a local endpoint base url, see next example
```
gary149, line 44: I would remove any comments from json code blocks

packages/tiny-agents/src/cli.ts
```diff
@@ -129,13 +134,24 @@ async function main() {
...
-		provider: config.provider,
-		model: config.model,
-		apiKey: process.env.HF_TOKEN ?? "",
-		servers: config.servers,
-		prompt,
-	});
+	const agent = new Agent(
+		config.endpointUrl
+			? {
+					endpointUrl: config.endpointUrl,
```
gary149, line 140: ok! (I expected to pass something like this when using custom endpoint btw:
``` base_url="",
  api_key=""```

gary149 (approved):

PR #1471, opened 2025-05-21
Add support to `image-to-video` task for `inference`

apolinario:
[Internal discussion](https://huggingface.slack.com/archives/C089DMDB415/p1747733633119449) 

The goal is:
- Keep inputs similar to image-to-image
- Keep outputs similar to text-to-video
- For now super generic `about.md` 
- `data.ts` for the `/tasks` on huggingface.co is pending [merging this PR](https://huggingface.co/datasets/huggingfacejs/tasks/discussions/7)


packages/inference/src/snippets/getInferenceSnippets.ts
```diff
@@ -266,6 +268,11 @@ const prepareImageToImageInput = (model: ModelDataMinimal): object => {
 	return { inputs: data.image, parameters: { prompt: data.prompt } };
 };
 
+const prepareImageToVideoInput = (model: ModelDataMinimal): object => {
+	const data = JSON.parse(getModelInputSnippet(model) as string);
```
Wauplin, line 272: Could you add a simple example in https://github.com/huggingface/huggingface.js/blob/main/packages/tasks/src/snippets/inputs.ts ? Something like this:

```ts
const inputsImageToVideo = () => `{
    "image": "cat.png",
    "prompt": "Make the cat dance."
}`;
```

and then 

```ts
    "image-to-video": inputsImageToVideo,
```
in the `modelInputSnippets` mapping.

This way, `const data = JSON.parse(getModelInputSnippet(model) as string);` will return a valid payload to generate the inference snippets.

packages/inference/src/tasks/cv/imageToVideo.ts
```diff
@@ -0,0 +1,23 @@
+import type { ImageToVideoInput } from "@huggingface/tasks";
```
Wauplin, line 1: Let's not add this file yet and wait for a follow-up PR. The current implementation assumes it works with FalAI and Replicate which is not the case yet -and not tested-. Better to separate them and keep this PR for specs definition only.

Wauplin (commented):
Thanks @apolinario ! Pretty solid PR already :) I've made some high-level comments while you are still working on it. I do think it's best to split the PR into 2 (one for spec definition, one for proper implementation for each provider).

apolinario:
Thanks @Wauplin ! I've kept it strictly spec-based for now and applied your suggestion for [snippets/inputs.ts](https://github.com/huggingface/huggingface.js/blob/main/packages/tasks/src/snippets/inputs.ts)

Wauplin (approved):
Looks good to me, thanks!

Let's wait for another review, especially for the about.md/data.ts part which I know less about

packages/tasks/src/tasks/image-to-video/data.ts
```diff
@@ -0,0 +1,172 @@
...
+			description: "A 14B parameter model for 720p image-to-video generation by Wan-AI.",
+			id: "Wan-AI/Wan2.1-I2V-14B-720P",
+		},
+		{
+			description: "A Diffusers version of the Wan2.1-I2V-14B-720P model for 720p image-to-video generation.",
+			id: "Wan-AI/Wan2.1-I2V-14B-720P-Diffusers",
+		},
+		{
+			description:
+				"A 14B parameter model for frame-level feature to video (FLF2V) generation by Wan-AI, producing 720p videos (Diffusers version).",
```
merveenoyan, line 123:
```suggestion
				"An image-to-video model that generates videos from frame-level features, producing 720p videos.",
```

packages/tasks/src/tasks/image-to-video/data.ts
```diff
@@ -0,0 +1,172 @@
...
+			id: "motion_score",
+		},
+	],
+	models: [
+		{
+			description: "LTX-Video, a 13B parameter model for high quality video generation",
+			id: "Lightricks/LTX-Video-0.9.7-dev",
+		},
+		{
+			description: "A 1.3B parameter model for reference-based video generation",
```
merveenoyan, line 74: fyi we add very few models, and simple explanations on them (this one is ok but others are a bit too long

packages/tasks/src/tasks/image-to-video/data.ts
```diff
@@ -0,0 +1,172 @@
...
+			id: "stabilityai/stable-video-diffusion-img2vid",
+		},
+		{
+			description: "A 5 billion parameter model for image-to-video generation by THUDM.",
+			id: "THUDM/CogVideoX-5b-I2V",
+		},
+	],
+	spaces: [
+		{
+			description: "Generate videos fast withthe LTX-Video distilled model.",
```
merveenoyan, line 146:
```suggestion
			description: "An application to generate videos fast.",
```

merveenoyan (commented):
thanks a lot 💗

packages/tasks/src/tasks/image-to-video/data.ts
```diff
@@ -0,0 +1,172 @@
+import type { TaskDataCustom } from "../index.js";
+
+const taskData: TaskDataCustom = {
+	datasets: [
+		{
+			description: "A dataset of images and short video clips for image-to-video generation research.",
+			id: "some/image-to-video-dataset",
+		},
+		{
+			description: "A benchmark dataset for reference-based video generation.",
```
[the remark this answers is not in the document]
apolinario: It's like you feed an image of a person and a background and it will generate a video of that person in that background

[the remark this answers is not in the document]
apolinario: they call it "reference-to-video" https://github.com/ali-vilab/VACE, I tried to phrase it better though

apolinario:
Thanks merve, applied your suggestions!

SBrandeis (approved):
Approving in principle - deferring to @merveenoyan re: `data.ts`

I will review the inference client implementation more thoroughly :)

PR #1476, opened 2025-05-22
feat(snippet): update llama.cpp snippet to include windows

mfuntowicz:
This PR adds the Windows way of installing and running llama.cpp on such OSes.

llama.cpp is now available on WinGet and as easy installable as brew on Windows machines now. 

```powershelll
PS C:\Users\momo-> winget install llama.cpp
PS C:\Users\momo-> # Load and run the model:
PS C:\Users\momo-> llama-cli -hf lmstudio-community/Meta-Llama-3.1-8B-Instruct-GGUF:Q4_K_M
load_backend: loaded RPC backend from C:\Users\momo-\AppData\Local\Microsoft\WinGet\Packages\ggml.llamacpp_Microsoft.Winget.Source_8wekyb3d8bbwe\ggml-rpc.dll
ggml_vulkan: Found 1 Vulkan devices:
ggml_vulkan: 0 = NVIDIA GeForce RTX 3090 (NVIDIA) | uma: 0 | fp16: 1 | warp size: 32 | shared memory: 49152 | int dot: 1 | matrix cores: KHR_coopmat
[...]
== Running in interactive mode. ==
 - Press Ctrl+C to interject at any time.
 - Press Return to return control to the AI.
 - To return control without starting a new line, end your input with '/'.
 - If you want to submit another line, end your input with '\'.
 - Not using system message. To change it, set a different value via -sys PROMPT


> Hey how are you?
I'm just a language model, I don't have feelings or emotions like humans do, but I'm functioning properly and ready to help with any questions or tasks you may have! How about you? How's your day going?
```

### DIFF
--- a/packages/tasks/src/local-apps.ts
+++ b/packages/tasks/src/local-apps.ts
@@ -117,7 +117,7 @@ const snippetLlamacpp = (model: ModelData, filepath?: string): LocalAppSnippet[]
 	};
 	return [
 		{
-			title: "Install from brew (MacOS)",
+			title: "Install from brew",
 			setup: "brew install llama.cpp",
 			content: command("llama-cli"),
 		},
--- a/packages/tasks/src/local-apps.ts
+++ b/packages/tasks/src/local-apps.ts
@@ -117,8 +117,13 @@ const snippetLlamacpp = (model: ModelData, filepath?: string): LocalAppSnippet[]
 	};
 	return [
 		{
-			title: "Install from brew",
+			title: "Install from brew (MacOS)",
 			setup: "brew install llama.cpp",
+			content: command("llama-cli"),
+		},
+		{
+			title: "Install from WinGet (Windows)",
+			setup: "winget install llama.cpp",
 			content: command("llama-cli"),
 		},
 		{


packages/tasks/src/local-apps.ts
```diff
@@ -117,10 +117,15 @@ const snippetLlamacpp = (model: ModelData, filepath?: string): LocalAppSnippet[]
 	};
 	return [
 		{
-			title: "Install from brew",
+			title: "Install from brew (MacOS)",
```
pcuenca, line 120:
```suggestion
			title: "Install from brew",
```

I have never tried it, but homebrew is also available on Linux (and Windows, I think)

mfuntowicz: What about (MacOS, Linux) - the llama.cpp formulae is not available through brew on Windows - wdyt?

julien-c (approved):
we could even make those options dependent on `navigator.userAgent.includes("Windows")` (when it's run in frontend code)

cc @gary149 on whether we want to do it or not.

In any case this can be in a later PR

packages/tasks/src/local-apps.ts
julien-c: i think most people know brew so no need to specify the platforms imo (only for winget)

mfuntowicz:
Failing tests look unrelated - ok to merge? @julien-c @pcuenca?

pcuenca (approved):

PR #1477, opened 2025-05-22
llama-cli -> llama-server.

Vaibhavs10:


packages/tasks/src/local-apps.ts
```diff
@@ -133,17 +133,17 @@ const snippetLlamacpp = (model: ModelData, filepath?: string): LocalAppSnippet[]
...
 			].join("\n"),
-			content: command("./llama-cli"),
+			content: command("./llama-server"),
 		},
 		{
 			title: "Build from source code",
 			setup: [
 				"git clone https://github.com/ggerganov/llama.cpp.git",
 				"cd llama.cpp",
 				"cmake -B build -DLLAMA_CURL=ON",
```
Vaibhavs10, line 143: prolly also can remove the `DLLAMA_CURL=ON` as well I think

packages/tasks/src/local-apps.ts
```diff
@@ -108,23 +108,18 @@ function getQuantTag(filepath?: string): string {
 const snippetLlamacpp = (model: ModelData, filepath?: string): LocalAppSnippet[] => {
 	const command = (binary: string) => {
 		const snippet = ["# Load and run the model:", `${binary} -hf ${model.id}${getQuantTag(filepath)}`];
```
ngxson, line 110: Temporary solution until we have the `--browser` flag

```suggestion
		const snippet = ["# Load and run the model:", `${binary} -hf ${model.id}${getQuantTag(filepath)}`, "# Then, access http://127.0.0.1:8080"];
```

ngxson (approved):

packages/tasks/src/local-apps.ts
Vaibhavs10: it think it's okay, since llama.cpp does show where to go in the CLI output. I'd prefer to keep it clean

packages/tasks/src/local-apps.ts
Vaibhavs10: removed

ngxson:
Nice, thanks! Let's 🚢  this

pcuenca (approved):

PR #1483, opened 2025-05-23
Use TSHY when releasing @huggingface/inference

coyotte508:
Same thing as https://github.com/huggingface/huggingface.js/pull/1029 but for `@huggingface/inference`

See https://arethetypeswrong.github.io/?p=%40huggingface%2Finference%403.13.2 currently we have errors

image

New types are green: https://arethetypeswrong.github.io/?p=%40huggingface%2Finference%403.13.2-tshy

image


packages/inference/src/InferenceClient.ts
```diff
@@ -1,8 +1,9 @@
-import * as tasks from "./tasks";
-import type { Options } from "./types";
-import { omit } from "./utils/omit";
-import { typedEntries } from "./utils/typedEntries";
+import * as tasks from "./tasks/index.js";
+import type { Options } from "./types.js";
+import { omit } from "./utils/omit.js";
+import { typedEntries } from "./utils/typedEntries.js";
 
+/import\s+([^\s]+)\s+from\s+(['"])(\.{1,2}\/[^'"]+?)(?<!\.js)\2/g;
```
SBrandeis, line 6: Why is this needed?

SBrandeis (approved):
nice!
is that an API breaking change?

packages/inference/src/InferenceClient.ts
coyotte508: nvm it's the regex I used to replace the imports

coyotte508:
```suggestion
```

coyotte508:
> nice! is that an API breaking change?

should not be

PR #1496, opened 2025-05-27
Add `tiny-agents` docs to ToC

hanouticelina:


coyotte508:
I guess you need to edit `packages/doc-internal/package.json` to at least copy the README to the docs folder, and also `.github/workflows/pr-documentation.yml` to watch more files

julien-c (approved):
we should remove the agents one (huggingface/agents) from the ToC it's quite outdated

Other than that, approving!

PR #1500, opened 2025-05-28
Add gearhash, xetchunk and blake3 WASM packages

coyotte508:
This PR is an experiment to see what a pure-wasm [gearhash](https://github.com/srijs/rust-gearhash) looks like, using [https://www.assemblyscript.org](AssemblyScript), a language purpose made to generate WASM.

Generating WASM from rust uses a lot of glue, the glue for assembly script seems a lot thinner:

```js
async function instantiate(module, imports = {}) {
  const { exports } = await WebAssembly.instantiate(module, imports);
  return exports;
}
export const {
  memory,
  add,
} = await (async url => instantiate(
  await (async () => {
    const isNodeOrBun = typeof process != "undefined" && process.versions != null && (process.versions.node != null || process.versions.bun != null);
    if (isNodeOrBun) { return globalThis.WebAssembly.compile(await (await import("node:fs/promises")).readFile(url)); }
    else { return await globalThis.WebAssembly.compileStreaming(globalThis.fetch(url)); }
  })(), {
  }
))(new URL("debug.wasm", import.meta.url));
```

It will also be interesting to compare wasm size between pure wasm & from-rust wasm, as well as provide `gearhash` to the wider JS/NPM ecosystem.

`gearhash` is a core component of the xet protocol chunking & hashing and a cool piece of tech!

cc @Wauplin @hanouticelina @assafvayner @rajatarya for viz

Note: also added `xetChunk` and `blake3` packages


coyotte508:
Going to merge this

Even if it's not getting used atm due to perf it provides nice reference implems in assembly script

PR #1509, opened 2025-06-02
Safetensors now optionally exposes total params

julien-c:


packages/hub/src/lib/parse-safetensors-metadata.ts
```diff
@@ -223,27 +252,45 @@ export async function parseSafetensorsMetadata(
...
-			...(params.computeParametersCount && {
-				parameterCount: computeNumOfParamsByDtypeSharded(headers),
-			}),
+			headers: shardedMap,
+			...(params.computeParametersCount
+				? {
+						parameterCount: computeNumOfParamsByDtypeSharded(shardedMap),
+						parameterTotal:
+							/// shortcut: get param count directly from metadata
+							index.metadata?.total_parameters ? parseInt(index.metadata.total_parameters.toString()) : undefined,
```
Kakulukian: It could be a NaN, can you ensure this field is a string

packages/hub/src/lib/parse-safetensors-metadata.ts
```diff
@@ -223,27 +252,45 @@ export async function parseSafetensorsMetadata(
...
 		return {
 			sharded: true,
 			index,
-			headers,
-			...(params.computeParametersCount && {
-				parameterCount: computeNumOfParamsByDtypeSharded(headers),
-			}),
+			headers: shardedMap,
+			...(params.computeParametersCount
+				? {
```
Kakulukian, line 291: The parameter counting logic is mixed with metadata parsing (i.e: total_parameters), I feel it should have a clear separation between metadata parsing and parameter computation

julien-c: in practice we always want to compute parameters, so I think it's fine

packages/hub/src/lib/parse-safetensors-metadata.ts
julien-c: c0fc9f3e3803db52ffcf31013eba5e791b054728

Kakulukian (approved):

julien-c:
i don't think the CI failures are related so i'll merge this one

PR #1539, opened 2025-06-17
Add NVIDIA B200 to hardware.ts

ethnzhng:
Source: https://www.techpowerup.com/gpu-specs/b200-sxm-192-gb.c4210

Add NVIDIA B200 to hardware list. The VRAM per GPU most commonly advertised by datacenters is 180GB, but some sources (like the link above) list the theoretical value at 192GB.


julien-c:
YES!!! we need the B200!! 🙇

pcuenca (approved):

PR #1547, opened 2025-06-20
Fix URL

pcuenca:
Ref: #1546


Vaibhavs10 (commented):
thanks pedro, sorry for making the cardinal mistake of not waiting for enough feedback before merge! 🫡

PR #1555, opened 2025-06-25
Adds support to Docker Model Runner

ilopezluna:
Adds support to run models in GGUF format via [Docker Model Runner](https://docs.docker.com/ai/model-runner/)


packages/tasks/src/local-apps.ts
```diff
@@ -292,11 +292,15 @@ const snippetMlxLm = (model: ModelData): LocalAppSnippet[] => {
 						setup: ["# Install MLX LM", "uv tool install mlx-lm"].join("\n"),
 						content: ["# Start the server", `mlx_lm.server --model "${model.id}"`, ...openaiCurl].join("\n"),
 					},
-			  ]
+				]
```
Vaibhavs10:
```suggestion
			]
```

Vaibhavs10:
Thanks a lot @ilopezluna for the PR, it should be live shortly!

PR #1567, opened 2025-06-28
fix(hub): test list datasets

quanghuynt14:
Hello there 👋 

It's my first time contributing. I tried to set up the project with `pnpm` and got error when running the tests:

```bash
 ⎯⎯⎯⎯⎯⎯⎯ Failed Tests 1 ⎯⎯⎯⎯⎯⎯⎯
│  FAIL  src/lib/list-datasets.spec.ts > listDatasets > should list datasets from hf-doc-builder
│ AssertionError: expected [ { …(7) }, { …(7) } ] to deeply equal [ { …(7) }, { …(7) } ]
│ - Expected
│ + Received
│   Array [
│     Object {
│       "downloads": 0,
│       "gated": false,
│ -     "id": "6356b19985da6f13863228bd",
│ +     "id": "636a1b69f2f9ec4289c4c19e",
│       "likes": 0,
│ -     "name": "hf-doc-build/doc-build",
│ +     "name": "hf-doc-build/doc-build-dev",
│       "private": false,
│       "updatedAt": 1970-01-01T00:00:00.000Z,
│     },
│     Object {
│       "downloads": 0,
│       "gated": false,
│ -     "id": "636a1b69f2f9ec4289c4c19e",
│ +     "id": "6356b19985da6f13863228bd",
│       "likes": 0,
│ -     "name": "hf-doc-build/doc-build-dev",
│ +     "name": "hf-doc-build/doc-build",
│       "private": false,
│       "updatedAt": 1970-01-01T00:00:00.000Z,
│     },
│   ]
│  ❯ src/lib/list-datasets.spec.ts:26:24
│      24|   }
│      25| 
│      26|   expect(results).deep.equal([
│        |                        ^
│      27|    {
│      28|     id: "6356b19985da6f13863228bd",
│ ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯[1/1]⎯
│  Test Files  1 failed | 39 passed (40)
│       Tests  1 failed | 135 passed (136)
│    Start at  02:32:04
│    Duration  14.37s (transform 650ms, setup 3ms, collect 2.44s, tests 65.94s, environment 3ms, prepare 2.74s)
```

So I fix it.

### DIFF
--- a/packages/hub/src/lib/list-datasets.spec.ts
+++ b/packages/hub/src/lib/list-datasets.spec.ts
@@ -23,7 +23,7 @@ describe("listDatasets", () => {
 			results.push(entry);
 		}
 
-		expect(results).deep.equal([
+		expect(results).toEqual(expect.arrayContaining([
 			{
 				id: "6356b19985da6f13863228bd",
 				name: "hf-doc-build/doc-build",
@@ -42,6 +42,6 @@ describe("listDatasets", () => {
 				likes: 0,
 				updatedAt: new Date(0),
 			},
-		]);
+		]));
 	});
 });
--- a/packages/hub/src/lib/list-datasets.spec.ts
+++ b/packages/hub/src/lib/list-datasets.spec.ts
@@ -7,7 +7,7 @@ describe("listDatasets", () => {
 		const results: DatasetEntry[] = [];
 
 		for await (const entry of listDatasets({ search: { owner: "hf-doc-build" } })) {
-			if (entry.name !== "hf-doc-build/doc-build-dev-test") {
+			if (entry.name !== "hf-doc-build/doc-build" && entry.name !== "hf-doc-build/doc-build-dev") {
 				continue;
 			}
 			if (typeof entry.downloads === "number") {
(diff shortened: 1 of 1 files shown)


packages/hub/src/lib/list-datasets.spec.ts
```diff
@@ -7,7 +7,7 @@ describe("listDatasets", () => {
 		const results: DatasetEntry[] = [];
 
 		for await (const entry of listDatasets({ search: { owner: "hf-doc-build" } })) {
-			if (entry.name === "hf-doc-build/doc-build-dev-test") {
+			if (entry.name !== "hf-doc-build/doc-build-dev-test") {
```
quanghuynt14: I'm inspired by `list-spaces.spec.ts`

packages/hub/src/lib/list-datasets.spec.ts
```diff
@@ -25,23 +25,14 @@ describe("listDatasets", () => {
 
 		expect(results).deep.equal([
```
coyotte508:
```suggestion
		expect(results.sort(a, b) => a.id.localeCompare(b.id)).deep.equal([
```

I think this would fix the tests (making sure results are in correct order)

quanghuynt14: Yes, we can do that, but in the future, we might add a new dataset to `hf-doc-build`, which could cause the test to fail. I was looking at the test `list-spaces.spec.ts`, and I noticed that we do the following:

```ts
if (entry.name !== "microsoft/visual_chatgpt") {
  continue;
}
```

And that makes sure we don't have problem when microsoft adds new space. But we will still have a problem if Microsoft deletes the ⁠microsoft/visual_chatgpt space, and that is the trade-off for not mocking. WDYT?

coyotte508: you can change the if above then:

```
if (entry.name === "hf-doc-build/doc-build-dev-test") {
```

to 

```
if (entry.name !== "hf-doc-build/doc-build" && entry.name !== "hf-doc-build/doc-build-dev") {
```

Would rather minimal changes when possible

quanghuynt14: oh yes, you're right. I fixed here https://github.com/huggingface/huggingface.js/pull/1567/commits/27c14a1608b6b0b20ecc3a66e749115ea38826f4

coyotte508 (approved):

quanghuynt14:
@coyotte508 we had an error in the test `src/utils/XetBlob.spec.ts` and CI didn't pass. What can we do to merge this PR?

PR #1568, opened 2025-06-29
feat(hub): list collections

quanghuynt14:
Issue: https://github.com/huggingface/huggingface.js/issues/271

This PR only includes the call of list collections. Why?
- It's one of my early contributions to the project, so I think I should split the [PR](https://github.com/huggingface/huggingface.js/pull/367) of @hackpk into many small PRs.

- > I was really hoping to be able to make a collection of our best models and datasets, so that they can be displayed on our landing page.
- I think list collections solve the problem of displaying collections in real time 

I’m having a bit of trouble with the type, and I need to search through the project to understand it better. We don't have any public schema API?

Thanks for review it.

### DIFF
--- a/packages/hub/src/lib/list-collections.spec.ts
+++ b/packages/hub/src/lib/list-collections.spec.ts
@@ -57,29 +57,37 @@ describe("listCollections", () => {
 			results.push(entry);
 		}
 
-		expect(results[0].slug).toBe("quanghuynt14/test-6862f3b263a5cc509b9ee24a");
-		expect(results[0].title).toBe("Test");
-		expect(results[0].description).toBe("this collection is only for test");
-		expect(results[0].gating).toBe(false);
-		expect(results[0].lastUpdated).toEqual("2025-07-21T00:36:29.000Z");
-		expect(results[0].owner).deep.equal({
-			_id: "6861a8be64cdb219ce188401",
-			avatarUrl: "https://cdn-avatars.huggingface.co/user-avatar.jpeg",
-			fullname: "Huy",
-			name: "quanghuynt14",
-			type: "user",
-			isPro: false,
-			isHf: false,
-			isHfAdmin: false,
-			isMod: false,
+		const collection = results[0];
+		const items = collection.items;
+		collection.items = [];
+
+		// Check all properties of the collection except items
+		expect(collection).deep.equal({
+			slug: "quanghuynt14/test-6862f3b263a5cc509b9ee24a",
+			title: "Test",
+			description: "this collection is only for test",
+			gating: false,
+			lastUpdated: "2025-07-21T00:36:29.000Z",
+			owner: {
+				_id: "6861a8be64cdb219ce188401",
+				avatarUrl: "https://cdn-avatars.huggingface.co/user-avatar.jpeg",
+				fullname: "Huy",
+				name: "quanghuynt14",
+				type: "user",
+				isPro: false,
+				isHf: false,
+				isHfAdmin: false,
+				isMod: false,
+			},
+			items: [],
+			theme: "pink",
+			private: false,
+			upvotes: 0,
+			isUpvotedByUser: false,
 		});
-		expect(results[0].theme).toBe("pink");
-		expect(results[0].private).toBe(false);
-		expect(results[0].upvotes).toBe(0);
-		expect(results[0].isUpvotedByUser).toBe(false);
 
 		// Check for model item
-		expect(results[0].items[0]).deep.equal({
+		expect(items[0]).deep.equal({
 			_id: "68632ede07867696ac5d7ab1",
 			position: 0,
 			type: "model",
@@ -107,7 +115,7 @@ describe("listCollections", () => {
 		});
 
 		// Check for dataset item
-		expect(results[0].items[1]).deep.equal({
+		expect(items[1]).deep.equal({
 			_id: "68632ee9c76d34d6031bab15",
 			position: 1,
 			type: "dataset",
@@ -123,10 +131,12 @@ describe("listCollections", () => {
 		});
 
 		// Check for space item
-		expect(results[0].items[2]).deep.equal({
+		expect(items[2]).deep.equal({
 			_id: "68632eeff8d2d420b6ad4b0d",
 			position: 2,
 			type: "space",
+			ai_category: "Chatbots",
+			ai_short_description: "Chat with an advanced virtual assistant",
 			author: "quanghuynt14",
 			authorData: {
 				_id: "6861a8be64cdb219ce188401",
@@ -179,7 +189,7 @@ describe("listCollections", () => {
 		});
 
 		// Check for collection item
-		expect(results[0].items[3]).deep.equal({
+		expect(items[3]).deep.equal({
 			_id: "68632ef6b13e0dc3b640854a",
 			position: 3,
 			type: "collection",
--- a/packages/hub/src/lib/list-collections.spec.ts
+++ b/packages/hub/src/lib/list-collections.spec.ts
@@ -86,7 +86,7 @@ describe("listCollections", () => {
 			isUpvotedByUser: false,
 		});
 
-		// Check for model item
+		// Check for item type model
 		expect(items[0]).deep.equal({
 			_id: "68632ede07867696ac5d7ab1",
 			position: 0,
@@ -114,7 +114,7 @@ describe("listCollections", () => {
 			isLikedByUser: false,
 		});
 
-		// Check for dataset item
+		// Check for item type dataset
 		expect(items[1]).deep.equal({
 			_id: "68632ee9c76d34d6031bab15",
 			position: 1,
@@ -130,7 +130,7 @@ describe("listCollections", () => {
 			isLikedByUser: false,
 		});
 
-		// Check for space item
+		// Check for item type space
 		expect(items[2]).deep.equal({
 			_id: "68632eeff8d2d420b6ad4b0d",
 			position: 2,
@@ -188,7 +188,7 @@ describe("listCollections", () => {
 			tags: ["gradio", "region:us"],
 		});
 
-		// Check for collection item
+		// Check for item type collection
 		expect(items[3]).deep.equal({
 			_id: "68632ef6b13e0dc3b640854a",
 			position: 3,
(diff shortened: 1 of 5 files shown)


packages/hub/src/lib/list-collections.ts
```diff
@@ -0,0 +1,132 @@
...
+import type { ApiCollectionInfo } from "../types/api/api-collection";
+
+export interface CollectionEntry {
+	slug: string;
+	title: string;
+	description: string;
+	gating: boolean;
+	lastUpdated: Date;
+	owner: string;
+	items: {
```
quanghuynt14: the item type is inspired by the python version: https://github.com/huggingface/huggingface_hub/blob/c8ffea9ca815b10edf07085443d524bac479153d/src/huggingface_hub/hf_api.py#L1189

packages/hub/src/lib/list-collections.ts
```diff
@@ -0,0 +1,136 @@
...
+		fetch?: typeof fetch;
+	} & Partial<CredentialsParams>
+): AsyncGenerator<CollectionEntry> {
+	const accessToken = params && checkCredentials(params);
+
+	let totalToFetch = params?.limit ?? Infinity;
+
+	const search = new URLSearchParams([
+		...Object.entries({
+			limit: String(Math.min(totalToFetch, 100)),
```
quanghuynt14: by the [doc](https://huggingface.co/docs/hub/api#get-apicollections), max number of collections par page is 100.

packages/hub/src/types/api/api-collection.ts
```diff
@@ -0,0 +1,22 @@
+export interface ApiCollectionInfo {
+	slug: string;
+	title: string;
+	description?: string;
+	gating: boolean;
+	lastUpdated: string;
+	owner: {
+		name: string;
```
quanghuynt14: The owner type includes more properties, but I only provided the `name` here. Then, format it to the `CollectionEntry` type, where owner is a string representing the name.

packages/hub/src/lib/list-collections.spec.ts
coyotte508, line 1: just create a test collection under your user account/org, and use that, instead of overwriting every prop :)

packages/hub/src/lib/list-collections.ts
```diff
@@ -0,0 +1,136 @@
...
+				description: collection.description,
+				gating: collection.gating,
+				lastUpdated: new Date(collection.lastUpdated),
+				owner: collection.owner.name,
+				items: collection.items.map((collectionItem) => ({
+					item_object_id: collectionItem._id,
+					item_id: collectionItem.id,
+					item_type: collectionItem.type,
+					position: collectionItem.position,
+				})),
```
coyotte508: Would rather just forward as is - rather than change prop names

That way we can also catch with E2Es when new props are added (and modify the type to reflect that)

packages/hub/src/lib/list-collections.ts
```diff
@@ -0,0 +1,136 @@
...
+		const collections: ApiCollectionInfo[] = await res.json();
+
+		for (const collection of collections) {
+			yield {
+				slug: collection.slug,
+				title: collection.title,
+				description: collection.description,
+				gating: collection.gating,
+				lastUpdated: new Date(collection.lastUpdated),
+				owner: collection.owner.name,
```
coyotte508: Same here

coyotte508 (commented):
Thanks for the PR!

I will look into providing a schema for the existing collection APIs

quanghuynt14:
> I will look into providing a schema for the existing collection APIs

If we’re using `⁠json-schema`, we can use ⁠`json-schema-to-typescript` to generate the types.

Without the schema, it's very difficult to guess what it is. By [reverse-engineering 10,000 collections](https://github.com/quanghuynt14/reverse-engineering-hf-api), I identified the type, but not 100% sure it's correct.

coyotte508:
Hey @quanghuynt14 , here's an OpenAPI JSON spec for some of the hub endpoints, including collection ones: https://gist.github.com/coyotte508/ec7d12713d94a4fe4988afa3fffc23f1

It includes json schema for the various responses & requests, and you can also load the openapi schema on https://editor-next.swagger.io/ for example

do you think it's helpful for your PRs?

quanghuynt14:
Yes, it's very complete and helpful. I will check if the type in this PR is correct.

quanghuynt14:
I checked the schema and found a mismatch in the schema of `"/api/collections/{namespace}/{slug}-{id}"`: 

We have:
-  required prop `"position": { "type": "number" }` 
-  required prop `"shareUrl": { "type": "string" }`

in the `collection` object.

After verifying `"slug": "google/gemma-3n-685065323f5984ef315c93f4"`, I found that it does not include this property.
```json
{
    "slug": "google/gemma-3n-685065323f5984ef315c93f4",
    "title": "Gemma 3n",
    "description": "",
    "gating": false,
    "lastUpdated": "2025-06-26T15:55:44.512Z",
    "owner": {
      ...
    },
    "items": [
      ...
    ],
    "theme": "purple",
    "private": false,
    "upvotes": 146,
    "isUpvotedByUser": false
  },
```

These 2 props (`position` and `shareUrl`) are present only in `items` of type `"collection"`. For example:
```json
{
    "slug": "kristenq/reasoning-684330e0ce0c4e30fe59456a",
    "title": "Reasoning",
    "description": "Advanced reasoning models. ",
    "gating": false,
    "lastUpdated": "2025-06-06T18:18:26.733Z",
    "owner": {
      ...
    },
    "items": [
      {
        "_id": "684330f27ce524322498baa7",
        "position": 0,
        "type": "collection",
        "id": "67ee7145ec3d31f7c7a75cab",
        "slug": "Tesslate/synthia-s1-reasoning-model-67ee7145ec3d31f7c7a75cab",
        "title": "Synthia-S1 REASONING MODEL",
        "description": "Creative, Scientific, and Coding",
        "lastUpdated": "2025-04-03T11:31:05.262Z",
        "numberItems": 3,
        "owner": {
          ...
        },
        "theme": "blue",
        "shareUrl": "https://huggingface.co/collections/Tesslate/synthia-s1-reasoning-model-67ee7145ec3d31f7c7a75cab", 
        "upvotes": 3,
        "isUpvotedByUser": false
      }
    ],
    "theme": "green",
    "private": false,
    "upvotes": 0,
    "isUpvotedByUser": false
  },
```

Here is the corrected schema: https://gist.github.com/quanghuynt14/1c55f44978248bc12889b1bde2359c0c

- I deleted prop `position` in the collection object.
- I moved prop `shareUrl` into `items` type `collection`.

coyotte508:
hmm not sure, https://huggingface.co/api/collections/google/gemma-3n-685065323f5984ef315c93f4 I do have `position` and `shareUrl`:
image

quanghuynt14:
ah sorry. I assumed that the type of collection returned by `GET /api/collections/{namespace}/{slug}-{id}` was the same as the ones returned by `GET /api/collections` but it's not :/

The collection object within the collection list does not contain `position` or `shareUrl` ✅ 

But the `shareUrl` of an item of type `collection` is not present in the schema; however, I see it in "open-sci/open-sci-ref-releases-684210f647554eb6e0bb9cf3" 
- https://huggingface.co/api/collections?limit=100&owner=open-sci
- https://huggingface.co/api/collections/open-sci/open-sci-ref-releases-684210f647554eb6e0bb9cf3

coyotte508:
yes indeed, we'll fix the schema on our side, thanks for pointing out

quanghuynt14:
You can review the PR. I think the type is now complete.

packages/hub/src/lib/list-collections.ts
coyotte508: technically it can be 10000 if all of:

- `owner` is set
- `expand` is set to false explicitly

but not sure we want to bother

packages/hub/src/lib/list-collections.ts
```diff
@@ -0,0 +1,83 @@
...
+	let totalToFetch = params?.limit ?? Infinity;
+
+	const search = new URLSearchParams([
+		...Object.entries({
+			limit: String(Math.min(totalToFetch, 100)),
+			...(params?.search?.owner ? { owner: Array.isArray(params.search.owner) ? params.search.owner.join(",") : params.search.owner } : undefined),
+			...(params?.search?.item ? { item: Array.isArray(params.search.item) ? params.search.item.join(",") : params.search.item } : undefined),
+			...(params?.search?.q ? { q: params.search.q } : undefined),
+		}),
+	]).toString();
```
coyotte508: When `params.search.owner` and `params.search.item` are arrays, we should just repeat it, eg:

```
for (const owner of params.search.owner) {
  search.append("owner", params.search.owner)
}
```

Eg https://huggingface.co/api/collections?owner=coyotte508&owner=julien-c if `owner=['coyotte508','julien-c']`

In constrast, https://huggingface.co/api/collections?owner=coyotte508,julien-c seems to return all collections... which is a bug that's getting fixed now it's been found!

---

By the way, for the JS lib we can force the user to put arrays for both `owner` and `item` if we want to.

coyotte508 (commented):
Thanks a lot!

I think just one change on how `owner` and `item` params are handled and it'll be ready to merge (I'll still re-review)

packages/hub/src/lib/list-collections.ts
quanghuynt14: yes, I fixed it ✅

packages/hub/src/types/api/api-collection.ts
```diff
@@ -0,0 +1,462 @@
...
+		isHf: boolean;
+		isHfAdmin: boolean;
+		isMod: boolean;
+		followerCount?: number;
+		type: "user";
+		isPro: boolean;
+		_id: string;
+		isUserFollowing?: boolean;
+	};
```
coyotte508: Can you extract this to `ApiAuthor`?

coyotte508 (approved):
Thanks looks good!

packages/hub/src/types/api/api-collection.ts
quanghuynt14: Yes https://github.com/huggingface/huggingface.js/pull/1568/commits/ad702e7fe96aa7d47916e9a83ebc5f366e0ff212

julien-c:
Thanks for your contribution @quanghuynt14! 🔥

PR #1575, opened 2025-06-30
Add support for image-to-video task type for Replicate

zeke:
This should unblock support for Replicate-powered inference for https://huggingface.co/MeiGen-AI/MeiGen-MultiTalk

cc @lucataco @apolinario @SBrandeis @Wauplin


SBrandeis:
Nice, thank you!

Is the end goal to map to https://replicate.com/zsxkib/multitalk/api/api-reference ? if so, the proposed change won't work (it does not match the API of model on Replicate).

Unfortunately there don't seem to be a standard API for image-to-video on Replicate (iiuc every model has a different, specific API), which will require editing this client code everytime a model needs to be supported.
For example, the API for https://replicate.com/kwaivgi/kling-v2.1/api/api-reference and https://replicate.com/zsxkib/multitalk/api/api-reference differ quite significantly.

Until Replicate all those models converge to a common, standard API, I don't think it will be possible for us to properly support the image-to-video task 😞

SBrandeis:
Note: I opened https://github.com/huggingface/huggingface.js/pull/1578 to add general support for the task in the Inference client

PR #1579, opened 2025-07-01
[Inference Providers] Fix structured output

hanouticelina:
equivalent PR to https://github.com/huggingface/huggingface_hub/pull/3082.
`sambanova` and `nebius` don’t fully follow OpenAI’s spec for the `response_format` field. This PR adds internal mappings for each provider. Seems like `fireworks-ai` is now following OpenAI specs for this field (double checked their docs as well).

### DIFF
--- a/packages/inference/src/providers/nebius.ts
+++ b/packages/inference/src/providers/nebius.ts
@@ -25,6 +25,7 @@ import {
 	type TextToImageTaskHelper,
 } from "./providerHelper.js";
 import { InferenceClientProviderOutputError } from "../errors.js";
+import type { ChatCompletionInput } from "../../../tasks/dist/commonjs/index.js";
 
 const NEBIUS_API_BASE_URL = "https://api.studio.nebius.ai";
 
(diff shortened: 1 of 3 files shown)


Wauplin (approved):
Looks good to me!

> Seems like together and fireworks-ai are now following OpenAI specs for this field (double checked their docs as well).

Does it mean Python SDK broke for those 2 ? :smile:

hanouticelina:
> Does it mean Python SDK broke for those 2 ? 😄

for fireworks-ai, it still works (I mean, they definitely shouldn't break that and not keep it backward compatible 😄) as for Together,  I initially thought they were now totally compatible, but it turns out they're still not fully compatible just yet, I pushed [bf05b9b](https://github.com/huggingface/huggingface.js/pull/1579/commits/bf05b9b259101a4b38af30c9c14997633bbb6053) to add a mapping.

julien-c:
(take with a grain of salt as i've been following from afar)

They're all likely to change to be fully OpenAI compliant so we could write code that conditionally support the current situation but also works if they move to full compliance, no?

hanouticelina:
@julien-c i tried to make things forward compatible, for Nebius, we already are. For together, their API rejects requests that contain both a `schema` key (their specs) and a `json_schema` key (OAI compatible) inside `response_format`. Same for sambanova, i'm not sure we can be forward compatible, they don't support strict mode so we have to hardcode the value to false for now 🤷‍♀️

julien-c:
ah yes re-reading i'm a bit dumb i was thinking of supporting evolutions in responses, but in requests there's not really a way to do it.

PR #1584, opened 2025-07-02
feat(hub): create and delete a collection

quanghuynt14:
Issue: https://github.com/huggingface/huggingface.js/issues/271

**API doc:** 
- create: https://huggingface.co/docs/hub/api#post-apicollections
- delete: https://huggingface.co/docs/hub/api#delete-apicollectionsnamespaceslug-id

**Python client:**
- create: https://huggingface.co/docs/huggingface_hub/v0.33.1/en/package_reference/hf_api#huggingface_hub.HfApi.create_collection
- delete: https://huggingface.co/docs/huggingface_hub/v0.33.1/en/package_reference/hf_api#huggingface_hub.HfApi.delete_collection

I decided not to include the first item in then param input when creating a collection (like the Python client).


packages/hub/src/lib/create-collection.ts
```diff
@@ -0,0 +1,39 @@
...
+import { checkCredentials } from "../utils/checkCredentials";
+
+export async function createCollection(
+	params: {
+		hubUrl?: string;
+		/**
+		 * Custom fetch function to use instead of the default one, for example to use a proxy or edit headers.
+		 */
+		fetch?: typeof fetch;
+	} & ApiCreateCollectionPayload &
```
quanghuynt14: should we put the payload to a variable `params.payload`? That way, if we decide to add new prop to `ApiCreateCollectionPayload` in the future, we will not need to remap the entire body as we foward all the props from `params.payload`.

coyotte508: yes, but name the parameter `collection`

packages/hub/src/types/api/api-create-collection.ts
```diff
@@ -0,0 +1,26 @@
...
+	 */
+	private?: boolean;
+	/**
+	 * First item to add to the collection upon creation.
+	 */
+	item?: {
+		type: "paper" | "collection" | "space" | "model" | "dataset";
+		id: string;
+	};
+}
```
coyotte508: TBH not sure it's worth it to add this param to the API for the hub library, since the user will most of the time want to add other items.

(It'd be nice if we had a way to init collection with multiple items, cc @Kakulukian)

quanghuynt14: yes, I think so too. I just removed it for this PR (and it's in line with what we do in the [Python client](https://github.com/huggingface/huggingface_hub/blob/6e944cf4b36856a359229e85b369e4b4a7380c67/src/huggingface_hub/hf_api.py#L8346))

packages/hub/src/lib/delete-collection.ts
```diff
@@ -0,0 +1,32 @@
...
+import { createApiError } from "../error";
+import type { CredentialsParams } from "../types/public";
+import { checkCredentials } from "../utils/checkCredentials";
+
+export async function deleteCollection(
+	params: {
+		/**
+		 * The slug of the collection to delete.
+		 */
+		collectionSlug: string;
```
coyotte508:
```suggestion
		slug: string;
```

quanghuynt14: yes, I also change it in `create-collection.ts`. WDYT?

quanghuynt14:
I was a bit confused when I saw the endpoint: `GET
/api/collections/{namespace}/{slug}-{id}` (https://huggingface.co/spaces/huggingface/openapi#tag/collections/get/api/collections/{namespace}/{slug}-{id})

But in fact, `slug = {namespace}/{slug}-{id}`

We also have `collectionSlug` in `collection-info.ts` (https://github.com/quanghuynt14/huggingface.js/blob/312da8611297a32231cef1652e060e0b6731298d/packages/hub/src/lib/collection-info.ts#L12). Should I create a PR refactor to rename `collectionSlug` to `slug`?

coyotte508:
> We also have `collectionSlug` in `collection-info.ts` ([quanghuynt14/huggingface.js@`312da86`/packages/hub/src/lib/collection-info.ts#L12](https://github.com/quanghuynt14/huggingface.js/blob/312da8611297a32231cef1652e060e0b6731298d/packages/hub/src/lib/collection-info.ts#L12)). Should I create a PR refactor to rename `collectionSlug` to `slug`?

You can add it to this PR

And yes it's confusing

quanghuynt14:
Hey @coyotte508, there are failed tests but they are unrelated. We merge this PR?

PR #1592, opened 2025-07-04
[Inference Providers] Fix LoRAs 

hanouticelina:
`adapter` and `adapterWeightsPath` were left out when normalizing inference provider mappings. This should fix LoRAs support.

cc @Vaibhavs10


SBrandeis:
Does that mean LoRAs were broken for 3 weeks? 🫠

PR #1598, opened 2025-07-07
Add custom inference logger for @hf/inference

Kakulukian:
The goal is to pass a custom logging tool (needed mainly by moon). 

I chose a global overwrite instead of an option like fetch to avoid too many code changes (btw, there are some places where fetch is not properly overwritten by the options).

I think the best approach would be to have a separate package for logging to use the same module in the hf.js ecosystem

### DIFF
--- a/packages/inference/src/providers/black-forest-labs.ts
+++ b/packages/inference/src/providers/black-forest-labs.ts
@@ -19,6 +19,7 @@ import {
 	InferenceClientProviderApiError,
 	InferenceClientProviderOutputError,
 } from "../errors.js";
+import { getLogger } from "../lib/logger.js";
 import type { BodyParams, HeaderParams, UrlParams } from "../types.js";
 import { delay } from "../utils/delay.js";
 import { omit } from "../utils/omit.js";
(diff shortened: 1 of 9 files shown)


SBrandeis (approved):
LGTM - I think your changes regarding the input types in packages/inference/src/lib/getInferenceProviderMapping.ts broke the tests

packages/inference/src/tasks/audio/audioClassification.ts
```diff
@@ -16,7 +16,7 @@ export async function audioClassification(
 	args: AudioClassificationArgs,
 	options?: Options
 ): Promise<AudioClassificationOutput> {
-	const provider = await resolveProvider(args.provider, args.model, args.endpointUrl);
+	const provider = await resolveProvider(args.provider, args.model, args.endpointUrl, options);
```
coyotte508: I don't get this and why the change in `Options`, since `getLogger` is called each time

Kakulukian: I passed the logger in options first, then rolled back without reverting everything - that's my bad

Kakulukian: There's also an issue with resolveProvider, it uses fetch but doesn't get the options fetch, but that's a separate issue

coyotte508 (commented):
LGTM, but up to @hanouticelina @SBrandeis @Wauplin

Wauplin (approved):
LGTM if it helps moon-landing integration 👍

packages/inference/src/providers/black-forest-labs.ts
```diff
@@ -67,10 +68,11 @@ export class BlackForestLabsTextToImageTask extends TaskProviderHelper implement
 		headers?: HeadersInit,
 		outputType?: "url" | "blob"
 	): Promise<string | Blob> {
+		const logger = getLogger();
```
julien-c, line 71: could have been called `c` 🎉

PR #1607, opened 2025-07-10
[Inference Providers] fix: zero-shot-classification HF helper

SBrandeis:
Flagged by @oOraph 

All zero-shot classification mappings are currently failing for HF Inference 

Two things:
- The validation function was incorrect / not actually validating the expected format as defined in tasks
- The HF Inference backend does not return the expected format (it's using a "legacy" format)

This PR addresses those two issues by fixing the response validation code to match and handling previous input to ease transition of the hf-inference background


SBrandeis:
Nope !

Wauplin (approved):
Looks good, thanks!

PR #1611, opened 2025-07-11
Richer conversational snippet for AutoModel

pcuenca:
I can't remember if we already discussed about this. This is an attempt to make the `AutoModel` snippet more self-contained, for the particular but frequent case of conversational transformers models. The pipeline version already got this treatment in #1434, but the `AutoModel` version is currently too barebones in comparison, as shown in the screenshot below.

In my opinion this could reduce the friction for users that want to experiment with the models in Colab or Kaggle.
cc @Vaibhavs10 @ariG23498 

Screenshot 2025-07-11 at 14 05 15

### DIFF
--- a/packages/tasks/src/model-libraries-snippets.ts
+++ b/packages/tasks/src/model-libraries-snippets.ts
@@ -1386,7 +1386,7 @@ export const transformers = (model: ModelData): string[] => {
 			`from transformers import ${info.processor}, ${info.auto_model}`,
 			"",
 			`${processorVarName} = ${info.processor}.from_pretrained("${model.id}"` + remote_code_snippet + ")",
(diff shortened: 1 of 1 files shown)


pcuenca:
I wanted to start the discussion because longer snippets could perhaps start making sense now that we auto-create notebooks, but happy to defer for now.

julien-c:
that's fine we can always try imo! let's see what others think too

Vaibhavs10 (approved):
LGTM and thanks! my only worry is that it makes the overall snippet dialog a bit too huge, maybe we can collapse the `AutoModel` snippet by default on UI. WDYT @julien-c @gary149 

Makes sense to start conservative with `image-text-to-text` and `text-generation` to begin with and the logic looks good to me too.

Otherwise let's merge and see how the users react to it!

PR #1616, opened 2025-07-11
Basic Xorb creation

coyotte508:
This doesn't handle dedup at all yet, will not be used as is even if I merge it.

What it does:

- Write chunks into xorb
- Write chunk headers
- Handle compression
- Handle xorb size limit and xorb chunk count limit

Questions:

- Are there xorb headers? I'm writing chunk headers before each chunk, but is there a xorb header at the beginning of the xorb to write?
- For now I only test lz4 compression, I do not bother with bg4. Should I test both?
- Is the xet backend hardened against invalid chunks/xorbs?

cc @assafvayner @seanses @hoytak @sirahd @rajatarya  for viz and if you have answers to the questions :)


packages/hub/src/vendor/xet-chunk/chunker_wasm_bg.wasm.base64.ts
coyotte508, line 1: Hopefully we'll be able to bring that wasm size down (eg maybe trimming down the merkle tree / hash logic for the xorb hash to the [mininum](https://github.com/huggingface/huggingface.js/blob/0065d2b58ccc6b7e6963950f21e06d7422614636/packages/xetchunk-wasm/assembly/xorb-hash.ts) would help)

packages/hub/src/vendor/xet-chunk/chunker_wasm.ts
coyotte508, line 1: rewrote the file generated by wasmbindgen to TS + make it work in node & web + add `init()` function

packages/hub/src/utils/createXorbs.ts
```diff
@@ -0,0 +1,121 @@
+/**
+ * Todo: add dedup: we actually need to remember chunks already written, and not add themm to the xorb, and also
```
Kakulukian:
```suggestion
 * Todo: add dedup: we actually need to remember chunks already written, and not add them to the xorb, and also
```

packages/hub/src/utils/createXorbs.ts
```diff
@@ -0,0 +1,121 @@
...
+			if (done) {
+				yield* addChunks(chunker.finish());
+				break;
+			}
+			sourceChunks.push(value);
+			yield* addChunks(chunker.add_data(value));
+		}
+	} finally {
+		chunker.free();
+		// ^ is this really needed ?
```
Kakulukian, line 83: WASM module can allocate memory that exists outside js garbage collector, so it depends really on the wasm Chunker code

packages/hub/src/utils/createXorbs.ts
```diff
@@ -0,0 +1,121 @@
...
+			for (const chunk of chunks) {
+				let chunkToCopy: Uint8Array;
+				if (chunk.length === sourceChunks[0].length) {
+					chunkToCopy = sourceChunks[0];
+					sourceChunks.shift();
+				} else if (chunk.length < sourceChunks[0].length) {
+					chunkToCopy = sourceChunks[0].subarray(0, chunk.length);
+					sourceChunks[0] = sourceChunks[0].subarray(chunk.length);
+				} else {
+					chunkToCopy = new Uint8Array(chunk.length);
```
Kakulukian, line 41: Could be optimized for less memory allocation btw

packages/hub/src/utils/createXorbs.ts
```diff
@@ -0,0 +1,121 @@
...
+		const reader = fileSource.stream().getReader();
+		let xorbOffset = 0;
+		let xorbChunks = Array<{ hash: string; length: number }>();
+
+		const addChunks = function* (chunks: Array<{ hash: string; length: number }>) {
+			for (const chunk of chunks) {
+				let chunkToCopy: Uint8Array;
+				if (chunk.length === sourceChunks[0].length) {
+					chunkToCopy = sourceChunks[0];
+					sourceChunks.shift();
```
Kakulukian, line 36: This looks more inefficient than using an index approach no?

packages/hub/src/vendor/xet-chunk/chunker_wasm.ts
```diff
@@ -0,0 +1,62 @@
...
+		return initPromise;
+	}
+	let resolve: (value: void) => void;
+	let reject: (reason?: unknown) => void;
+	initPromise = new Promise((_resolve, _reject) => {
+		resolve = _resolve;
+		reject = _reject;
+	});
+
+	await Promise.resolve();
```
Kakulukian, line 21: A bit suspicious as a method to create a singleton you can't recreate if the wasm import is failing

packages/hub/src/utils/createXorbs.ts
coyotte508: this is just the leftover data from the chunking process, there should be at most 128kB of it I think and even in 16kB chunks it's only 8 chunks, eg `sourceChunks.length <= 8`

So with those params I don't think we care too much

packages/hub/src/utils/createXorbs.ts
coyotte508: one thing we could do is have a permanent MAX_CHUNK_SIZE Uint8 array that we reuse?

(since we only have one `chunkToCopy` at a time)

packages/hub/src/utils/createXorbs.ts
coyotte508: There's some code in `chunker_wasm_bg.js` but I can't tell for sure, so being prudent atm

packages/hub/src/vendor/xet-chunk/chunker_wasm.ts
coyotte508: this `await` in particular is to enforce `resolve` and `reject` are assigned before running the rest of the code

not sure it's necessary

packages/hub/src/vendor/xet-chunk/chunker_wasm_bg.wasm.base64.ts
coyotte508: thanks to @hoytak's work we already had a 30% reduction :)

coyotte508:
Merging for now, just added wasm generation directly from xet-core, thanks @assafvayner and @hoytak

And testing bg4 compression

Will open another PR for shard generation

don't worry it's not exported / availalbe in the published lib yet

PR #1617, opened 2025-07-11
[Inference] get the raw response for textToImage

krampstudio:
Allow to get the full response for `text-to-image` inference, instead of directly the blob or the url:

```js
const res = await textToImage({ inputs:  "A dog with a baseball cap" },{ outputType: "json" });
``` 

we have the following ouput (depends on the provider)

```json
{
  "images": [
    {
      "url": "data:image/png;base64,iVBORw0KGgoAAAAN......rkJggg==",
      "width": 1024,
      "height": 768,
      "content_type": "image/png"
    }
  ],
  "timings": {
    "inference": 1.2445615287870169
  },
  "seed": 982874191,
  "has_nsfw_concepts": [
    false
  ],
  "prompt": "A dog with a baseball cap"
} 
```

needed for (internal https://github.com/huggingface-internal/moon-landing/pull/14421)

### DIFF
--- a/packages/inference/src/providers/black-forest-labs.ts
+++ b/packages/inference/src/providers/black-forest-labs.ts
@@ -66,8 +66,8 @@ export class BlackForestLabsTextToImageTask extends TaskProviderHelper implement
 		response: BlackForestLabsResponse,
 		url?: string,
 		headers?: HeadersInit,
-		outputType?: "url" | "blob"
-	): Promise<string | Blob> {
+		outputType?: "url" | "blob" | "json"
+	): Promise<string | Blob | Record<string, unknown>> {
 		const logger = getLogger();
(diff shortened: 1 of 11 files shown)


coyotte508:
Why do we need that? 

In the internal link, we only do `const imageBlob = isImage(res) ? res : new Blob([res], { type: "image/png" });` so it seems we could just call `textToImage` with `output: "blob"` ?

krampstudio:
> Why do we need that?
> 
> In the internal link, we only do `const imageBlob = isImage(res) ? res : new Blob([res], { type: "image/png" });` so it seems we could just call `textToImage` with `output: "blob"` ?

I've updated the internal link, we need the full response for educational purpose: we show it to the user. 

And more generaly, it can be usefull for developers to access the full response.

PR #1620, opened 2025-07-11
Add download stats filter for MagentaRT

osanseviero:


packages/tasks/src/model-libraries.ts
```diff
@@ -565,6 +565,12 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		repoUrl: "https://github.com/SandAI-org/MAGI-1",
 		countDownloads: `path:"ckpt/vae/config.json"`,
 	},
+	"magenta-rt": {
+		prettyLabel: "magenta-realtime",
+		repoName: "Magenta RT",
```
pcuenca: I think the latest recommendation is to use the same content for both these fields (they might be joined soon).

packages/tasks/src/model-libraries.ts
```diff
@@ -565,6 +565,12 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
...
+		repoUrl: "https://github.com/magenta/magenta-realtime",
+		countDownloads: `path:"checkpoints/llm_base_x4286_c1860k.tar" OR path:"checkpoints/llm_large_x3047_c1860k.tar" OR path:"checkpoints/llm_large_x3047_c1860k/checkpoint"`,
```
pcuenca, line 572: Is the last path (`checkpoints/llm_large_x3047_c1860k/checkpoint`) correct?

pcuenca (commented):
Looks good! We'd also need a corresponding PR in https://huggingface.co/google/magenta-realtime to set the library name, wouldn't we?

packages/tasks/src/model-libraries.ts
osanseviero: Yes, this corresponds to https://huggingface.co/google/magenta-realtime/blob/main/checkpoints/llm_large_x3047_c1860k/checkpoint

packages/tasks/src/model-libraries.ts
osanseviero: Updated

chrisdonahue-goog:
Library name has been set, thanks!

https://huggingface.co/google/magenta-realtime/commit/c1578ac0682a4849f6313d8e57ff9b6c6139fcc8

packages/tasks/src/model-libraries.ts
```diff
@@ -565,6 +565,12 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		repoUrl: "https://github.com/SandAI-org/MAGI-1",
 		countDownloads: `path:"ckpt/vae/config.json"`,
 	},
+	"magenta-realtime": {
+		prettyLabel: "magenta-realtime",
```
```suggestion
		prettyLabel: "Magenta RT",
```

this is the name that will show-up here. Usually better to use a human-friendly naming (as for repo name)

Image

Wauplin (approved):
LGTM!

KEHANG:
@Wauplin thanks for the review! I'm one of the Magenta RealTime owners, looking forward to getting this merged. Do you happen to have any insights on why the browser test is failing on this case?

Wauplin:
Hi @KEHANG , failing CI is unrelated so we can merge it now :) Expect 3-4 days before getting it live on the Hub!

KEHANG:
Thanks @Wauplin !

PR #1630, opened 2025-07-16
[Inference] openai client snippet for "auto" provider

krampstudio:
image

(also needed for internal https://github.com/huggingface-internal/moon-landing/pull/14421)


Wauplin (approved):
Thanks!

PR #1635, opened 2025-07-17
[Tasks] disable chat completion automatic specs generation

hanouticelina:
https://github.com/huggingface/huggingface.js/pull/1610#discussion_r2200286219 

let's not generate the chat completion specs from TGI, the schema is currently stable, and we want to ensure full OpenAI-compatibility for chat completion, and we ve already updated the schema manually, so relying on automatic generation is unnecessary now.


Wauplin (approved):
Thanks!

PR #1643, opened 2025-07-18
Update VideoPrism function call name in model-libraries-snippets.ts

jenjsun:
We updated the some functions in VideoPrism to make the model loading work with model.id from the huggingface repo; this is a small update to videoprism inside model-libraries-snippets.ts so the code still runs


Vaibhavs10 (approved):

Vaibhavs10 (commented):
Thanks for fixing the snippet, let's wait for another approval before merging

Wauplin (approved):
Nice improvement!!

PR #1651, opened 2025-07-24
Deprecate pendingSSO, missingMFA fields

Kakulukian:
This pull request updates the `UserInfo` interface in `packages/hub/src/lib/oauth-handle-redirect.ts` to improve how security restrictions are represented. It introduces a new `securityRestrictions` field and marks two existing fields as deprecated.

### Updates to the `UserInfo` interface:

* Added a new `securityRestrictions` field to represent various security restrictions (`"mfa"`, `"sso"`, `"ip"`, `"token-policy"`) in a unified way.
* Deprecated the `pendingSSO` field, recommending the use of `securityRestrictions` with `"sso"` instead.
* Deprecated the `missingMFA` field, recommending the use of `securityRestrictions` with `"mfa"` instead.

### DIFF
--- a/packages/hub/src/lib/oauth-handle-redirect.ts
+++ b/packages/hub/src/lib/oauth-handle-redirect.ts
@@ -77,17 +77,26 @@ export interface UserInfo {
 		 */
 		roleInOrg?: string;
 		/**
+		 * @deprecated Use securityRestrictions instead with "sso"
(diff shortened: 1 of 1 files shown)


coyotte508 (approved):
note that there is auth0 and nextauth to update probably :grimacing:

PR #1661, opened 2025-07-29
update installation instructions for mlx on CUDA.

Vaibhavs10:
The intent is to allow users on CUDA backend to know how to install the library in addition to this, it may allow non-quantized models to work with MLX on Colab too.

cc: @awni


pcuenca (approved):
Awesome, thank you @Vaibhavs10! 🙌

We could potentially look in `model.config.quantization_config` if we want to restrict to certain types, but we can do it if/when we need it.

PR #1664, opened 2025-08-01
Fix model download count for Canary-Qwen-2.5B (safetensors format in NeMo)

pzelasko:


packages/tasks/src/model-libraries.ts
```diff
@@ -703,7 +703,7 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		repoUrl: "https://github.com/NVIDIA/NeMo",
 		snippets: snippets.nemo,
 		filter: true,
-		countDownloads: `path_extension:"nemo" OR path:"model_config.yaml"`,
+		countDownloads: `path_extension:"nemo" OR path:"model_config.yaml" OR path_extension:"safetensors"`,
```
Vaibhavs10:
```suggestion
		countDownloads: `path_extension:"nemo" OR path:"model_config.yaml" OR path_extension:"json"`,
```

better to do config.json IMO: https://huggingface.co/nvidia/canary-qwen-2.5b/blob/main/config.json

pzelasko: Done

Vaibhavs10:
Should be live in 24-48 hours! 🤗

Issue #1280, opened 2025-03-12
Optimize stream operations


coyotte508:
Off the top of my head
 
- Use the byte utilities provided by creating a `ReadableStream` with `{type: "bytes"}` that avoid copying data to stream internals
- Avoid copying data when doing `result.value = new Uint8Array([...leftoverBytes, ...result.value]);`
- maybe check lz4 decompression performance & impact

This is to investigate, to see if any of those has any real impact

PR #1667, opened 2025-08-04
perf: uint8array set method instead of spread

vvnsrzn:
From https://github.com/huggingface/huggingface.js/issues/1280

Benchmark: https://jsben.ch/jze3P


packages/hub/src/utils/XetBlob.ts
```diff
@@ -327,8 +327,11 @@ export class XetBlob extends Blob {
 					totalFetchBytes += result.value.byteLength;
 
 					if (leftoverBytes) {
-						result.value = new Uint8Array([...leftoverBytes, ...result.value]);
-						leftoverBytes = undefined;
```
coyotte508, line 331: I think the `leftoverBytes = undefined`  part is important

coyotte508: can you also make the function into a `combineUint8Arrays` helper? ( a dedicated file in this folder)

vvnsrzn: Of course, I'll take care of it.

coyotte508:
some formatting issues, you can run `pnpm -r format` or `pnpm --filter hub format`

vvnsrzn:
> some formatting issues, you can run `pnpm -r format` or `pnpm --filter hub format`

Oh sorry for that, I didn't run the right command.

PR #1668, opened 2025-08-04
add wan2.2 library

kelseyee:
Add wan2.2 library support to count downloads correctly

### DIFF
--- a/packages/tasks/src/model-libraries.ts
+++ b/packages/tasks/src/model-libraries.ts
@@ -1170,6 +1170,12 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		countDownloads: `path_extension:"pt"`,
 		snippets: snippets.vui,
 	},
+	"wan2.2":{ 
+		prettyLabel: "Wan2.2",
+		repoName: "Wan2.2",
+		repoUrl: "https://github.com/Wan-Video/Wan2.2",
+		countDownloads:  `path:"high_noise_model/config.json"`,
+	},
 	wham: {
 		prettyLabel: "WHAM",
 		repoName: "wham",


Wauplin:
Hi @kelseyee , thanks for the PR! Is "Wan2.2" really a new library or just an extension of a broader "Wan" library? (genuinely asking).

I have checked https://huggingface.co/models?other=wan2.2 and it doesn't seem to me that there is a common pattern to store files in these repos. I would prefer to see a bit more consistency on how the library deals with files before adding the download counter rule.

kelseyee:
> Hi @kelseyee , thanks for the PR! Is "Wan2.2" really a new library or just an extension of a broader "Wan" library? (genuinely asking).
> 
> I have checked https://huggingface.co/models?other=wan2.2 and it doesn't seem to me that there is a common pattern to store files in these repos. I would prefer to see a bit more consistency on how the library deals with files before adding the download counter rule.

"Wan2.2" is actually a new, standalone library, and the download counter rule is only being enabled specifically for Wan2.2. This is because Wan2.2 is a special MoE model — future versions of the Wan model may not be MoE-based, so we're treating it separately.

Also, I was wondering, since the config.json is the default query file, why doesn't it work when it's placed in a subfolder? Is there something in the current logic that expects it to be at the root level? Just trying to better understand how it works.

Wauplin:
>  Is there something in the current logic that expects it to be at the root level? 

Yes the default counting rule counts only the config.json file at root. So doing what you just did is correct.

packages/tasks/src/model-libraries.ts
```diff
@@ -1170,6 +1170,12 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		countDownloads: `path_extension:"pt"`,
 		snippets: snippets.vui,
 	},
+	"wan2.2": { 
+		prettyLabel: "Wan2.2",
+		repoName: "Wan2.2",
+		repoUrl: "https://github.com/Wan-Video/Wan2.2",
+		countDownloads: `path:"high_noise_model/config.json" OR path:"low_noise_model/config.json"`,
```
```suggestion
		countDownloads: `path_filename:"config" AND path_extension:"json"`,
```

An even better rule is to set it to "filename is config and extension is json" => this way all config.json files in all subdirectories are counted correctly. Meaning it'll work with:
- https://huggingface.co/Wan-AI/Wan2.2-TI2V-5B/tree/main
- https://huggingface.co/Wan-AI/Wan2.2-T2V-A14B/tree/main
- https://huggingface.co/Wan-AI/Wan2.2-I2V-A14B/tree/main

Wauplin (commented):
Except a last comment on the download count rule, it looks good to me for a merge.

Could you also open PRs to add `library_name: wan2.2` in the model card metadata of Wan2.2 repos? So at least:
- https://huggingface.co/Wan-AI/Wan2.2-TI2V-5B
- https://huggingface.co/Wan-AI/Wan2.2-T2V-A14B
- https://huggingface.co/Wan-AI/Wan2.2-I2V-A14B

Once that's done, we should be able to release it.

kelseyee:
I've added `library_name: wan2.2` to the two 14B repos. The 5B repo doesn't need it since it's not an MoE model, and its download count is already correct. Thanks a lot!

Wauplin:
Great! I'm merging it now. Should be online in a few days.


> The 5B repo doesn't need it since it's not an MoE model, and its download count is already correct.

I still think it'd be good to tag it as `wan2.2` since this is its main library to use it, right? The download count would still work the same. Tagging it is better for discoverability and to get a direct link to the Wan2.2 repo from this model page. 

If the 5B repo is not to be run with Wan2.2 library, I think the naming of the repo should be updated (it's confusing otherwise).

This doesn't block merging this PR but would still be good to address it while you are at it.

packages/tasks/src/model-libraries.ts
```diff
@@ -1170,6 +1170,12 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		countDownloads: `path_extension:"pt"`,
 		snippets: snippets.vui,
 	},
+	"wan2.2": {
```
```suggestion
	"wan2.2": {
```

kelseyee:
> I still think it'd be good to tag it as `wan2.2` since this is its main library to use it, right? The download count would still work the same. Tagging it is better for discoverability and to get a direct link to the Wan2.2 repo from this model page.

Good point! I’ve gone ahead and added library_name: wan2.2 to the 5B repo too. Better for visibility and keeps everything aligned. Appreciate the feedback!

kelseyee:
@Wauplin Hi, it looks like the download count is still not working for the two 14B repos. Is this because the commit hasn't been released to production yet?

Wauplin:
It has been released only yesterday! Sorry about the delay. You can see it here:

image


For the download count, it needs to be processed and should appear in the next hours/days

PR #1669, opened 2025-08-04
Fix the issue that `sdk` parameter is not used.

viocha:


packages/hub/src/lib/create-repo.ts
```diff
@@ -47,7 +47,7 @@ export async function createRepo(
 			...(repoId.type === "space"
 				? {
 						type: "space",
-						sdk: "static",
+						sdk: sdk || "static",
```
Kakulukian:
```suggestion
						sdk: sdk ?? "static",
```

Style preferred

packages/hub/src/lib/create-repo.ts
```diff
@@ -47,7 +47,7 @@ export async function createRepo(
 			...(repoId.type === "space"
 				? {
 						type: "space",
-						sdk: "static",
+						sdk: sdk ?? "static",
```
coyotte508:
```suggestion
						sdk: params.sdk ?? "static",
```

PR #1673, opened 2025-08-06
[safetensors] parameters count based on quantization config

mishig25:
Enhance safetensors metadata parsing to support new data types and quantization configurations

- Added support for sub-byte data types: F4, F6_E2M3, F6_E3M2, E8M0, FP4, and UE8.
- Implemented fetching and parsing of model configuration for quantization information.
- Updated parameter counting functions to consider quantization effects on tensor parameters.
- Added tests to validate the handling of new data types and parameter counting for large models.


SunMarc (commented):
Wow, thanks for taking into account quantization for parameters count !

PR #1677, opened 2025-08-07
Add MXFP4 GGUF QuantizationType

CISC:
Added in GPT-OSS PR ggml-org/llama.cpp#15091


ngxson (approved):

CISC:
Hmmm, I think this test is wrong, it check two different things against eachother:
https://github.com/huggingface/huggingface.js/actions/runs/16801643530/job/47613401870?pr=1677#step:7:95

CISC:
It checks `GGMLQuantizationType` against `GGUF_QUANT_ORDER` which is `GGMLFileQuantizationType`, they are not equivalent:
https://github.com/huggingface/huggingface.js/blob/d67a464473ca07fee9811a129e5fac8cc7487098/packages/tasks/src/gguf.ts#L66

ngxson:
The CI seems to be broken, but this is quite a trivial change so I guess it's ok to merge without CI being green

PR #1687, opened 2025-08-14
Add BBoxMaskPose to download count

MiraPurkrabek:
Add BBoxMaskPose paper to download count.

HuggingFace link - [BBoxMaskPose on HuggingFace](https://huggingface.co/vrg-prague/BBoxMaskPose)

GitHub link - [BBoxMaskPose on GitHub](https://github.com/MiraPurkrabek/BBoxMaskPose)

### DIFF
--- a/packages/tasks/src/model-libraries.ts
+++ b/packages/tasks/src/model-libraries.ts
@@ -130,6 +130,13 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		filter: false,
 		countDownloads: `path:"llm_config.json"`,
 	},
+	bboxmaskpose: {
+		prettyLabel: "BBoxMaskPose",
+		repoName: "BBoxMaskPose",
+		repoUrl: "https://github.com/MiraPurkrabek/BBoxMaskPose",
+		filter: false,
+		countDownloads: `path_extension:"pth"`,
+	},
 	ben2: {
 		prettyLabel: "BEN2",
 		repoName: "BEN2",


packages/tasks/src/model-libraries.ts
```diff
@@ -130,6 +130,13 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		filter: false,
 		countDownloads: `path:"llm_config.json"`,
 	},
+	bboxmaskpose: {
```
Vaibhavs10, line 133: note: you'd need to add this tag to the model README as well: https://huggingface.co/vrg-prague/BBoxMaskPose

packages/tasks/src/model-libraries.ts
```diff
@@ -130,6 +130,13 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
...
+		prettyLabel: "BBoxMaskPose",
+		repoName: "BBoxMaskPose",
+		repoUrl: "https://github.com/MiraPurkrabek/BBoxMaskPose",
+		filter: false,
+		countDownloads: `path_extension:"pth"`,
```
Vaibhavs10, line 138: Are all the three models used for inference or is one used at a time?

The reason I ask is, there are currently three different model checkpoint in the repo: https://huggingface.co/vrg-prague/BBoxMaskPose/tree/main

and the download counter will +=1 download counter for each hit, i.e. if all three are used for single inference, it'd result in +=3 which would make downloads inaccurate.

Vaibhavs10 (commented):
Thanks for the PR. left 2 small comments, let me know if they are not clear! 🤗

packages/tasks/src/model-libraries.ts
MiraPurkrabek: Done

packages/tasks/src/model-libraries.ts
MiraPurkrabek: Thank you for the review, @Vaibhavs10 .

Yes, all 3 models are needed to run the whole method -- it consists of their iterative usage and smart integration. However, all models could also work as standalone models if the user does not want to use them in the loop.

Therefore, it make sense to me to have all models in the same repo as they are connected to the same paper and together form the main method.

On the other hand, it also make sense to count downloads for each model. A user can download only one of the models and it should be counted. 

Let me know if you see it differently or if I should change anything.

Thank you

Vaibhavs10 (approved):
Thanks for iterating here. Let's wait for one more approval before merge.

PR #1695, opened 2025-08-20
Local apps: Add Paddler

mcharytoniuk:
This PR adds Paddler as a local app.

Link to the project: https://paddler.intentee.com/
GitHub repository: https://github.com/intentee/paddler

Paddler allows to run servers in a cluster, so it is really more than a local runner (it is intended for scalable deployments rather than desktop use), but it is still local, so I think it fits the category (unless you think it can be put somewhere else also?).

You can also check out the [video overview](https://www.youtube.com/watch?v=aT6QCL8lk08).

cc @Vaibhavs10

### DIFF
--- a/packages/tasks/src/local-apps.ts
+++ b/packages/tasks/src/local-apps.ts
@@ -340,6 +340,20 @@ const snippetLemonade = (model: ModelData, filepath?: string): LocalAppSnippet[]
 	];
 };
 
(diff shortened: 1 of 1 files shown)


malzag:
Hello, I'm one of the co-maintainers of Paddler. I was wondering if it would be possible to review this PR. Of course, if you have any questions or suggestions, please let us know!

PR #1719, opened 2025-09-02
:bug: fix `Use this model snippets` for PaddleOCR models

TingquanGao:
vlm models don't support save_to_img


Wauplin (approved):
Thanks for the fix @TingquanGao :)

PR #1730, opened 2025-09-05
WIP: upload progress bar on CLI

assafvayner:
Adds progress bars to file uploads.

Current issue/todo, xet file upload only emits progress values at 0 and 100% completion, this is not very helpful, no progress is shown.

Screenshot 2025-09-04 at 5 15 29 PM

### DIFF
--- a/packages/hub/cli.ts
+++ b/packages/hub/cli.ts
@@ -8,6 +8,124 @@ import { stat } from "node:fs/promises";
 import { basename, join } from "node:path";
 import { HUB_URL } from "./src/consts";
 import { version } from "./package.json";
+import type { CommitProgressEvent } from "./src/lib/commit";
+import type { MultiBar, SingleBar } from "cli-progress";
+
+// Import types for cli-progress (available at compile time even with dynamic imports)
+// type MultiBar = import("cli-progress").MultiBar;
+// type SingleBar = import("cli-progress").SingleBar;
+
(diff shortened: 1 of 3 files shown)


packages/hub/cli.ts
```diff
@@ -8,6 +8,115 @@ import { stat } from "node:fs/promises";
...
+	private truncateFilename(filename: string, maxLength: number): string {
+		if (filename.length <= maxLength) return filename;
+		return "..." + filename.slice(-(maxLength - 3));
+	}
+
+	stop(): void {
+		if (!this.isQuiet && this.cliProgressAvailable && this.multibar) {
+			this.multibar.stop();
+		}
+	}
```
coyotte508, line 116: you can also add:

```ts
[Symbol.dispose](): void { this.stop();}
```

then you can do `using progressManager = ...` and avoid try/catch on it

coyotte508:
let me know when this is ready for review :)

packages/hub/cli.ts
assafvayner: I tried that but for some reason it caused the progress bar to never finish and the cli was hanging after committing (possible but unlikely caused by something else)

coyotte508: ok feel free to revert

coyotte508:
nice, thank you!

PR #1746, opened 2025-09-12
Enhance token generation by including isPullRequest flag in xetWriteToken and related functions for better handling of pull requests

mishig25:
I was getting error when trying to create pull request on repos that I do **not** own:
```
Error: Forbidden: pass `create_pr=1` as a query parameter to create a Pull Request. URL: https://huggingface.co/api/models/reach-vb/TinyLlama-1.1B-Chat-v1.0-q4_k_m-GGUF/xet-write-token/main
```

I was using new API from https://github.com/huggingface/huggingface.js/pull/1718

```ts
// Edit first kB of file
await commit({
  repo,
  accessToken: "hf_...",
  operations: [{
    type: "edit",
    originalContent: originalFile,
    edits: [{
      start: 0,
      end: 1000,
      content: new Blob(["blablabla"])
    }]
  }]
})
```

Let me know if this PR is the right way to handle this issue

### DIFF
--- a/packages/hub/src/lib/commit.ts
+++ b/packages/hub/src/lib/commit.ts
@@ -410,6 +410,7 @@ export async function* commitIter(params: CommitParams): AsyncGenerator<CommitPr
 								repo: repoId,
 								// todo: maybe leave empty if PR?
 								rev: params.branch ?? "main",
+								isPullRequest: params.isPullRequest,
 								yieldCallback: (event) => yieldCallback({ ...event, state: "uploading" }),
 							})) {
 								if (event.event === "file") {
--- a/packages/hub/src/utils/createXorbs.ts
+++ b/packages/hub/src/utils/createXorbs.ts
@@ -199,7 +199,7 @@ export async function* createXorbs(
 
 					let cacheData = chunkCache.getChunk(chunk.hash, chunkModule.compute_hmac);
 					if (cacheData === undefined && chunk.dedup && bytesSinceRemoteDedup >= INTERVAL_BETWEEN_REMOTE_DEDUP) {
-						const token = await xetWriteToken(params);
+						const token = await xetWriteToken({ ...params, isPullRequest: params.isPullRequest });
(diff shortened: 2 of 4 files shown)


mishig25:
updated the description

coyotte508:
this should be the right way @mishig25

coyotte508:
hmm maybe this should be taken into account in https://github.com/huggingface-internal/moon-landing/pull/14572 too

seanses:
> hmm maybe this should be taken into account in [huggingface-internal/moon-landing#14572](https://github.com/huggingface-internal/moon-landing/pull/14572) too

Probably not.. To `git push` LFS files to a repo clients first need to communicate with the LFS Batch API, and that requires write access.

coyotte508:
> > hmm maybe this should be taken into account in [huggingface-internal/moon-landing#14572](https://github.com/huggingface-internal/moon-landing/pull/14572) too
> 
> Probably not.. To `git push` LFS files to a repo clients first need to communicate with the LFS Batch API, and that requires write access.

@seanses There are multiple cases in moon-landing's current behavior: 

- If discussions are disabled for the repo, the user needs write access to a repo
- If a ref is specified in the LFS protocol, the user needs write access to that ref (eg if it's a ref they opened a PR with)
- If no ref is specified in the LFS protocol (eg opening a PR) the user need read access to the repo only

The JS lib has this snippet in their LFS call:

https://github.com/huggingface/huggingface.js/blob/bc98deb199e00a450162546c91f1fb91fd9ff4a4/packages/hub/src/lib/commit.ts#L330-L334

My guess if the user only has read access, eg they don't specify a ref, then the xet-write-token url the git-xet client needs to use to refresh the token needs to have a `?create_pr=1` query param. If they specify a ref, we should forward the ref to the xet-write-token url given to the client.

(will re-review https://github.com/huggingface-internal/moon-landing/pull/14572 monday)

cc @Wauplin

seanses:
@coyotte508 I see. Thanks for explaining this!

PR #1753, opened 2025-09-17
Fix MXFP4 quant label

kefranabg:
I noticed the MXFP4 quant type was not parsed properly for GGUF models on the hub ([see here for instance](https://huggingface.co/lmstudio-community/gpt-oss-120b-GGUF)). 

I believe the name format "MXFP4_MOE" is incorrect, it should have been referred to as "MXFP4", and since were using the enum key to parse the quant label in the file name, that's why it fails. cc @ngxson

### DIFF
--- a/packages/tasks/src/gguf.ts
+++ b/packages/tasks/src/gguf.ts
@@ -40,7 +40,7 @@ export enum GGMLFileQuantizationType {
 	Q4_0_8_8 = 35,
 	TQ1_0 = 36,
 	TQ2_0 = 37,
-	MXFP4_MOE = 38,
+	MXFP4 = 38,
 
 	// custom quants used by unsloth
 	// they are not officially a scheme enum value in GGUF, but only here for naming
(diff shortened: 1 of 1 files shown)


pcuenca:
I think there was some discussion about needing both, cc @ngxson maybe

julien-c:
@mishig25 for viz too just in case

Vaibhavs10 (commented):
AFAIR - `MXFP4_MOE` is intentional: https://huggingface.co/ggml-org/gpt-oss-20b-GGUF/tree/main?show_file_info=gpt-oss-20b-mxfp4.gguf (look at the official quants)

ngxson:
I think we should not change this list of enum as it's kept in sync with the official definition from GGML (which use `MXFP4_MOE`)

I would suggest adding an exception in `parseGGUFQuantLabel`, which will resolve `-mxfp4` file name to `MXFP4_MOE` enum type

kefranabg:
> I would suggest adding an exception in parseGGUFQuantLabel, which will resolve -mxfp4 file name to MXFP4_MOE enum type

Yes probably the best option then 👍 Will open a new PR for this next week

PR #1767, opened 2025-09-25
Debug mode for xet upload

coyotte508:
cc @mishig25 @assafvayner 

This script will save all downloaded dedup shards, outputted xorbs & shards when uploading a file

It also has a "replay" mode to reuse those

Should help debug xet uploads


coyotte508:
merging so the debug script is on main

PR #1770, opened 2025-09-25
[safetensors] Fix GPTQ/AWQ quantized model parameter counting

mishig25:
### Fix GPTQ/AWQ quantized model parameter counting

Fixes parameter count calculation for GPTQ/AWQ quantized models by:
Applying 8x multiplier to qweight tensors based on quantization bits (32/4 = 8 for 4-bit)
Skipping auxiliary tensors (qzeros, g_idx, scales) from parameter count
Defaulting quantized tensor detection when no exclusion list is provided

Before: [RedHatAI/Meta-Llama-3.1-8B-Instruct-quantized.w4a16](https://huggingface.co/RedHatAI/Meta-Llama-3.1-8B-Instruct-quantized.w4a16) reported ~2B parameters
After: Correctly reports ~8B parameters

### DIFF
--- a/packages/hub/src/lib/parse-safetensors-metadata.spec.ts
+++ b/packages/hub/src/lib/parse-safetensors-metadata.spec.ts
@@ -207,6 +207,31 @@ describe("parseSafetensorsMetadata", () => {
 		assert.strictEqual(parameterCount.E8M0, 24);
 	});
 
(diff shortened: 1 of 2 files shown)


mishig25:
follow up to https://github.com/huggingface/huggingface.js/pull/1673

julien-c:
did you agree with this implem @ngxson?

PR #1771, opened 2025-09-26
xet upload: fix chunk cache desync

coyotte508:


packages/hub/src/utils/ChunkCache.ts
```diff
@@ -12,6 +12,16 @@ export class ChunkCache {
 	hmacs = new Set<string>(); // todo : remove old hmacs
 
 	addChunkToCache(hash: string, xorbIndex: number, chunkIndex: number, hmac: string | null): void {
+		if (this.map.has(hash)) {
+			// Happens when we receive an existing chunk from remote dedup info (eg duplicate chunk in shard? Or shards with same hmac key
+			// sharing chunks/xorbs)
+
+			// processing this chunk again would desync the cache, as `this.map.size` would not increase, as opposed to `this.index`
+
+			// Ideally we'd still process it to evict it later ("refresh it") but would need more complex handling, or stop using
```
coyotte508, line 21: Note, maybe a not-to-complex solution:

- remove and readd the key in the map (to put the key at the end)
- in the eviction code, instead of checking on `map.size > MAX_SIZE`, check on `map.size > 1 && firstValue === this.index`  (or something approaching)

Would need unit tests and careful thinking

PR #1788, opened 2025-10-07
Update `image-to-image` target size param docstring

hanouticelina:
Related to https://github.com/huggingface/huggingface_hub/pull/3399, some providers or models may ignore this parameter if it's not supported. This PR updates the docstring to make that clearer and align it with the one in `huggingface_hub`.

### DIFF
--- a/packages/tasks/src/tasks/image-to-image/inference.ts
+++ b/packages/tasks/src/tasks/image-to-image/inference.ts
@@ -41,15 +41,15 @@ export interface ImageToImageParameters {
 	 */
 	prompt?: string;
 	/**
-	 * The size in pixel of the output image. This parameter is only supported by some providers
-	 * and for specific models. It will be ignored when unsupported.
+	 * The size in pixels of the output image. This parameter is only supported by some
+	 * providers and for specific models. It will be ignored when unsupported.
 	 */
 	target_size?: TargetSize;
 	[property: string]: unknown;
 }
 /**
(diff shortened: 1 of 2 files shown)


pcuenca (approved):
Thank you! 🙌 Is there a way to know the providers / models that support it?

packages/tasks/src/tasks/image-to-image/inference.ts
```diff
@@ -41,13 +41,15 @@ export interface ImageToImageParameters {
 	 */
 	prompt?: string;
 	/**
-	 * The size in pixel of the output image.
+	 * The size in pixel of the output image. This parameter is only supported by some providers
```
pcuenca:
```suggestion
	 * The size in pixels of the output image. This parameter is only supported by some providers
```

maybe (not sure)? Same for the others. if you decide to update.

hanouticelina:
> Thank you! 🙌 Is there a way to know the providers / models that support it? 

@pcuenca no there is no automatic way to check that unfortunately - someone has to look it up in each provider's documentation. Also for a given provider, some models might support it while others don't

packages/tasks/src/tasks/image-to-image/inference.ts
hanouticelina: yes, addressed in [c3bfab6](https://github.com/huggingface/huggingface.js/pull/1788/commits/c3bfab60c293068d5e33e354240301e09865124e)

PR #1791, opened 2025-10-11
📈 Download tracking for new diff-interpretation-tuning project

ed1d1a8d:
Registering a new library per the instructions on https://huggingface.co/docs/hub/models-adding-libraries#register-your-library and https://huggingface.co/docs/hub/models-download-stats.

This library is used by https://huggingface.co/diff-interpretation-tuning/loras model repo, and we are adding it so that downloads can be accurately counted. Note that every .pt file contains one or more adapter files, hence we track the total number of .pt file downloads.

### DIFF
--- a/packages/tasks/src/model-libraries.ts
+++ b/packages/tasks/src/model-libraries.ts
@@ -289,7 +289,7 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 	},
 	"diff-interpretation-tuning": {
 		prettyLabel: "Diff Interpretation Tuning",
-		repoName: "diff-interpretation-tuning",
+		repoName: "Diff Interpretation Tuning",
 		repoUrl: "https://github.com/Aviously/diff-interpretation-tuning",
 		filter: false,
 		countDownloads: `path_extension:"pt"`,


packages/tasks/src/model-libraries.ts
```diff
@@ -287,6 +287,13 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		snippets: snippets.dia,
 		filter: false,
 	},
+	"diff-interpretation-tuning": {
+		prettyLabel: "Diff Interpretation Tuning",
+		repoName: "diff-interpretation-tuning",
```
```suggestion
		repoName: "Diff Interpretation Tuning",
```

Wauplin:
Merged it :) expect to see it live on the Hub in the next days

ed1d1a8d:
@Wauplin Thanks for the help here. I actually wanted to follow up. I think the download counts on the repository (https://huggingface.co/diff-interpretation-tuning/loras) may be broken:
image

Do you know what is going on?

Wauplin:
We can't know for sure but likely someone / a CI that pulled all loras at once (and potentially multiple times). Unfortunately there is little we can do about it

PR #1796, opened 2025-10-13
ollama-utils: do not PR if only changes are comments

coyotte508:


ngxson:
We could also just disable these comments altogether. Though I prefer to leave the comments and the PR creation as-is, so it can acts as a weekly reminder

coyotte508:
> Though I prefer to leave the comments and the PR creation as-is, so it can acts as a weekly reminder

ok

PR #1798, opened 2025-10-13
add pylate code snippet

merveenoyan:
pylate is wrapping sentence transformers, but the auto snippets showing up on pylate models actually don't work (and won't work) because sentence transformers doesn't have maxsim, so I updated the snippet. @tomaarsen can you kindly review? 🙏🏻🥹

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,7 +98,7 @@ You can run our packages with vanilla JS, without any bundler, by using a CDN or
 
 ```html
 <script type="module">
-    import { InferenceClient } from 'https://cdn.jsdelivr.net/npm/@huggingface/inference@4.11.1/+esm';
+    import { InferenceClient } from 'https://cdn.jsdelivr.net/npm/@huggingface/inference@4.11.2/+esm';
     import { createRepo, commit, deleteRepo, listFiles } from "https://cdn.jsdelivr.net/npm/@huggingface/hub@2.6.12/+esm";
 </script>
 ```
--- a/packages/inference/README.md
+++ b/packages/inference/README.md
@@ -67,6 +67,7 @@ Currently, we support the following providers:
 - [Cohere](https://cohere.com)
 - [Cerebras](https://cerebras.ai/)
 - [Groq](https://groq.com)
+- [Wavespeed.ai](https://wavespeed.ai/)
 - [Z.ai](https://z.ai/)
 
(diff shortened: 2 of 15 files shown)


packages/tasks/src/model-libraries-snippets.ts
```diff
@@ -1364,6 +1364,49 @@ function get_widget_examples_from_st_model(model: ModelData): string[] | undefin
...
+	documents,
+	is_query=False,
+)
+
+reranked_documents = rank.rerank(
+	documents_ids=documents_ids,
+	queries_embeddings=queries_embeddings,
+	documents_embeddings=documents_embeddings,
+)
+print(reranked_documents)`,
```
tomaarsen: @NohTow (the PyLate maintainer) mentioned eventually wanting to update this structure, so I think we might want to wait for that to be completed?

merveenoyan: sounds good!

NohTow: Hello!
The snippet you provided @merveenoyan should be fine!
I actually wanted to update the snippet a long time ago but did not do it in the end because I thought that exposing a similarity function that can be called directly (without reranking) would be maybe simpler for newcomers, but reranking is fine if we want to merge right now, maybe we can update this once I make the modifications to PyLate
(We would just need to update the similarity function to handle the padding internally to not ask the user to do that)

packages/tasks/src/model-libraries-snippets.ts
```diff
@@ -1364,6 +1364,49 @@ function get_widget_examples_from_st_model(model: ModelData): string[] | undefin
 
 export const sentenceTransformers = (model: ModelData): string[] => {
 	const remote_code_snippet = model.tags.includes(TAG_CUSTOM_CODE) ? ", trust_remote_code=True" : "";
+	
+	if (model.tags.includes("PyLate")) {
```
Vaibhavs10, line 1368: This is a bit too long for a snippet - generally we push for concise and complete snippets.

NohTow: makes sense
though note that even if we change the reranking to a simple similarity function it'll mostly only removed the document_ids part
the code will be very similar to the one of ST
were you referring to docs/queries being long?

merveenoyan:
@Vaibhavs10 I wanted the snippet to be equivalent of the one in sentence-transformers, I just removed reranking and made documents shorter, can you please review again?

packages/tasks/src/model-libraries-snippets.ts
merveenoyan: I have removed reranking completely, it's up to user to do what they want with query and document embeddings

PR #1799, opened 2025-10-13
hub: spaces reloading client

cbensimon:
Meant to be (mostly) used by hub frontend


cbensimon:
Will first implement in Hub internal codebase then move to huggingface.js once stable

PR #1804, opened 2025-10-14
feat: Add DeepInfra as an inference provider

Thachnh:
### What’s in this PR

This PR adds DeepInfra as a new inference provider for conversational and text-generation tasks. The integration is via DeepInfra’s OpenAI-compatible API endpoints.

### Test

```
pnpm --filter @huggingface/inference test "test/InferenceClient.spec.ts" -t "DeepInfra"

> @huggingface/inference@4.11.2 test /home/thach/huggingface/huggingface.js/packages/inference
> vitest run --config vitest.config.mts test/InferenceClient.spec.ts -t DeepInfra


 RUN  v0.34.6 /home/thach/huggingface/huggingface.js/packages/inference

 ✓ test/InferenceClient.spec.ts (128) 1042ms
   ✓ InferenceClient (128) 1042ms
     ↓ backward compatibility (1) [skipped]
       ↓ works with old HfInference name [skipped]
     ...
     ✓ DeepInfra (3) 1041ms
       ✓ chatCompletion 988ms
       ✓ chatCompletion stream
       ✓ textGeneration 1002ms
     ...

 Test Files  1 passed (1)
      Tests  3 passed | 125 skipped (128)
   Start at  22:16:13
   Duration  3.34s (transform 1.09s, setup 23ms, collect 975ms, tests 1.04s, environment 0ms, prepare 771ms)
```

### What’s next

Follow-up PRs will expand DeepInfra support to additional tasks such as text-to-image and text-to-audio.

### DIFF
--- a/packages/inference/src/providers/deepinfra.ts
+++ b/packages/inference/src/providers/deepinfra.ts
@@ -68,7 +68,7 @@ export class DeepInfraTextGenerationTask extends BaseTextGenerationTask {
 				? {
 						max_tokens: parameters.max_new_tokens,
 						...omit(parameters, ["max_new_tokens"]),
-				  }
+					}
 				: undefined),
 		};
 		return res;
@@ -88,7 +88,7 @@ export class DeepInfraTextGenerationTask extends BaseTextGenerationTask {
 		}
 
 		throw new InferenceClientProviderOutputError(
-			"Received malformed response from DeepInfra text-generation API: expected OpenAI completion payload"
+			"Received malformed response from DeepInfra text-generation API: expected OpenAI completion payload",
 		);
 	}
 }
--- a/packages/inference/test/InferenceClient.spec.ts
+++ b/packages/inference/test/InferenceClient.spec.ts
@@ -1970,9 +1970,7 @@ describe.skip("InferenceClient", () => {
 	describe.concurrent(
 		"DeepInfra",
 		() => {
-			const client = new InferenceClient(
-				env.DEEPINFRA_API_KEY ?? env.HF_DEEPINFRA_KEY ?? "dummy"
-			);
+			const client = new InferenceClient(env.DEEPINFRA_API_KEY ?? env.HF_DEEPINFRA_KEY ?? "dummy");
 
 			const HF_MODEL = "google/gemma-3-4b-it";
 			const PROVIDER_ID = "google/gemma-3-4b-it";
@@ -2041,7 +2039,7 @@ describe.skip("InferenceClient", () => {
 				expect(generation.generated_text?.length ?? 0).toBeGreaterThan(0);
 			});
 		},
-		TIMEOUT
+		TIMEOUT,
 	);
 	describe.concurrent(
 		"Cerebras",


packages/inference/src/providers/deepinfra.ts
```diff
@@ -0,0 +1,78 @@
+import type { TextGenerationOutput } from "@huggingface/tasks";
```
hanouticelina, line 22:
```suggestion
/**
 * See the registered mapping of HF model ID => DeepInfra model ID here:
 *
 * https://huggingface.co/api/partners/deepinfra/models
 *
 * This is a publicly available mapping.
 *
 * If you want to try to run inference for a new model locally before
 it's registered on huggingface.co,
* you can add it to the dictionary "HARDCODED_MODEL_ID_MAPPING" in
consts.ts, for dev purposes.
*
* - If you work at DeepInfra and want to update this mapping, please
use the model mapping API we provide on huggingface.co
* - If you're a community member and want to add a new supported HF
model to DeepInfra, please open an issue on the present repo
* and we will tag DeepInfra team members.
*
* Thanks!
*/

import type { TextGenerationOutput } from "@huggingface/tasks";
```

packages/inference/src/providers/deepinfra.ts
```diff
@@ -0,0 +1,78 @@
...
+
+interface DeepInfraCompletionChoice {
+	text?: string;
+	message?: { content?: string };
+}
+
+interface DeepInfraCompletionResponse {
+	choices: DeepInfraCompletionChoice[];
+	model: string;
+}
```
hanouticelina, line 40: Since DeepInfra uses the standard OpenAI `/completions` endpoint, it returns `text`, not `message`. The message field is only used by the `/chat/completions` endpoint. https://platform.openai.com/docs/guides/completions

```suggestion
interface DeepInfraCompletionOutput {
        choices: Array<{
                text: string;
                finish_reason: string;
                index: number;
        }>;
        model: string;
  }
```

packages/inference/src/providers/deepinfra.ts
```diff
@@ -0,0 +1,78 @@
...
+			...omit(params.args, ["inputs", "parameters"]),
+			...(parameters
+				? {
+						max_tokens: parameters.max_new_tokens,
+						stop: parameters.stop_strings,
+						...omit(parameters, ["max_new_tokens", "stop_strings"]),
+				  }
+				: undefined),
+		};
+		return res;
```
hanouticelina, line 74: the `stop_strings` -> `stop` mapping is not needed since `omit` passes through remaining params unchanged.

```suggestion
		return {
                model: params.model,
                ...omit(params.args, ["inputs", "parameters"]),
                ...(params.args.parameters
                        ? {
                                        max_tokens: parameters.max_new_tokens,
                                        ...omit(parameters, "max_new_tokens"),
                                }
                        : undefined),
                prompt: params.args.inputs,
        };
```
```

File: packages/inference/src/providers/deepinfra.ts [hunk 4]
```diff
@@ -0,0 +1,78 @@
...
+			response.choices.length > 0
+		) {
+			const choice = response.choices[0];
+			const completion =
+				choice.text ??
+				(typeof choice.message?.content === "string" ? choice.message.content : undefined);
+			if (typeof completion === "string") {
+				return { generated_text: completion };
+			}
+		}
```
hanouticelina, line 88: no need to fall back to `message?.content` no? since DeepInfra uses standard OpenAI format

```suggestion
        if (
                typeof response === "object" &&
                "choices" in response &&
                Array.isArray(response.choices) &&
                response.choices.length > 0 &&
                typeof response.model === "string"
        ) {
                const completion = response.choices[0];
                return {
                        generated_text: completion.text,
                };
        }
```

hanouticelina (commented):
Thank you! I left a couple of comments but it should be quite easy to address :)

packages/inference/src/providers/deepinfra.ts
Thachnh: @hanouticelina the https://huggingface.co/api/partners/deepinfra/models endpoint returns error. WHat can we do here?

packages/inference/src/providers/deepinfra.ts
Thachnh: `message` field has been removed.

packages/inference/src/providers/deepinfra.ts
Thachnh: the stop_strings -> stop mapping  has been removed

packages/inference/src/providers/deepinfra.ts
Thachnh: since message field has been removed, we don't need to do any fallback. Use `choice.text` directly.

packages/inference/src/providers/deepinfra.ts
hanouticelina: don't worry about it, it will work when we register deepinfra server-side. it's not a blocker for this PR, we can apply the suggestion above

Thachnh:
> Thanks @Thachnh for the iteration! let's add the header mentioned in this [comment](https://github.com/huggingface/huggingface.js/pull/1804/changes#r2768444882) then we will be good to merge 🙌

@hanouticelina The suggestion has been applied. Please help us merge the PR. Thanks.

packages/inference/src/types.ts
hanouticelina, line 29: @Thachnh can you add `deepinfra: "DeepInfra"` to `PROVIDERS_HUB_ORGS` please? 🙏

Thachnh: @hanouticelina done. Thanks

PR #1806, opened 2025-10-15
Provider => hub mapping

julien-c:


Wauplin (approved):
let's use it in docs as well (I like to have single source of truth^^)

julien-c:
yes 100% but i'll let that part to you as i'm not too familiar with the hub-docs scripts

PR #1808, opened 2025-10-16
Update memory configurations for MI50 and Radeon Pro VII

bjoernh:
There are also 32 GB Variants of this GPUs


Vaibhavs10 (approved):
Thanks! 🤗

PR #1813, opened 2025-10-20
[Inference] Add Text-Generation support for Replicate provider

lucataco:
This PR adds support for the `Text Generation` task type for Replicate models.

Example:

- [huggingface.co/akhaliq/gpt-5](https://huggingface.co/akhaliq/gpt-5)
- [replicate.com/openai/gpt-5](https://replicate.com/openai/gpt-5)

### DIFF
--- a/packages/inference/src/lib/getProviderHelper.ts
+++ b/packages/inference/src/lib/getProviderHelper.ts
@@ -153,13 +153,13 @@ export const PROVIDERS: Record<InferenceProvider, Partial<Record<InferenceTask, 
 	publicai: {
 		conversational: new PublicAI.PublicAIConversationalTask(),
 	},
-        replicate: {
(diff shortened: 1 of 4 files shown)


SBrandeis:
Hi @lucataco !
Does Replicate support OpenAI's [chat completion API](https://platform.openai.com/docs/api-reference/chat) or [Responses API](https://platform.openai.com/docs/api-reference/responses/create)?

Those are the APIs used to power the chat widget on model pages and our chat API at router.huggingface.co for conversational models (such as GPT-5).

If you support such APIs, adding support for them in the JS client should be very easy, see this for example:
https://github.com/huggingface/huggingface.js/blob/e80852fbf05cb61d018586dd6516eb13c4e51067/packages/inference/src/providers/groq.ts#L32-L41

PR #1816, opened 2025-10-20
[Inference] Add async polling support for Replicate provider

lucataco:
This PR adds async polling for Replicate
To help the huggingface team so that they no longer need an allow-list of Replicate models to show Hub users 

Screenshot 2025-10-20 at 2 33 56 PM

@SBrandeis

### DIFF
--- a/packages/inference/src/providers/replicate.ts
+++ b/packages/inference/src/providers/replicate.ts
@@ -19,7 +19,7 @@ import { isUrl } from "../lib/isUrl.js";
 import type { BodyParams, HeaderParams, RequestArgs, UrlParams } from "../types.js";
 import { omit } from "../utils/omit.js";
 import {
-	TaskProviderHelper,
+        TaskProviderHelper,
 	type AutomaticSpeechRecognitionTaskHelper,
 	type ImageToImageTaskHelper,
 	type TextToImageTaskHelper,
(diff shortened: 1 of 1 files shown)


packages/inference/src/providers/replicate.ts
```diff
@@ -69,6 +114,97 @@ abstract class ReplicateTask extends TaskProviderHelper {
...
+			if (!pollResponse.ok) {
+				throw new InferenceClientProviderOutputError(
+					`Failed to poll Replicate prediction status: HTTP ${pollResponse.status}`
+				);
+			}
+
+			const prediction = (await pollResponse.json()) as ReplicateAsyncResponse;
+			const predictionStatus = prediction.status;
+
+			if (!predictionStatus || predictionStatus === "succeeded") {
```
hanouticelina, line 173: is it intentional to treat "no status" as a successful response?

packages/inference/src/providers/replicate.ts
hanouticelina, line 1: I think we have to remove `Prefer: wait` from the headers (defined in `prepareHeaders`) now that all the tasks are using async polling

packages/inference/src/providers/replicate.ts
```diff
@@ -33,6 +33,51 @@ export interface ReplicateOutput {
 	output?: string | string[];
 }
 
+type ReplicatePredictionStatus = "starting" | "processing" | "succeeded" | "failed" | "canceled" | "queued";
```
hanouticelina, line 36: i believe this gives tighter typings and better autocomplete support
```suggestion
const REPLICATE_STATUSES = [
  "starting",
  "processing",
  "succeeded",
  "failed",
  "canceled",
  "queued",
] as const;

type ReplicatePredictionStatus = (typeof REPLICATE_STATUSES)[number];
```

hanouticelina (commented):
thanks @lucataco for the PR! I left some comments

PR #1817, opened 2025-10-21
Make safetensor shard filename parsing more flexible

kefranabg:
Context:

A lot of deepseek models have unconventional number of digits for the total shard number in their filename, which leads to a filename parsing fail. It breaks the file info navigation on the hub:

https://github.com/user-attachments/assets/75676c00-387c-4a92-808d-0d84caba7e51

[here](https://huggingface.co/deepseek-ai/DeepSeek-V3.2-Exp?show_file_info=model-00001-of-000163.safetensors) for instance, `model-00001-of-000163.safetensors` should be `model-00001-of-00163.safetensors` (there's 1 extra "0" in "model-00001-of-.safetensors")

Most of deepseek models with a lot of shards have this issue, since they are important models, I suggest we make the regex more flexible to accept a variable number of digits for shards in the filename.

Would that be acceptable?

### DIFF
--- a/packages/hub/src/lib/parse-safetensors-metadata.ts
+++ b/packages/hub/src/lib/parse-safetensors-metadata.ts
@@ -309,10 +309,10 @@ export async function parseSafetensorsMetadata(
 								? typeof header.__metadata__.total_parameters === "number"
 									? header.__metadata__.total_parameters
 									: typeof header.__metadata__.total_parameters === "string"
-										? parseInt(header.__metadata__.total_parameters)
-										: undefined
+									  ? parseInt(header.__metadata__.total_parameters)
+									  : undefined
 								: undefined,
-					}
+				  }
 				: undefined),
 		};
 	} else if (
@@ -336,10 +336,10 @@ export async function parseSafetensorsMetadata(
 								? typeof index.metadata.total_parameters === "number"
 									? index.metadata.total_parameters
 									: typeof index.metadata.total_parameters === "string"
-										? parseInt(index.metadata.total_parameters)
-										: undefined
+									  ? parseInt(index.metadata.total_parameters)
+									  : undefined
 								: undefined,
-					}
+				  }
 				: undefined),
 		};
 	} else {


packages/hub/src/lib/parse-safetensors-metadata.ts
```diff
@@ -14,7 +14,7 @@ export const SAFETENSORS_INDEX_FILE = "model.safetensors.index.json";
 export const RE_SAFETENSORS_FILE = /\.safetensors$/;
 export const RE_SAFETENSORS_INDEX_FILE = /\.safetensors\.index\.json$/;
 export const RE_SAFETENSORS_SHARD_FILE =
-	/^(?<prefix>(?<basePrefix>.*?)[_-])(?<shard>\d{5})-of-(?<total>\d{5})\.safetensors$/;
+	/^(?<prefix>(?<basePrefix>.*?)[_-])(?<shard>\d+)-of-(?<total>\d+)\.safetensors$/;
```
julien-c: at least enforce the same number of digits 🙏  (not sure how to do this in the regex, maybe it's an OR or the different digits for 5 and 6 digits)

coyotte508: or we just don't care about the 5X and just allow regular numbers to avoid this issue again in the future?

julien-c: i would keep some convention

kefranabg: In that case maybe it's better to handle this edge case directly on hub side and fix the shard digit numbers before passing it to `parseSafetensorsShardFilename` no? 

Like transforming `000163` to `00163` before parsing it

julien-c: hmm no i think here is fine (we use this hub-side anyways)

kefranabg: Done here: https://github.com/huggingface/huggingface.js/pull/1817/commits/df147f8ae1a49c0058d4c7704f1833fbebb8b5a4

Supporting both 5 OR 6 digits for shards

coyotte508 (approved):

packages/hub/src/lib/parse-safetensors-metadata.ts
```diff
@@ -14,7 +14,7 @@ export const SAFETENSORS_INDEX_FILE = "model.safetensors.index.json";
 export const RE_SAFETENSORS_FILE = /\.safetensors$/;
 export const RE_SAFETENSORS_INDEX_FILE = /\.safetensors\.index\.json$/;
 export const RE_SAFETENSORS_SHARD_FILE =
-	/^(?<prefix>(?<basePrefix>.*?)[_-])(?<shard>\d{5})-of-(?<total>\d{5})\.safetensors$/;
+	/^(?<prefix>(?<basePrefix>.*?)[_-])(?<shard>\d{5,6})-of-(?<total>\d{5,6})\.safetensors$/;
```
coyotte508, line 17: maybe it should be 
```suggestion
export const RE_SAFETENSORS_SHARD_FILE =
	/^(?<prefix>(?<basePrefix>.*?)[_-])(?<shard>\d+)-of-(?<total>\d+)\.safetensors$/;
```

and a function should check that both `shard` and `total` have the same number of digits

🤷

kefranabg: The thing is that for the deepseek case, shard and total don't have the same number of digits 😭

`model-00001-of-000163.safetensors` => 5 digits for shard, 6 for total

julien-c: yes ok that's fine then, approving like this

PR #1824, opened 2025-10-28
New provider SiliconFlow

jackalcooper:


jackalcooper:
Hi @julien-c . I tried to make POST request to `https://huggingface.co/api/partners/{provider}/models` as described in https://huggingface.co/docs/inference-providers/register-as-a-provider#3-model-mapping-api but got 404. Could you please confirm if our partner account need to be explicitly activated or configured first? 
- Our official org is https://huggingface.co/siliconflow 
- The org i am using to test provider registration is https://huggingface.co/siliconflow-community

PR #1826, opened 2025-10-30
Add `chronos-forecasting` as a library

abdulfatir:
Adding `chronos-forecasting` as a library as suggested by @NielsRogge here: https://github.com/amazon-science/chronos-forecasting/issues/339

### DIFF
--- a/packages/tasks/src/model-libraries.ts
+++ b/packages/tasks/src/model-libraries.ts
@@ -209,6 +209,12 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		filter: false,
 		countDownloads: `path:"asset/GPT.pt"`,
 	},
+	"chronos-forecasting": {
+		prettyLabel: "Chronos",
+		repoName: "Chronos",
+		repoUrl: "https://github.com/amazon-science/chronos-forecasting",
+		snippets: snippets.chronos_forecasting,
+	},
 	"cloud-agents": {
 		prettyLabel: "Cloud Agents",
 		repoName: "Cloud Agents",
@@ -228,12 +234,6 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		repoName: "COMET",
 		repoUrl: "https://github.com/Unbabel/COMET/",
 		countDownloads: `path:"hparams.yaml"`,
-	},
-	"chronos-forecasting": {
-		prettyLabel: "Chronos",
-		repoName: "chronos-forecasting",
-		repoUrl: "https://github.com/amazon-science/chronos-forecasting",
-		snippets: snippets.chronos_forecasting,
 	},
 	contexttab: {
 		prettyLabel: "ConTextTab",


packages/tasks/src/model-libraries.ts
```diff
@@ -229,6 +229,13 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		repoUrl: "https://github.com/Unbabel/COMET/",
 		countDownloads: `path:"hparams.yaml"`,
 	},
+	"chronos-forecasting": {
+		prettyLabel: "Chronos",
+		repoName: "chronos-forecasting",
+		repoUrl: "https://github.com/amazon-science/chronos-forecasting",
+		countDownloads: `path:"config.json"`,
```
abdulfatir: I am not sure if I can remove this since `config.json` is the default.

Wauplin: no need for it indeed! Download counter is already working :)

Wauplin: Image

Wauplin:
```suggestion
```

packages/tasks/src/model-libraries.ts
```diff
@@ -229,6 +229,13 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		repoUrl: "https://github.com/Unbabel/COMET/",
 		countDownloads: `path:"hparams.yaml"`,
 	},
+	"chronos-forecasting": {
+		prettyLabel: "Chronos",
+		repoName: "chronos-forecasting",
```
```suggestion
		prettyLabel: "Chronos Forecasting",
		repoName: "Chronos Forecasting",
```

We are in the process of merging these 2 fields. Is `Chronos Forecasting` ok for you?

packages/tasks/src/model-libraries.ts
```diff
@@ -229,6 +229,13 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		repoUrl: "https://github.com/Unbabel/COMET/",
 		countDownloads: `path:"hparams.yaml"`,
 	},
+	"chronos-forecasting": {
```
Wauplin: can you move it up in the mapping (just after `chat_tts`) to keep alphabetical order? Thanks!

packages/tasks/src/model-libraries.ts
abdulfatir: `Chronos Forecasting` works but I would slightly prefer just `Chronos`. Would that be okay?

packages/tasks/src/model-libraries.ts
abdulfatir: Done.

packages/tasks/src/model-libraries.ts
Wauplin: yes ofc!

abdulfatir:
@Wauplin what's the way to check if this change was deployed to HF?

Wauplin:
Hey @abdulfatir, it should be deployed tomorrow :)
You'll see a "Chronos" section in the "use this model" sidebar.

abdulfatir:
@Wauplin Thank you! It's now [live on HF](https://huggingface.co/amazon/chronos-2). The Chronos snippet shows up correctly but the UI still shows an option for `Transformers`. However, since transformers does not really natively support Chronos yet, the snippet is incorrect. Is there a way to disable this and only show Chronos? 

image

Wauplin:
Hi @abdulfatir, this is because the model is tagged as `transformers` in the model card. Removing the tag would remove the snippet in the "use this model" section. I opened a PR directly on the repo to do that, you just have to merge it :hugs: 
https://huggingface.co/amazon/chronos-2/discussions/2

abdulfatir:
@Wauplin Merged, thank you!

Sorry, but looks like there's another issue: the download counter has stopped since this PR's changes went live and I changed the `library_name`. My understanding is that the count changes every day but it did not change today. Should I open a PR and add the following back?

```
countDownloads: `path:"config.json"`,
```

Wauplin:
No need for it no. Let's wait for a few days to see how it evolves. Everything should be good now :)

1-day delays can happen from time to time due to failed jobs on our end. Data is recovered on the next trigger.

abdulfatir:
@Wauplin sounds good. Thank you for the clarification.

abdulfatir:
@Wauplin it works. Thank you! 🎉

Wauplin:
Great to hear that!

PR #1828, opened 2025-11-03
Feat: support 'cheapest' & 'fastest' policies for inference snippets

kefranabg:
Adding 'cheapest' & 'fastest' policies support for inference snippets

### DIFF
--- a/packages/inference/src/lib/getInferenceProviderMapping.ts
+++ b/packages/inference/src/lib/getInferenceProviderMapping.ts
@@ -179,6 +179,10 @@ export async function resolveProvider(
 		);
 		provider = "auto";
 	}
+	if (provider === "cheapest" || provider === "fastest") {
+		// TODO: Implement cheapest and fastest provider selection
+		throw new InferenceClientInputError(`'cheapest' and 'fastest' policies are not supported yet.`);
+	}
 	if (provider === "auto") {
 		if (!modelId) {
 			throw new InferenceClientInputError("Specifying a model is required when provider is 'auto'");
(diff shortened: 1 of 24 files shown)


packages/inference/src/lib/getInferenceProviderMapping.ts
```diff
@@ -179,6 +179,10 @@ export async function resolveProvider(
 		);
 		provider = "auto";
 	}
+	if (provider === "cheapest" || provider === "fastest") {
+		// TODO: Implement cheapest and fastest provider selection
+		throw new InferenceClientInputError(`'cheapest' and 'fastest' policies are not supported yet.`);
```
kefranabg, line 184: Not sure exactly how we handle the `cheapest/fastest` policy, if anyone can take care of this part or provide some guidance would be nice :)

packages/tasks-gen/snippets-fixtures/conversational-llm-non-stream/js/fetch/0.fastest.js
```diff
@@ -0,0 +1,27 @@
...
+}
+
+query({ 
+    messages: [
+        {
+            role: "user",
+            content: "What is the capital of France?",
+        },
+    ],
+    model: "meta-llama/Llama-3.1-8B-Instruct:fastest",
```
SBrandeis, line 24: How did those get generated?

kefranabg:
Seen with @SBrandeis actually this is PR is not needed, we can probably just add the suffix on the modelId from hub side 👀

PR #1830, opened 2025-11-03
[Inference Providers] Snippets: prefer the `namespace/model:provider` syntax for conversational

SBrandeis:
# TL;DR

For conversational task, when using the Inference Clients libraries, prefer the `namespace/model:provider` syntax over explicitly specifying the provider in the parameters

Also don't specify `provider="auto"` because it's the default


hanouticelina (approved):
LGTM!

PR #1833, opened 2025-11-05
[inference] fix: streaming conversational with `auto` provider

SBrandeis:
Follow-up to #1810 

Applies the same changes to the streaming counterpart


julien-c (approved):

Wauplin (approved):
Thanks! Sorry missing this one out 😬

PR #1835, opened 2025-11-06
fix(inference): support string URL output in Replicate text-to-image

zeke:
This PR adds support for Replicate text-to-image models that return a single URL string instead of an array of URLs.

This should fix https://github.com/replicate/huggingface-model-mappings/pull/31, which is currently failing:

image

## Solution

The `ReplicateTextToImageTask.getResponse` method previously only handled array outputs, causing errors for models that return a single string URL. This change makes the text-to-image implementation consistent with other tasks like `ReplicateImageToImageTask` and `ReplicateTextToVideoTask`, which already support both formats.

## Prompts

> Does the replicate text-to-image implementation in this codebase support models that return a URL as output? Or does it always expect an array of image URLs?

> Let's update it to allow URL strings as a valid output response

> commit the changes to a branch, push it, and open a PR against the upstream repo: huggingface/huggingface.js

### DIFF
--- a/packages/inference/src/providers/replicate.ts
+++ b/packages/inference/src/providers/replicate.ts
@@ -97,12 +97,7 @@ export class ReplicateTextToImageTask extends ReplicateTask implements TextToIma
 		void headers;
 
 		// Handle string output
-		if (
-			typeof res === "object" &&
-			"output" in res &&
-			typeof res.output === "string" &&
-			isUrl(res.output)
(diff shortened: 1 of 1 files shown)


hanouticelina (approved):
thanks for the fix @zeke! If there are any issues with a model next time, feel free to set its status to `staging` and switch it back to `live` once it's fixed 🤗

PR #1837, opened 2025-11-11
Adding LightGlue entry

ariG23498:
The root issue for context: https://github.com/alexstoken/image-matching-models/issues/45

I am trying to add more custom image matching libraries to the hub, so that we can track their download better and have more models uploaded to the Hub.

CC: @pcuenca


packages/tasks/src/model-libraries.ts
```diff
@@ -191,7 +191,7 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		prettyLabel: "Champ",
 		repoName: "Champ",
 		repoUrl: "https://github.com/fudan-generative-vision/champ",
-		countDownloads: `path:"champ/motion_module.pth"`,
+		countDownloads: `path:"champ/motion_module."`,
```
pcuenca: I think you accidentally deleted `pth` substrings lol

packages/tasks/src/model-libraries.ts
```diff
@@ -255,9 +255,9 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		docsUrl: "https://deepforest.readthedocs.io/en/latest/",
 		repoUrl: "https://github.com/weecology/DeepForest",
 	},
-	"depth-anything-v2": {
-		prettyLabel: "DepthAnythingV2",
-		repoName: "Depth Anything V2",
+	"de-anything-v2": {
+		prettyLabel: "DeAnythingV2",
+		repoName: "De Anything V2",
```
pcuenca: same here

packages/tasks/src/model-libraries.ts
ariG23498: Thanks for catching this 😥

packages/tasks/src/model-libraries.ts
```diff
@@ -623,6 +623,13 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		filter: false,
 		snippets: snippets.lerobot,
 	},
+	lightglue: {
+		prettyLabel: "LightGlue",
+		repoName: "LightGlue",
+		repoUrl: "https://github.com/cvg/LightGlue",
+		filter: false,
+		countDownloads: `path_extension:"pth"`,
```
Wauplin: Looking at the 5 repos in https://huggingface.co/models?other=lightglue, they all have the same structure with config.json + model.safetensors but no `.pth` file. Since downloads are counted by default based on `config.json`, the download counter is already enabled for these models (see screenshot). Is there a reason for changing that?

Image

Wauplin: Note that it's still possible to register `lightblue` as a library for better metadata + link to GH repo but I'd remove the download counter rule to fallback to default one:


```suggestion
```

ariG23498:
I see that [lightglue](https://huggingface.co/docs/transformers/main/model_doc/lightglue) is part of `transformers` which helps in building the `config.json` when uploaded. (@Vaibhavs10 these are the models that are uploaded to the Hub at the moment)

What I wanted to do was to upload all the checkpoints from the [GitHub release](https://github.com/cvg/LightGlue/releases/tag/v0.1_arxiv) to the [Hub](https://huggingface.co/ariG23498/LightGlue) and try to track downloads from it.

@Wauplin I agree that `.pth` as a counter would be counter intuitive, but for the sake of making least code changes and also helping people track number of downloads, I would need to track the `.pth` files. What do you suggest I do here?

pcuenca:
The counting rule is there because the original library code ([here](https://github.com/cvg/LightGlue/tree/1fd587b6f5a7a0befd31188f1e5d2f05740006d7) and through [this popular wrapper](https://github.com/alexstoken/image-matching-models)) doesn't use the transformers implementation or a `config.json` file.

I would suggest to add a download counting rule that covers both cases:

```
		countDownloads: `path_extension:"pth" OR path:"config.json"`,
```

@ariG23498, if we only use `path_extension:"pth"` then we'd disable the counts for the other models.

Wauplin:
Agree with @pcuenca on how to proceed!

PR #1839, opened 2025-11-13
[inference provider] wavespeed support images field

arabot777:
Support i2i and i2v to pass in the public parameters of images

test:
img_v3_02s0_e3d125d3-59e4-47d3-ac3d-8743a458e79g

### DIFF
--- a/packages/inference/src/providers/wavespeed.ts
+++ b/packages/inference/src/providers/wavespeed.ts
@@ -77,8 +77,8 @@ async function buildImagesField(
 		new Uint8Array(inputs instanceof ArrayBuffer ? inputs : await (inputs as Blob).arrayBuffer())
 	);
 	const images =
-		Array.isArray(hasImages) && (hasImages as unknown[]).every((x) => typeof x === "string")
-			? (hasImages as string[])
+		Array.isArray(hasImages) && hasImages.every((value): value is string => typeof value === "string")
+			? hasImages
 			: [base];
 	return { base, images };
 }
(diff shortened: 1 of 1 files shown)


packages/inference/src/providers/wavespeed.ts
```diff
@@ -69,6 +69,20 @@ interface WaveSpeedAISubmitTaskResponse {
...
+	inputs: Blob | ArrayBuffer,
+	hasImages: unknown
+): Promise<{ base: string; images: string[] }> {
+	const base = base64FromBytes(
+		new Uint8Array(inputs instanceof ArrayBuffer ? inputs : await (inputs as Blob).arrayBuffer())
+	);
+	const images =
+		Array.isArray(hasImages) && (hasImages as unknown[]).every((x) => typeof x === "string")
+			? (hasImages as string[])
+			: [base];
```
hanouticelina, line 82:
```suggestion
	const images =
		Array.isArray(hasImages) && hasImages.every((value): value is string => typeof value === "string")
			? hasImages
			: [base];
```

packages/inference/src/providers/wavespeed.ts
```diff
@@ -188,13 +202,11 @@ export class WavespeedAIImageToImageTask extends WavespeedAITask implements Imag
...
-		return {
-			...args,
-			inputs: args.parameters?.prompt,
-			image: base64FromBytes(
-				new Uint8Array(args.inputs instanceof ArrayBuffer ? args.inputs : await (args.inputs as Blob).arrayBuffer())
-			),
-		};
+		const hasImages =
+			(args as Record<string, unknown>)["images"] ??
+			(args.parameters as Record<string, unknown> | undefined)?.["images"];
```
hanouticelina: better to use dot accessing here 
```suggestion
		const hasImages =
			(args as { images?: unknown }).images ?? (args.parameters as Record<string, unknown> | undefined)?.images;
```

packages/inference/src/providers/wavespeed.ts
```diff
@@ -204,12 +216,10 @@ export class WavespeedAIImageToVideoTask extends WavespeedAITask implements Imag
...
```
hanouticelina: same as above
```suggestion
		const hasImages =
			(args as { images?: unknown }).images ?? (args.parameters as Record<string, unknown> | undefined)?.images;
```

PR #1872, opened 2025-12-08
Add Clara as a library

kyle-gion:
This PR is to add Clara library (ckpt file) tracking to enable download count for the model repos in:
1. https://huggingface.co/collections/nvidia/clara-molecular
2. https://huggingface.co/collections/nvidia/clara-biology

CC: @merveenoyan as a follow up to our slack conversation

### DIFF
--- a/packages/tasks/src/model-libraries.ts
+++ b/packages/tasks/src/model-libraries.ts
@@ -246,7 +246,7 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		repoName: "Clara",
 		filter: false,
 		repoUrl: "https://github.com/nvidia/clara",
-		countDownloads: `path_extension:"ckpt"`,
+		countDownloads: `path_extension:"ckpt" OR path:"config.json"`,
 	},
 	clipscope: {
 		prettyLabel: "clipscope",


packages/tasks/src/model-libraries.ts
```diff
@@ -241,6 +241,13 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		repoUrl: "https://github.com/amazon-science/chronos-forecasting",
 		snippets: snippets.chronos_forecasting,
 	},
+	clara: {
```
Wauplin, line 244: as @pcuenca mentioned, the models in https://huggingface.co/collections/nvidia/clara-molecular and https://huggingface.co/collections/nvidia/clara-biology should be tagged as `clara` so that they show up under https://huggingface.co/models?other=clara.

You can do that by updating each model card with 

```
library_name: clara
```

packages/tasks/src/model-libraries.ts
```diff
@@ -241,6 +241,13 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
...
+		prettyLabel: "Clara",
+		repoName: "Clara",
+		filter: false,
+		repoUrl: "https://github.com/nvidia/clara",
+		countDownloads: `path_extension:"ckpt"`,
```
Wauplin: I've checked all the models listed in the collections. Some of them are indeed using .ckpt files but others have a structure with `config.json` + `xxx.safetensors` (e.g. https://huggingface.co/nvidia/NV-CodonFM-Encodon-1B-v1/tree/main, https://huggingface.co/nvidia/NV-CodonFM-Encodon-Cdwt-1B-v1/tree/main, https://huggingface.co/nvidia/NV-CodonFM-Encodon-TE-1B-v1/tree/main, etc.). Would all the models with tagged as `clara` are not? 
- if yes, then we should update the count download rule to count both on .ckpt AND config.json files 
- if no, then the download count rule is fine but some models would not have the `clara` tag

Let us know what you think!

merveenoyan: @Wauplin we have discussed on Slack previously not to tag models with safetensors but only the ones with ckpt 🙌🏻

merveenoyan:
@kyle-gion will merge once model repos are updated!

packages/tasks/src/model-libraries.ts
Wauplin: Great!

kyle-gion: Yes, some of our models use safetensors + config.json (already have download metrics). This PR is meant to enable download metrics for our models that use ckpt files.

kyle-gion:
> @kyle-gion will merge once model repos are updated!

@merveenoyan 
The appropriate models have been tagged with `library_name: clara`

packages/tasks/src/model-libraries.ts
Wauplin: don't want to reopen the topic if it has been already discussed in depth but the advantage of having a rule like 

```suggestion
		countDownloads: `path_extension:"ckpt" OR path:"config.json"`,
```

would be to:
- still have the same counting rule for the repo where it already exists
- add a new rule for ckpt-based repos
- have both ckpt and safetensors based repos tagged as `clara` (which makes sense if they are all from the same family/library of models)
- all have a link to the github repo https://github.com/nvidia/clara

kyle-gion: @merveenoyan indicated that the `clara` tag won't be visible as a library filter. @Wauplin does extending the `clara` tag to model repos that use safetensors add discoverability?

Yes, they are all part of the same family of models -> our Clara open models

Wauplin: It won't add the `Clara` library in this list no

image

but from each model repo, it'll be possible to list all `Clara` models by clicking on this tag 

image


So overall, yes, I do believe it helps with discoverability

kyle-gion: Great! I accepted the suggestion, thanks for flagging it!

Wauplin:
Great! Expect to see it live on the Hub within a few days :hugs:

PR #1875, opened 2025-12-10
Add VideoX Fun library details

bubbliiiing:


packages/tasks/src/model-libraries.ts
```diff
@@ -1370,6 +1370,13 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		snippets: snippets.vibevoice,
 		filter: false,
 	},
+	"videox_fun": {
+		prettyLabel: "VideoX Fun",
+		repoName: "VideoX Fun",
+		repoUrl: "https://github.com/aigc-apps/VideoX-Fun",
+		filter: false,
+		countDownloads: `path_extension:"safetensors"`,
```
Wauplin, line 1401: This rule means that downloading any safetensors files from the repo will count as a download. This can be detrimental if multiple safetensors files are present in the repo as they would double-count them.

Looking at https://huggingface.co/alibaba-pai/CogVideoX-Fun-V1.5-5b-InP, it seems there are a few json config files in the repo. Would it be possible to count downloads on a specific one so we are sure we never double count?

bubbliiiing: Hello, since videox-fun may contain many models—some with numerous files and others with only a few—how should I configure it?

Wauplin: Do you have more examples of repos so I can have a look? Is there any pattern in the repos or not? How does VideoX Fun knows which files to load or not?

bubbliiiing: Some libraries follow a format like that of Diffusers—similar to the one you provided—while others contain only a single .safetensors file.

https://huggingface.co/alibaba-pai/Z-Image-Turbo-Fun-Controlnet-Union

Wauplin: Ok I've reviewed most repos in https://huggingface.co/alibaba-pai/models?search=fun and I think it's fine to keep `path_extension:"safetensors"` as it is done now :+1:

packages/tasks/src/model-libraries.ts
```diff
@@ -1370,6 +1370,13 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		snippets: snippets.vibevoice,
 		filter: false,
 	},
+	"videox_fun": {
```
```suggestion
	videox_fun: {
```

bubbliiiing:
I've got it all done.

Wauplin:
Excellent! Merging it now. Expect to get it live on the Hub in the next few days

PR #1879, opened 2025-12-12
[Inference Providers] add inference support of `image-text-to-image` and `image-text-to-video` for fal

hanouticelina:
this PR adds support for the `image-text-to-image` and `image-text-to-video` tasks for fal-ai provider. These tasks support models that accept both optional image + text input, like FLUX.2 and LTX-Video. 

when an image is provided, the full endpoint is used (e.g., `fal-ai/flux-2/edit`). When no image is provided, the last path segment is stripped to use the text-only endpoint (e.g., `fal-ai/flux-2`).


packages/inference/src/providers/fal-ai.ts
```diff
@@ -284,6 +291,38 @@ export class FalAIImageToImageTask extends FalAiQueueTask implements ImageToImag
...
+	private useTextOnlyEndpoint = false;
+
+	constructor() {
+		super();
+		this.task = "image-text-to-image";
+	}
+
+	override makeRoute(params: UrlParams): string {
+		let model = params.model;
+		if (this.useTextOnlyEndpoint) {
```
Wauplin: Note that this makes the helper subject to concurrency issues. We discussed a bit internally and no good solution so far 😕 

The main problem is to make `makeRoute` dependent on `preparePayloadAsync` input.

hanouticelina: @SBrandeis any good idea regarding that?

Wauplin: (turning PR is "changes requested" to avoid forgetting about this thread)

SBrandeis: Would something like 6a3a9163fea78feea65fcea6b487a97055dbaf30 work?

```typescript
export class FalAIImageTextToImageTask extends FalAIImageToImageTask implements ImageTextToImageTaskHelper {

	/// [...]
	override async preparePayloadAsync(args: ImageTextToImageArgs): Promise<RequestArgs> {
		if (args.inputs) {
			return super.preparePayloadAsync(args as ImageToImageArgs);
		}
		return {
			...omit(args, ["inputs", "parameters"]),
			...(args.parameters as Record<string, unknown>),
			prompt: args.parameters?.prompt,
			urlTransform: (url) => {
				const urlObj = new URL(url);
				// Strip last path segment: fal-ai/flux-2/edit => fal-ai/flux-2
				urlObj.pathname = urlObj.pathname.split("/").slice(0, -1).join("/");
				return urlObj.toString();
			},
		} as RequestArgs;
	}
}
```

I'm not sure to understand the "concurrency issue" argument though - do you have links to internal context maybe?

Wauplin: (as discussed in DMs, the concurrency issue mentioned was due to `this.useTextOnlyEndpoint` being a class attribute but used and updated for each individual request. Meaning that if requests were made in parallel, it could lead to unexpected behavior)

PR #1882, opened 2025-12-15
[Inference Providers] Add `image-text-to-image` and `image-text-to-video` support for wavespeed

hanouticelina:
Discussed internally in this [thread](https://huggingface.slack.com/archives/C089DMDB415/p1765379147053229).
This PR adds support for the `image-text-to-image` and `image-text-to-video` tasks for wavespeed provider. These tasks support models that accept both optional image + text input.
For these tasks, we use the same endpoints as for `image-to-image` and `image-to-video`, when no image is provided, we send a 1x1 fully transparent image. Note that wavespeed's doesn't follow a consistent endpoint naming pattern that would allow us to have the same logic as for fal in #1879.

I've tested it (i.e. text only input) with [FLUX-2](https://huggingface.co/black-forest-labs/FLUX.2-dev) and  and the results looks reasonable to me:

"A robot playing chess in a garden"
wavespeed_placeholder_5

"A butterfly landing on a flower"

https://github.com/user-attachments/assets/8e751770-765f-4000-8aa4-57978339a0bc


pcuenca:
> 1x1 fully transparent image

Nice!

pcuenca (approved):

PR #1884, opened 2025-12-16
Add SGLang to local apps

LysandreJik:


packages/tasks/src/local-apps.ts
```diff
@@ -401,6 +437,15 @@ export const LOCAL_APPS = {
 			(model.pipeline_tag === "text-generation" || model.pipeline_tag === "image-text-to-text"),
 		snippet: snippetVllm,
 	},
+	sglang: {
+		prettyLabel: "SGLang",
+		docsUrl: "https://docs.sglang.io",
+		mainTask: "text-generation",
+		displayOnModelPage: (model: ModelData) =>
+			isTransformersModel(model) &&
+			(model.pipeline_tag === "text-generation" || model.pipeline_tag === "image-text-to-text"),
```
pcuenca: Are VLMs supported already? Reminder to open a PR [to this page](https://github.com/sgl-project/sglang/blob/4792d1f452031fafe3dadb723aaee7f568765e52/docs/supported_models/transformers_fallback.md?plain=1#L3) then. We can also update [the blog post](https://huggingface.co/blog/transformers-backend-sglang), it's the first Google hit for "sglang transformers".


^ I can take these whenever we are ready.

packages/tasks/src/local-apps.ts
```diff
@@ -251,6 +252,41 @@ curl -X POST "http://localhost:8000/v1/completions" \\
 		},
 	];
 };
+const snippetSglang = (model: ModelData): LocalAppSnippet[] => {
+	const messages = getModelInputSnippet(model) as ChatCompletionInputMessage[];
+
+	const setup = ["# Install SGLang from pip:", "pip install sglang"].join("\n");
+	const serverCommand = `# Start the SGLang serve:\npython3 -m sglang.launch_server --model-path ${model.id} \\
```
pcuenca:
```suggestion
	const serverCommand = `# Start the SGLang server:\npython3 -m sglang.launch_server --model-path ${model.id} \\
```

packages/tasks/src/local-apps.ts
LysandreJik: SGLang does support VLMs, but not with the transformers backend

Wauplin:
```suggestion
			(model.pipeline_tag === "text-generation" || model.pipeline_tag === "image-text-to-text") &&
			model.tags.includes("conversational"),
```

shouldn't we restrict to conversational models only? (i.e. the ones with a chat template)

julien-c: can't hurt 👍

PR #1886, opened 2025-12-17
Add SAME (embodied navigation) as library

GengzeZhou:
## Description

Adds download tracking support for [SAME](https://github.com/GengzeZhou/SAME) (State-Adaptive Mixture of Experts), an ICCV 2025 paper on generic language-guided visual navigation.

SAME distributes pretrained model weights as `.pt` files on HuggingFace: https://huggingface.co/ZGZzz/SAME

## Changes

- Added `same` entry to `model-libraries.ts` with `countDownloads` for `.pt` extension
- Added usage snippet to `model-libraries-snippets.ts`

## Related

- Paper: [SAME: Learning Generic Language-Guided Visual Navigation with State-Adaptive Mixture of Experts](https://arxiv.org/abs/2412.05552)
- GitHub: https://github.com/GengzeZhou/SAME
- HuggingFace: https://huggingface.co/ZGZzz/SAME

### DIFF
--- a/packages/tasks/src/model-libraries-snippets.ts
+++ b/packages/tasks/src/model-libraries-snippets.ts
@@ -1466,17 +1466,6 @@ outputs = estimator.process_one_image(image)
 rend_img = visualize_sample_together(image, outputs, estimator.faces)`,
 ];
 
-export const same = (model: ModelData): string[] => [
-	`# Download pretrained weights
-python download.py --checkpoints
-
-# Or load checkpoint directly with PyTorch
-import torch
-from huggingface_hub import hf_hub_download
-
-checkpoint = torch.load(hf_hub_download("${model.id}", "ckpt/SAME.pt"))`,
-];
-
 export const sampleFactory = (model: ModelData): string[] => [
 	`python -m sample_factory.huggingface.load_from_hub -r ${model.id} -d ./train_dir`,
 ];
--- a/packages/tasks/src/model-libraries.ts
+++ b/packages/tasks/src/model-libraries.ts
@@ -1034,8 +1034,7 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		repoName: "SAME",
 		repoUrl: "https://github.com/GengzeZhou/SAME",
 		filter: false,
-		snippets: snippets.same,
-		countDownloads: `path_extension:"pt"`,
+		countDownloads: `path:"ckpt/SAME.pt" OR path:"pretrain/Attnq_pretrained_ckpt.pt"`,
 	},
 	"sample-factory": {
 		prettyLabel: "sample-factory",


packages/tasks/src/model-libraries-snippets.ts
```diff
@@ -1466,6 +1466,17 @@ outputs = estimator.process_one_image(image)
...
+	`# Download pretrained weights
+python download.py --checkpoints
+
+# Or load checkpoint directly with PyTorch
+import torch
+from huggingface_hub import hf_hub_download
+
+checkpoint = torch.load(hf_hub_download("${model.id}", "ckpt/SAME.pt"))`,
+];
+
```
pcuenca: Snippets are usually short and concise inference examples, not necessarily download instructions.

packages/tasks/src/model-libraries.ts
```diff
@@ -1029,6 +1029,14 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		snippets: snippets.sam_3d_objects,
 		countDownloads: `path:"checkpoints/pipeline.yaml"`,
 	},
+	"same": {
+		prettyLabel: "SAME",
+		repoName: "SAME",
+		repoUrl: "https://github.com/GengzeZhou/SAME",
+		filter: false,
+		snippets: snippets.same,
+		countDownloads: `path_extension:"pt"`,
```
pcuenca: This would match `SAME.pt` in [this folder](https://huggingface.co/ZGZzz/SAME/tree/main/ckpt), but also [these checkpoints](https://huggingface.co/ZGZzz/SAME/tree/main/pretrain). If the intention is to only count downloads of SAME.pt, then we can use a more restrictive rule, such as `path:"ckpt/SAME.pt"`.

packages/tasks/src/model-libraries-snippets.ts
Wauplin: Agree with @pcuenca , a built-in script to load the model would be awesome!

packages/tasks/src/model-libraries.ts
```diff
@@ -1029,6 +1029,14 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		snippets: snippets.sam_3d_objects,
 		countDownloads: `path:"checkpoints/pipeline.yaml"`,
 	},
+	"same": {
```
```suggestion
	same: {
```

(for linting)

packages/tasks/src/model-libraries-snippets.ts
GengzeZhou: Thanks for the feedback! SAME is an embodied navigation agent that works quite differently from standard models. It requires processing language instructions + visual observations at each timestep to output navigation actions. It doesn't fit the typical `model.generate()` pattern since navigation involves agentic rollouts through simulated environments and is not a single-step QA.

The model inference is tightly coupled to an agent class and simulation environment. The agent handles the rollout loop, querying the model at each step with preprocessed inputs (tokenized instructions, encoded visual features, spatial encodings, read connectivity graphs) from the environment, and the environment class provides feedback for the model. Because of this, a minimal standalone forward-pass snippet would be incomplete and not runnable without the full agent/environment setup.

Here I wrote three options:

**Option A: Full runnable pipeline**
```python
import torch
from huggingface_hub import snapshot_download
from omegaconf import OmegaConf
from tasks.loaders import create_environments, create_dataloaders
from tasks.agents import load_agent

snapshot_download("${model.id}", local_dir="SAME")
config = OmegaConf.load("src/configs/test.yaml")
envs = create_environments(config, logger=None)
val_loaders = create_dataloaders(config, None, envs, training=False, device="cuda")
agent = load_agent("duet", config, envs, "cuda")

ckpt = torch.load("ckpt/SAME.pt", map_location="cuda")
agent.model.load_state_dict(ckpt["model_state_dict"])
agent.model.eval()

for name, loader in val_loaders.items():
    dataset = loader.get_dataset()
    for batch in loader:
        trajectories, _ = agent.rollout(batch, dataset, None, feedback="argmax", validate=True)
        print(trajectories[0]["path"])  # predicted navigation path
        break
```
Pros: Complete and runnable
Cons: Long

**Option B: Model API example (not directly runnable)**
```python
import torch
from huggingface_hub import hf_hub_download
from omegaconf import OmegaConf
from models.model import VLNBert

config = OmegaConf.load("configs/default.yaml")
model = VLNBert(config).cuda().eval()
ckpt = torch.load(hf_hub_download("${model.id}", "ckpt/SAME.pt"))
model.load_state_dict(ckpt["model_state_dict"])

# Forward: language -> panorama -> navigation
txt_embeds = model("language", {"txt_ids": tokens, "txt_masks": masks})
pano_embeds, _ = model("panorama", {"view_img_fts": features, "loc_fts": loc_fts})
action = model("navigation", {...})["fused_logits"].argmax(dim=1)
```
Pros: Shows model API
Cons: Inputs (tokens, features) require preprocessing not shown

**Option C: Model loading only**
```python
import torch
from huggingface_hub import hf_hub_download
from omegaconf import OmegaConf
from models.model import VLNBert

config = OmegaConf.load("configs/default.yaml")
model = VLNBert(config).cuda().eval()
ckpt = torch.load(hf_hub_download("${model.id}", "ckpt/SAME.pt"))
model.load_state_dict(ckpt["model_state_dict"])
```
Pros: Short and concise
Cons: No inference shown

Which approach do you prefer?

---

packages/tasks/src/model-libraries.ts
GengzeZhou: Good catch. Users will download either the pretrained checkpoint (`pretrain/Attnkv_pretrained_ckpt.pt`) or the finetuned checkpoint (`ckpt/SAME.pt`). Should I update it to:

```
countDownloads: `path:"ckpt/SAME.pt" OR path:"pretrain/Attnkv_pretrained_ckpt.pt"`
```

Wauplin: yes, that sounds the correct form to use!

packages/tasks/src/model-libraries-snippets.ts
Wauplin: Thanks for the detailed answer! Honestly it's hard to tell what's the best solution here^^ I'd be interested in knowing @pcuenca's opinion but maybe since the library in itself is more focused on research purposes than "easy to run inference", we could just skip the snippet for now? Would be great to document this in the repo's README directly instead.

Wauplin: Adding a snippet is optional and often skipped by libraries (if too complex to provide a one-off self.contained snippet). To do that, you'd just have to remove this line from your PR:

```
snippets: snippets.same
```

Let's me know what you think!

pcuenca: Thanks a lot for the details! 🙌 

I agree with @Wauplin, perhaps the best solution is to skip the snippet and provide more details in the model card!

Wauplin (approved):
All good! Thanks for the update @GengzeZhou , I'll merge once CI is green :)

Wauplin:
(expect to see it live on the Hub within a few days :hugs: )

PR #1887, opened 2025-12-18
Add hy-worldplay library

onpix:


onpix:
> Thanks for your PR @onpix ! Nice and clean as I like them 😃 And https://huggingface.co/models?other=hy-worldplay is already populated so we are good to merge. Expect it to be live on the Hub within a few days 🤗

Hi, although I have updated the model library and configured the download track method in this PR, the download count of our model repo still does not show correctly:

截屏2025-12-18 23 51 16

Could you please advise on how to resolve this issue?

Wauplin:
Yes, as mentioned in my message above it will be shipped on the Hub in the next days :)

> Expect it to be live on the Hub within a few days 🤗

ZuoXuhui:
image
Hi, just wanted to check in on something. We've updated the *library_name* to "hy-worldplay", but I'm still seeing "diffusers" displayed. Shouldn't the *prettyLabel* be showing as "HY-worldplay" instead? Thanks! :)

Wauplin:
As mentioned above, the PR has been merged but not yet deployed on the Hub. It will happen in the next few days

ZuoXuhui:
> As mentioned above, the PR has been merged but not yet deployed on the Hub. It will happen in the next few days

Oh, thanks for your reply.

PR #1888, opened 2025-12-18
Add sharp as a library

pcuenca:
Model: https://huggingface.co/apple/Sharp

### DIFF
--- a/packages/tasks/src/model-libraries.ts
+++ b/packages/tasks/src/model-libraries.ts
@@ -788,6 +788,13 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		filter: true,
 		countDownloads: `path_extension:"onnx"`,
 	},
+	"ml-sharp": {
+		prettyLabel: "Sharp",
+		repoName: "Sharp",
+		repoUrl: "https://github.com/apple/ml-sharp",
+		filter: false,
+		countDownloads: `path_extension:"pt"`,
+	},
 	mlx: {
 		prettyLabel: "MLX",
 		repoName: "MLX",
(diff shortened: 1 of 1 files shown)


pcuenca:
Skipped the snippet for now.

packages/tasks/src/model-libraries.ts
```diff
@@ -1089,6 +1089,13 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		snippets: snippets.setfit,
 		filter: true,
 	},
+	sharp: {
```
pcuenca: I saw some diffusion models that use "sharp" as a tag. Perhaps we could have gone with "ml-sharp" instead, but the library_name was [already merged](https://huggingface.co/apple/Sharp/blob/main/README.md?code=true#L3)

merveenoyan:
```suggestion
	"ml-sharp": {
```
let's go with this. they tagged sharp themselves, they previously merged my Image-to-3D tag so they're responsive

pcuenca: It'd be `"ml-sharp"` in this case 😂 (linter quotes)

packages/tasks/src/model-libraries.ts
```diff
@@ -1089,6 +1089,13 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		snippets: snippets.setfit,
 		filter: true,
 	},
+	"ml-sharp": {
```
Wauplin: Agree with "ml-sharp" given it's also the repo name 👍 Could you move the entry up in the mapping for alphabetical order?

pcuenca: 🤦

pcuenca: > alphabetical order

lol, forgot, thanks for the catch!

pcuenca:
@merveenoyan I'll let you merge and open the Hub PR as you see fit, since you've interacted with the Sharp team more.

packages/tasks/src/model-libraries.ts
Wauplin: :hugs:

Wauplin:
I've merged it (likely) get it in tomorrow's release but don't forget to update the model metadata :)

pcuenca:
Thanks @Wauplin! 🙌

PR #1889, opened 2025-12-19
Add llama-cli example to llama.cpp snippets

gary149:
Add `llama-cli` command alongside `llama-server` for all installation methods.

## Changes
- Add `llama-cli` snippet for direct terminal inference
- Update `llama-server` comment to mention web UI
- Build from source now compiles both targets

image

### DIFF
--- a/packages/tasks/src/local-apps.ts
+++ b/packages/tasks/src/local-apps.ts
@@ -110,20 +110,27 @@ function getQuantTag(filepath?: string): string {
 }
 
 const snippetLlamacpp = (model: ModelData, filepath?: string): LocalAppSnippet[] => {
-	const command = (binary: string) => {
-		const snippet = ["# Load and run the model:", `${binary} -hf ${model.id}${getQuantTag(filepath)}`];
+	const serverCommand = (binary: string) => {
+		const snippet = [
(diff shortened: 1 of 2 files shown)


Wauplin (approved):
LGTM 👍

packages/tasks/src/local-apps.ts
```diff
@@ -110,20 +110,27 @@ function getQuantTag(filepath?: string): string {
 }
 
 const snippetLlamacpp = (model: ModelData, filepath?: string): LocalAppSnippet[] => {
-	const command = (binary: string) => {
-		const snippet = ["# Load and run the model:", `${binary} -hf ${model.id}${getQuantTag(filepath)}`];
+	const serverCommand = (binary: string) => {
+		const snippet = [
+			"# Start a local OpenAI-compatible server with a web UI:",
```
ngxson, line 115: small nits (not related to the current PR), can we modify the copy button such that it doesn't copy the line starts with `#` (I mean comment lines)

gary149: yes good catch @copilot open an issue about it 👀

julien-c: @copilot please open an issue about this

PR #1910, opened 2026-01-14
✨ Add apps  params to listModels

nagesh161007:
## What does this PR do?

Adds support for filtering models by `apps`parameters in the `listModels` function, enabling users to filter models based on supported applications

## Changes

- **Add `apps?: string[]`** - Filter models by supported applications (vllm, ollama, tgi, mlx-lm, etc.)

## Example Usage

```typescript
import { listModels } from "@huggingface/hub";

for await (const model of listModels({
  search: { 
    apps: ["mlx-lm"],
  },
  limit: 10
})) {
  console.log(model.name);
}
```

### DIFF
--- a/packages/hub/src/lib/list-models.ts
+++ b/packages/hub/src/lib/list-models.ts
@@ -68,6 +68,9 @@ export async function* listModels<
 			 * Will search for models that have one of the inference providers in the list.
 			 */
 			inferenceProviders?: string[];
+			/**
+			 * Will search for models that support at least one of those local apps (eg "lmstudio", "mlx-lm", ...)
+			 */
 			apps?: string[];
 		};
 		hubUrl?: string;


nagesh161007:
@Wauplin, when you have some time, could you please take a look at this small item. Also needs workflow approval from maintainers

packages/hub/src/lib/list-models.spec.ts
```diff
@@ -134,4 +134,71 @@ describe("listModels", () => {
...
+		for await (const entry of listModels({
+			search: { apps },
+			limit: 10,
+		})) {
+			count++;
+			expect(entry.name).to.be.a("string").and.is.not.empty;
+		}
+
+		expect(count).to.equal(10);
+	});
```
Wauplin: These tests are not really testing the apps (it's only counting that the limit is correct)

packages/hub/src/lib/list-models.spec.ts
```diff
@@ -134,4 +134,71 @@ describe("listModels", () => {
...
+		for await (const entry of listModels({
+			search: { apps: ["vllm"] },
+			limit: 10,
+		})) {
+			count++;
+			expect(entry.name).to.be.a("string").and.is.not.empty;
+		}
+
+		expect(count).to.equal(10);
+	});
```
Wauplin:
```suggestion
	it("should search models by apps", async () => {
		let count = 0;
		for await (const entry of listModels({
			search: { apps: ["mlx-ml"] },
			limit: 10,
		})) {
			count++;
			expect(entry.tags).to.include("mlx");
		}

		expect(count).to.equal(10);
	});
```

I would replace the first one with an MLX search + remove the second one

packages/hub/src/lib/list-models.ts
```diff
@@ -92,6 +94,8 @@ export async function* listModels<
 			...(params?.search?.inferenceProviders
 				? { inference_provider: params.search.inferenceProviders.join(",") }
 				: undefined),
+			...(params?.search?.apps ? { apps: params.search.apps.join(",") } : undefined),
+			...(params?.search?.library ? { library: params.search.library.join(",") } : undefined),
```
Wauplin: This filter (`library`) doesn't work in the API.

See:
- https://huggingface.co/models => first model is `zai-org/GLM-Image`
- https://huggingface.co/models?library=onnx&sort=trending => first model is `Supertone/supertonic-2`
- https://huggingface.co/api/models?library=onnx&limit=1 => first model is `zai-org/GLM-Image` as if it wasn't taken into account

The correct solution to filter by library is to use the generic `filter`: https://huggingface.co/api/models?filter=onnx&limit=1

nagesh161007: Thanks @Wauplin — I’ll update the tests and remove library from the params.

Where can I find the documentation for the https://huggingface.co/api/models API? I checked https://huggingface.co/spaces/huggingface/openapi, but this endpoint doesn’t seem to be listed there.

Also, a quick question: does this API require a user token, or is it publicly accessible (possibly with rate limits)?

packages/hub/src/lib/list-models.spec.ts
nagesh161007: done

packages/hub/src/lib/list-models.spec.ts
nagesh161007: removed and made correction

packages/hub/src/lib/list-models.ts
Wauplin: > Where can I find the documentation for the https://huggingface.co/api/models API? I checked https://huggingface.co/spaces/huggingface/openapi, but this endpoint doesn’t seem to be listed there.


If it's not listed there, then it's not properly documentation (this is a WIP). You can usually get inspiration from the Python SDK ([here](https://github.com/huggingface/huggingface_hub/blob/fe64ebdd55feb16fa6ef04177b98fcd6e1a20128/src/huggingface_hub/hf_api.py#L2705)).

> Also, a quick question: does this API require a user token, or is it publicly accessible (possibly with rate limits)?

This API is public for public repos and obviously requires a token for private repos. Rate limits for unauthenticated users is rather low so it's usually recommended to use a user token - not just for this API but in general on the Hub. Authenticated users get higher rate limits, and much higher when they sign up for PRO.

packages/hub/src/lib/list-models.ts
```diff
@@ -68,6 +68,8 @@ export async function* listModels<
 			 * Will search for models that have one of the inference providers in the list.
 			 */
 			inferenceProviders?: string[];
+			apps?: string[];
+			library?: string[];
```
Wauplin:
```suggestion
```

packages/hub/src/lib/list-models.ts
nagesh161007: Thank you so much for patiently answering my queries

packages/hub/src/lib/list-models.ts
```diff
@@ -68,6 +68,7 @@ export async function* listModels<
 			 * Will search for models that have one of the inference providers in the list.
 			 */
 			inferenceProviders?: string[];
+			apps?: string[];
```
coyotte508, line 74:
```suggestion
			/**
			 * Will search for models that support at least one of those local apps (eg "lmstudio", "mlx-lm", ...)
			 */
			apps?: string[];
```

PR #1902, opened 2026-01-11
Add glm-tts as a library

zRzRzRzRzRzRzR:

PR #1911, opened 2026-01-14
Update model-libraries.ts

AkaliKong:


pcuenca:
Hello @AkaliKong, you are modifying the type definitions, which won't work. If you want to register a new library, [here's an example PR](https://github.com/huggingface/huggingface.js/pull/1902/files) that does it, and [here are the docs](https://huggingface.co/docs/hub/en/models-adding-libraries#register-your-library). Also, please try to provide a meaningful title and a concise description when you open a new PR.

PR #1916, opened 2026-01-14
Add Pocket-TTS as a library

NielsRogge:
This PR registers [Pocket-TTS](https://github.com/kyutai-labs/pocket-tts) as a library. This enables download stats + a custom code snippet which people will see when clicking the "Use this model" button at the top right.

The `library_name: pocket-tts` was already merged here: https://huggingface.co/kyutai/pocket-tts-without-voice-cloning/discussions/1. This one should be merged too: https://huggingface.co/kyutai/pocket-tts/discussions/5 (alongside adding `pipeline_tag: text-to-speech`)

cc @gabrieldemarmiesse

### DIFF
--- a/packages/tasks/src/model-libraries.ts
+++ b/packages/tasks/src/model-libraries.ts
@@ -950,7 +950,7 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		repoUrl: "https://github.com/kyutai-labs/pocket-tts",
 		snippets: snippets.pocket_tts,
 		filter: false,
-		countDownloads: `path_extension:"safetensors"`,
+		countDownloads: `path:"tts_b6369a24.safetensors"`,
 	},
 	"pruna-ai": {
 		prettyLabel: "Pruna AI",


packages/tasks/src/model-libraries.ts
```diff
@@ -944,6 +944,14 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		filter: false,
 		countDownloads: `path_extension:"pth"`,
 	},
+	"pocket-tts": {
+		prettyLabel: "Pocket-TTS",
+		repoName: "PocketTTS",
+		repoUrl: "https://github.com/kyutai-labs/pocket-tts",
+		snippets: snippets.pocket_tts,
+		filter: false,
+		countDownloads: `path_extension:"safetensors"`,
```
```suggestion
		countDownloads: `path:"tokenizer.model"`,
```

Feels like we should aim at counting downloads on a single file (so either [tokenizer.model](https://huggingface.co/kyutai/pocket-tts/blob/main/tokenizer.model) or [tts_b6369a24.safetensors](https://huggingface.co/kyutai/pocket-tts/blob/main/tts_b6369a24.safetensors)) to avoid double counting on each voice embedding download. I've chosen the tokenizer file arbitrarily as it felt more generic

gabrieldemarmiesse: Could we count the downloads of tts_b6369a24.safetensors separately for the two repos on huggingface? https://huggingface.co/kyutai/pocket-tts and https://huggingface.co/kyutai/pocket-tts-without-voice-cloning so that we can have stats about the model with voice cloning and without? Note that in all cases, only the  tokenizer.model of the repo without voice cloning is used as it's not gated.

Wauplin: Yes sure, as you prefer it :) 

```suggestion
		countDownloads: `path:"tts_b6369a24.safetensors"`,
```

Wauplin: The download count is always per repo so you'll get 2 counts as expected

Wauplin:
Merging it now. @gabrieldemarmiesse expect 2-3 days before getting it live on the Hub :)

gabrieldemarmiesse:
Thanks a lot!!!

PR #1917, opened 2026-01-15
add yolo26 as library

merveenoyan:
I assume community will be pushing models soon @pcuenca as discussed


Wauplin (approved):
Looks good 👍

PR #1918, opened 2026-01-15
🐛 Fix local file parsing in gguf package

### LINKED ISSUE #1915: @huggingface/gguf: ArrayBuffer resize fails on certain GGUF files with local file access
## Description

`gguf()` fails with `ArrayBuffer.prototype.resize: Invalid length parameter` when parsing certain GGUF files using `allowLocalFile: true`.

## Reproduction

```javascript
const { gguf } = require('@huggingface/gguf')
const result = await gguf('/path/to/embeddinggemma-300m-Q4_0.gguf', { allowLocalFile: true })
```

**Failing model:** https://huggingface.co/unsloth/embeddinggemma-300m-GGUF/resolve/main/embeddinggemma-300m-Q4_0.gguf

## Environment

- Node.js: v22.17.0
- @huggingface/gguf: 0.3.2
- OS: macOS

## Notes

The same file parses successfully using the Python reference implementation (`pip install gguf` from llama.cpp). This confirms the issue is specific to the JS library.

The error occurs in the `appendBuffer` method when calling `this.buffer.resize()`. Possibly related to Node.js v22's native `ArrayBuffer.resize` implementation or the `HTTP_TOTAL_MAX_SIZE` (50MB) limit.

yuranich:
## What problem does this PR solve?

Fixes #1915 - `gguf()` fails with `ArrayBuffer.prototype.resize: Invalid length parameter` when parsing GGUF files using `allowLocalFile: true`.

**Two bugs were identified:**

1. **Off-by-one error in chunk fetching:** `RangeViewLocalFile.fetchChunk()` calculates range as `[start, end-1]` (inclusive end, matching HTTP Range headers), but passes it directly to `Blob.slice(start, end)` which uses an **exclusive** end. This causes each chunk to be 1 byte short, creating gaps at chunk boundaries that break metadata parsing.

2. **Negative slice size for small files:** `FileBlob.slice()` could create slices with negative size when the requested range starts beyond EOF. This affects small files and sharded GGUF files where the file is smaller than the 2MB chunk size.

## How does it solve it?

1. Changed `fetchChunk()` to pass the correct exclusive end to `Blob.slice()`
2. Added bounds clamping in `FileBlob.slice()` to ensure start position never exceeds file end

## Breaking changes

None - this is a bugfix that makes local file parsing work correctly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,7 +98,7 @@ You can run our packages with vanilla JS, without any bundler, by using a CDN or
 
 ```html
 <script type="module">
-    import { InferenceClient } from 'https://cdn.jsdelivr.net/npm/@huggingface/inference@4.13.8/+esm';
+    import { InferenceClient } from 'https://cdn.jsdelivr.net/npm/@huggingface/inference@4.13.9/+esm';
     import { createRepo, commit, deleteRepo, listFiles } from "https://cdn.jsdelivr.net/npm/@huggingface/hub@2.7.1/+esm";
 </script>
 ```
--- a/packages/inference/package.json
+++ b/packages/inference/package.json
@@ -1,6 +1,6 @@
 {
 	"name": "@huggingface/inference",
-	"version": "4.13.8",
+	"version": "4.13.9",
 	"packageManager": "pnpm@10.10.0",
 	"license": "MIT",
 	"author": "Hugging Face and Tim Mikeladze <tim.mikeladze@gmail.com>",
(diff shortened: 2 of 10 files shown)


coyotte508:
Thanks, nice catches

coyotte508:
cc @mishig25 / @ngxson

yuranich:
thanks!
@mishig25 @ngxson could you help please with review and merge?

ngxson (approved):

PR #1919, opened 2026-01-16
[Inference Providers] fix outputType for all providers

hanouticelina:
related to [this comment](https://github.com/huggingface-internal/moon-landing/pull/16383#issuecomment-3758868630) (private link).

When using `textToImage` with `outputType: "url"`, together and nebius providers returned `base64` URLs instead of actual http URLs. Some providers don't support URL format : `hf-inference`, `nscale`, `hyperbolic`.

cc @kefranabg

### DIFF
--- a/packages/inference/src/providers/hf-inference.ts
+++ b/packages/inference/src/providers/hf-inference.ts
@@ -73,6 +73,7 @@ import type {
 
 import { TaskProviderHelper } from "./providerHelper.js";
 import { base64FromBytes } from "../utils/base64FromBytes.js";
+import { dataUrlFromBlob } from "../utils/dataUrlFromBlob.js";
 import type { ImageToImageArgs } from "../tasks/cv/imageToImage.js";
 import type { AutomaticSpeechRecognitionArgs } from "../tasks/audio/automaticSpeechRecognition.js";
 import { omit } from "../utils/omit.js";
@@ -156,26 +157,17 @@ export class HFInferenceTextToImageTask extends HFInferenceTask implements TextT
 				return await base64Response.blob();
 			}
 			if ("output" in response && Array.isArray(response.output)) {
-				if (outputType === "dataUrl") {
-					// Fetch the URL and convert to dataUrl
-					const urlResponse = await fetch(response.output[0]);
-					const blob = await urlResponse.blob();
-					const b64 = await blob.arrayBuffer().then((buf) => Buffer.from(buf).toString("base64"));
-					return `data:image/jpeg;base64,${b64}`;
-				}
 				const urlResponse = await fetch(response.output[0]);
 				const blob = await urlResponse.blob();
-				return blob;
+				return outputType === "dataUrl" ? dataUrlFromBlob(blob) : blob;
 			}
 		}
 		if (response instanceof Blob) {
 			if (outputType === "dataUrl") {
-				const b64 = await response.arrayBuffer().then((buf) => Buffer.from(buf).toString("base64"));
-				return `data:image/jpeg;base64,${b64}`;
+				return dataUrlFromBlob(response);
 			}
 			if (outputType === "json") {
-				const b64 = await response.arrayBuffer().then((buf) => Buffer.from(buf).toString("base64"));
-				return { output: `data:image/jpeg;base64,${b64}` };
+				return { output: await dataUrlFromBlob(response) };
 			}
 			return response;
 		}
--- a/packages/inference/src/providers/zai-org.ts
+++ b/packages/inference/src/providers/zai-org.ts
@@ -21,6 +21,7 @@ import {
 } from "../errors.js";
 import { isUrl } from "../lib/isUrl.js";
 import type { BodyParams, HeaderParams, OutputType } from "../types.js";
+import { dataUrlFromBlob } from "../utils/dataUrlFromBlob.js";
 import { delay } from "../utils/delay.js";
 import { omit } from "../utils/omit.js";
 import { BaseConversationalTask, TaskProviderHelper, type TextToImageTaskHelper } from "./providerHelper.js";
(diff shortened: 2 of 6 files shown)


packages/inference/src/providers/hyperbolic.ts
```diff
@@ -118,6 +118,7 @@ export class HyperbolicTextToImageTask extends TaskProviderHelper implements Tex
 				return { ...response };
 			}
 			if (outputType === "url") {
+				console.warn("hyperbolic provider does not support URL output, returning base64 data URL instead");
 				return `data:image/jpeg;base64,${response.images[0].image}`;
```
kefranabg, line 126: Personally, I'd expect an exception to be thrown rather than an alternative format. IMO choosing the outputType should be left to the user, based on the formats supported by the providers

hanouticelina: indeed!

hanouticelina: addressed in [ba01d73](https://github.com/huggingface/huggingface.js/pull/1919/commits/ba01d7352aa82048bdce27318f34ad65085c70d5)

packages/inference/src/providers/hf-inference.ts
```diff
@@ -141,7 +141,9 @@ export class HFInferenceTextToImageTask extends HFInferenceTask implements TextT
 			if ("data" in response && Array.isArray(response.data) && response.data[0].b64_json) {
 				const base64Data = response.data[0].b64_json;
 				if (outputType === "url") {
-					return `data:image/jpeg;base64,${base64Data}`;
+					throw new InferenceClientInputError(
```
kefranabg: Isn't there a way to know if the provider supports the output before the inference? so we can throw before? Cause I would be quite frustrated to consume some inference credits to get an exception lol

packages/inference/src/providers/hyperbolic.ts
Wauplin: could we add a new `dataUrl` type then? because here we are introducing a breaking change so at least we should provide a drop-in replacement I think

packages/inference/src/providers/hf-inference.ts
hanouticelina: yes, my bad on this one, i didn't push the commit that moved this part to `preparePayload` 🙈 (although i'm not sure if there is a better way/place to raise the error)

done in [7e18606](https://github.com/huggingface/huggingface.js/pull/1919/commits/7e186065792db30a55a1a82adc51cbafff341eaa)

packages/inference/src/providers/hyperbolic.ts
hanouticelina: ^yes agree, i added `dataUrl` type in [7e18606](https://github.com/huggingface/huggingface.js/pull/1919/commits/7e186065792db30a55a1a82adc51cbafff341eaa) 
regarding the breaking change, I first thought it was okay since it was kind of a bug to return base64 URLs when outputType: "url" (similar to `response_format: "url"` in other APIs), users would expect an actual http URL but agree about adding a drop-in replacement 👍

SBrandeis (approved):

packages/inference/src/providers/hf-inference.ts
```diff
@@ -140,25 +149,33 @@ export class HFInferenceTextToImageTask extends HFInferenceTask implements TextT
...
 				const base64Response = await fetch(`data:image/jpeg;base64,${base64Data}`);
 				return await base64Response.blob();
 			}
 			if ("output" in response && Array.isArray(response.output)) {
-				if (outputType === "url") {
-					return response.output[0];
+				if (outputType === "dataUrl") {
+					// Fetch the URL and convert to dataUrl
+					const urlResponse = await fetch(response.output[0]);
+					const blob = await urlResponse.blob();
```
Wauplin: this is redundant with lines 166 and 167 no? They could be move above the `if (outputType === "dataUrl") {` statement and blob is either converted to base64 or return directly

packages/inference/src/providers/zai-org.ts
```diff
@@ -179,7 +179,12 @@ export class ZaiTextToImageTask extends TaskProviderHelper implements TextToImag
 				}
 
 				const imageResponse = await fetch(imageUrl);
-				return await imageResponse.blob();
+				const blob = await imageResponse.blob();
+				if (outputType === "dataUrl") {
+					const b64 = await blob.arrayBuffer().then((buf) => Buffer.from(buf).toString("base64"));
+					return `data:image/jpeg;base64,${b64}`;
```
Wauplin: I feel that we could have a utility "dataUrlFromBlob" given how many times we use it. And then do

```ts
return outputType === "dataUrl" ? dataUrlFromBlob(blob) : blob;
```

everywhere

Wauplin: (can be another naming^^)

Wauplin (approved):
Just some nitpicking comments but otherwise logic looks good to me ✔️

packages/inference/src/providers/zai-org.ts
hanouticelina: yes! addressed in [4e5bd1a](https://github.com/huggingface/huggingface.js/pull/1919/commits/4e5bd1a34164e7e1c5eff7f7818b9cf8e8ac1982)

packages/inference/src/providers/hf-inference.ts
hanouticelina: addressed in [4e5bd1a](https://github.com/huggingface/huggingface.js/pull/1919/commits/4e5bd1a34164e7e1c5eff7f7818b9cf8e8ac1982)

hanouticelina:
thank you for the reviews!
@kefranabg i merged and released `@huggingface/inference@4.13.10`

PR #1932, opened 2026-01-26
Adding pickle to the track

theSPGN:


theSPGN:
@pcuenca Hi, yeah I wanted to add my private model to the track, so whenever pikle file with whole model is downloaded it counts it in stats. I created 2 PRs for a pikle format. Do I need to specify it with the url to HF model? I think now yaml file is being counted.

PR #1933, opened 2026-01-26
Add pickle function for model loading

theSPGN:
Add a function to load a model using pickle from Hugging Face Hub.


theSPGN:
> hey sorry but we don't promote pickle anymore as it's quite insecure for community sharing

Hi, is there any other format you recommend instead?

Wauplin:
> Hi, is there any other format you recommend instead?

Yes! For saving model weights, we recommend `safetensors` which has been specifically designed with safety in mind: https://huggingface.co/docs/safetensors/index
For general purpose pickles it really depends on what's saved in it.

PR #1935, opened 2026-01-27
Add StormScope registry

pzharrington:
Adds the `registry.json` tracker for new StormScope model on HuggingFace

### DIFF
--- a/packages/tasks/src/model-libraries.ts
+++ b/packages/tasks/src/model-libraries.ts
@@ -1190,14 +1190,6 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		countDownloads: `path:"model.safetensors"`,
 		snippets: snippets.stable_audio_tools,
 	},
-	stormscope: {
-		prettyLabel: "StormScope",
-		repoName: "StormScope",
-		repoUrl: "https://github.com/NVIDIA/earth2studio",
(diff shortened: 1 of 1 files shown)


packages/tasks/src/model-libraries.ts
```diff
@@ -1190,6 +1190,14 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		countDownloads: `path:"model.safetensors"`,
 		snippets: snippets.stable_audio_tools,
 	},
+	stormscope: {
+		prettyLabel: "StormScope",
+		repoName: "StormScope",
+		repoUrl: "https://github.com/NVIDIA/earth2studio.git",
```
```suggestion
		repoUrl: "https://github.com/NVIDIA/earth2studio",
```

pzharrington:
Hi @Wauplin thanks for the responses! I've addressed (1) and (2) as you suggested

pzharrington:
Actually, sorry for the churn @Wauplin but I think we don't need this update anymore. For context, we version pin all our imports of these packages to specific commit hashes for stability/reproducibility, and were trying to avoid having to change the model's commit hash so current users of our package would have their downloads counted. I realized that since we updated the model card in the HF repo already to address (1) via your PR, the "current" git commit hash already changed, so now we need to update it in our source code anyways.

PR #1936, opened 2026-01-28
Update translation, summarization and image-to-text pipeline snippets to warn on v5 removal

Wauplin:
**EDIT:**

**Expectation:**  warning message displayed in the pipeline snippet on https://huggingface.co/facebook/nllb-200-distilled-1.3B?library=transformers.

```
# Warning: Pipeline type "${model.pipeline_tag}" is no longer supported in transformers v5.
# You must load the model directly (see below) or downgrade to v4.x with:
# 'pip install "transformers<5.0.0'
```

----

(related to internal slack threads [here](https://huggingface.slack.com/archives/C01N44FJDHT/p1768843199944399?thread_ts=1768482758.054089&cid=C01N44FJDHT) and [here](https://huggingface.slack.com/archives/C02FRVDEZND/p1769542989564529))

Related to `transformers v5` release, `translation`, `summarization`, and `image-to-text` pipelines no longer exist. See migration guide: https://github.com/huggingface/transformers/blob/main/MIGRATION_GUIDE_V5.md#pipelines

> pipeline classes are intended as a high-level beginner-friendly API, but for almost all text-to-text tasks a modern chat model and TextGenerationPipeline will provide much higher quality output. As a result, we felt it was misleading for beginners to offer the older pipelines.

Note that these pipelines are not defined anymore but the models are still usable with the lower-level `AutoModel` API. 

Affected models:
- https://huggingface.co/models?pipeline_tag=translation
- https://huggingface.co/models?pipeline_tag=summarization
- https://huggingface.co/models?pipeline_tag=image-to-text

cc @julien-c @LysandreJik @Rocketknight1

### DIFF
--- a/packages/tasks/src/library-to-tasks.ts
+++ b/packages/tasks/src/library-to-tasks.ts
@@ -71,10 +71,5 @@ export const LIBRARY_TASK_MAPPING: Partial<Record<ModelLibraryKey, PipelineType[
 	mindspore: ["image-classification"],
 };
 
-
 // Pipeline types that were supported in legacy transformers versions (<5.0.0)
-export const REMOVED_IN_V5_TRANSFORMERS_PIPELINES: PipelineType[] = [
-	"summarization",
-	"translation",
-	"image-to-text",
-];
+export const REMOVED_IN_V5_TRANSFORMERS_PIPELINES: PipelineType[] = ["image-to-text", "summarization", "translation"];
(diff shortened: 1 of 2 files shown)


packages/tasks/src/library-to-tasks.ts
```diff
@@ -49,19 +49,19 @@ export const LIBRARY_TASK_MAPPING: Partial<Record<ModelLibraryKey, PipelineType[
...
-		"summarization",
+		// "summarization", dropped in transformers v5.0.0
 		"table-question-answering",
 		"text-classification",
 		"text-generation",
 		"text-to-audio",
 		"text-to-speech",
 		"token-classification",
-		"translation",
+		// "translation", dropped in transformers v5.0.0
```
julien-c: i would move those to a `REMOVED_IN_V5_TRANSFORMERS_PIPELINES` const

Wauplin:
@julien-c PR updated to keep the pipeline snippet but show a warning message

PR #1938, opened 2026-01-28
[Inference Providers] fix Replicate `image-to-image` compatibility with different model schemas

hanouticelina:
flagged by @merveenoyan [here](https://huggingface.slack.com/archives/C089DMDB415/p1769608268868169) (private).

this PR fixes Replicate `image-to-image` compatibility with different model schemas by passing the input image in multiple payload keys (`image`, `images`, `input_image`, `input_images`)
a bit hacky but we already do that for other tasks/providers.

### DIFF
--- a/packages/tasks-gen/snippets-fixtures/image-to-image/python/requests/0.replicate.py
+++ b/packages/tasks-gen/snippets-fixtures/image-to-image/python/requests/0.replicate.py
@@ -17,7 +17,14 @@ def query(payload):
 image_bytes = query({
     "input": {
         "prompt": "Turn the cat into a tiger.",
-        "input_image": "cat.png"
+        "image": "cat.png",
+        "images": [
+            "cat.png"
+        ],
+        "input_image": "cat.png",
+        "input_images": [
+            "cat.png"
+        ]
     }
 })
 


packages/inference/src/providers/replicate.ts
```diff
@@ -236,18 +233,23 @@ export class ReplicateAutomaticSpeechRecognitionTask
...
 		return {
 			input: {
 				...omit(params.args, ["inputs", "parameters"]),
 				...params.args.parameters,
-				input_image: params.args.inputs, // This will be processed in preparePayloadAsync
+				// Different Replicate models expect the image in different keys
+				image: imageInput,
+				images: [imageInput],
+				input_image: imageInput,
+				input_images: [imageInput],
```
SBrandeis, line 252: lmao

hanouticelina: 🤷‍♀️ 😭

Wauplin: 🫠

Wauplin: what if `imageInput` is a 1MB base64 payload? can it happen?

Wauplin: I'm fine with this solution though if that's the only way. Can we contact Replicate in parallel though, would be best if they can make their API more consistent

hanouticelina: Yes, it can happen but to be honest i don't think there is another solution client-side, we have one model <> one schema unfortunately

PR #1952, opened 2026-02-01
Add negative step support in jinja range

TonyLianLong:
## Summary of changes
  - Add negative step support to `range` in jinja utilities.
  - Extend `utils` tests to cover negative-step behavior.

## Reason for the change
  - Right now negative steps in jinja is not supported. This leads to issues handling chat templates such as [this one](https://huggingface.co/spaces/huggingfacejs/chat-template-playground?modelId=moonshotai/Kimi-K2-Thinking) that uses a negative step size.

## Tests
Tested locally with a script:
```javascript
import { Template } from "./dist/index.js";

const tmpl = new Template("{% for i in range(3, -1, -1) %}{{ i }} {% endfor %}");
console.log(tmpl.render({}));
```

The output should be:
```
3 2 1 0
```


xenova:
Thanks for the PR! 🤗 Honestly, I'm surprised we never encountered this issue before 😅 
~Perhaps some logic can be shared with the slice function (which does support negative indexing). https://github.com/TonyLianLong/huggingface.js/blob/04574d45e4ecc3c1cc036471eece8295d9f93ec5/packages/jinja/src/utils.ts#L39~ Actually not necessary because that logic takes array length into consideration.

In addition to the utils test you added, can you add some template tests in https://github.com/huggingface/huggingface.js/blob/main/packages/jinja/test/templates.test.js? Thanks!

TonyLianLong:
Thanks for the quick review. A test has been added.

xenova:
Thanks again ✅ new release out, and I've bumped the version in the [chat template playground](https://huggingface.co/spaces/huggingfacejs/chat-template-playground).
image

PR #1955, opened 2026-02-02
 ✅ fix(tasks,hub): update test expectations to match current API responses

rgthelen:
**Summary**                                                                                                        
                                                                                                                 
Fixes failing tests by updating expectations to match current API behavior.                                    
                                                                                                                 
**Changes**                                                                                                        
                                                                                                                 
packages/tasks/src/local-apps.spec.ts                                                                          
                                                                                                                 
Updated vLLM test expectations to match snippet output changed in #1948:                                       
  - Load and run the model: → # Start the vLLM server:                                                         
  - Call the server using curl: → # Call the server using curl (OpenAI-compatible API):                        
                                                                                                                 
packages/hub/src/lib/list-models.spec.ts                                                                       
                                                                                                                 
  Changed test from deepseek-ai to meta-llama owner. The first deepseek-ai model returned (DeepSeek-OCR-2) now   
  has an empty inferenceProviderMapping, while meta-llama models consistently have providers.                    
                                                                                                                 
packages/hub/src/lib/paths-info.spec.ts                                                                        
                                                                                                                 
  Removed expectation for securityFileStatus field which is no longer returned by the API.

### DIFF
--- a/packages/tasks/src/model-libraries-snippets.ts
+++ b/packages/tasks/src/model-libraries-snippets.ts
@@ -2277,6 +2277,29 @@ export const pythae = (model: ModelData): string[] => [
 model = AutoModel.load_from_hf_hub("${model.id}")`,
 ];
 
+export const qwen3_tts = (model: ModelData): string[] => [
+	`# pip install qwen-tts
+import torch
+import soundfile as sf
+from qwen_tts import Qwen3TTSModel
+
+model = Qwen3TTSModel.from_pretrained(
+    "${model.id}",
+    device_map="cuda:0",
+    dtype=torch.bfloat16,
+    attn_implementation="flash_attention_2",
+)
+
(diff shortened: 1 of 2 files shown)


packages/hub/src/lib/paths-info.spec.ts
```diff
@@ -49,7 +48,6 @@ describe("pathsInfo", () => {
 
 		// should include expand info
 		expect(modelPathInfo.lastCommit).toBeDefined();
-		expect(modelPathInfo.securityFileStatus).toBeDefined();
```
Wauplin, line 52: @coyotte508 can you confirm this is expected and not a regression server-side ? (the fact that security status is not returned in  `paths-info`

coyotte508: well it's still in the code probably this specific model doesn't have securityStatus info

PR #1959, opened 2026-02-04
bugfix: use-this-model of PaddleOCR-VL models

TingquanGao:
I didn’t realize that the model ID includes the organization name. PTAL ~

### DIFF
--- a/packages/tasks/src/model-libraries-snippets.ts
+++ b/packages/tasks/src/model-libraries-snippets.ts
@@ -1159,8 +1159,8 @@ for res in output:
 	}
 
 	if (model.tags.includes("document-parse")) {
-		const rawVersion = model.id.replace("PaddleOCR-VL-", "v");
-		const version = rawVersion === "PaddleOCR-VL" ? "v1" : rawVersion;
+		const rawVersion = model.id.replace("PaddlePaddle/PaddleOCR-VL-", "v");
+		const version = rawVersion === "PaddlePaddle/PaddleOCR-VL" ? "v1" : rawVersion;
 		return [
 			`# See https://www.paddleocr.ai/latest/version3.x/pipeline_usage/PaddleOCR-VL.html to installation
 


pcuenca:
https://huggingface.co/PaddlePaddle/PaddleOCR-VL?library=PaddleOCR looks like this:

Screenshot 2026-02-04 at 09 59 29

And https://huggingface.co/PaddlePaddle/PaddleOCR-VL-1.5?library=PaddleOCR looks like this:
Screenshot 2026-02-04 at 10 00 03

packages/tasks/src/model-libraries-snippets.ts
```diff
@@ -1159,8 +1159,8 @@ for res in output:
 	}
 
 	if (model.tags.includes("document-parse")) {
-		const rawVersion = model.id.replace("PaddleOCR-VL-", "v");
-		const version = rawVersion === "PaddleOCR-VL" ? "v1" : rawVersion;
+		const rawVersion = model.id.replace("PaddlePaddle/PaddleOCR-VL-", "v");
+		const version = rawVersion === "PaddlePaddle/PaddleOCR-VL" ? "v1" : rawVersion;
```
pcuenca, line 1163: So the idea is just to use "v1.5" for https://huggingface.co/PaddlePaddle/PaddleOCR-VL-1.5 and "v1" for https://huggingface.co/PaddlePaddle/PaddleOCR-VL?

It looks like that [according to the source code](https://github.com/PaddlePaddle/PaddleOCR/blob/main/paddleocr/_pipelines/paddleocr_vl.py#L26), but the model card for the v1 model [does not include a pipeline version argument in the code snippet](https://huggingface.co/PaddlePaddle/PaddleOCR-VL#basic-usage).

As an alternative, you could go with something like `model.id.endsWith("-1.5")`.

TingquanGao: Yes, you are absolutely right! There is indeed an issue here: [this model card](https://huggingface.co/PaddlePaddle/PaddleOCR-VL#basic-usage), and we will fix it as soon as possible. Additionally, the reason we are not using `model.id.endsWith("-1.5")` is to ensure compatibility with future versions such as v2.0 and beyond.

TingquanGao: Thanks for your suggestion! I have fixed: https://huggingface.co/PaddlePaddle/PaddleOCR-VL/discussions/87

pcuenca:
Merging then, thanks @Wauplin! 🙌

PR #1967, opened 2026-02-09
Add VisMatch library for image matching models

gmberton:


ariG23498 (commented):
The org in HF that holds `vismatch` models: https://huggingface.co/vismatch

This looks good to me, thanks @gmberton 

Just to be sure, do we want https://github.com/huggingface/huggingface.js/blob/f3d12aa5c98e3801fa0e7123a32236204813d9a3/packages/tasks/src/model-libraries.ts#L598 and `vismatch` together? Or we can replace one in favour of the other?

Wauplin:
Hey! Your PR looks good overall, but there are three things we need to sort out before merging:

### 1. There's Already an `image-matching-models` Library
**The Issue:** The Hub already has `image-matching-models` listed, and all 65+ models under the `vismatch` org are using it.

**What to Do:** You've got two options here:
- **Rename it:** If you're transitioning from `image-matching-models` to `vismatch`:
  - Remove the old `image-matching-models` entry
  - Update all those model cards to use `library_name: vismatch` instead
- **Keep both:** If they serve different purposes, just let us know how they differ

**Why this matters:** Having both without clarity will confuse users browsing the Hub.

### 2. No Models Are Actually Using `vismatch` Yet
**The Issue:** Searching for `vismatch` on the Hub returns zero models (see https://huggingface.co/models?other=vismatch`)

**What to Do:** Update at least one model card to include:
```yaml
library_name: vismatch
```

**Why this matters:** The Hub needs at least one real model using the library before it can be added.

### 3. The Download Counter Won't Work
**The Issue:** Your rule looks for `vismatch.yaml`, but that file doesn't exist in any of the repos (they only have `README.md`, `.gitattributes`, and `dummy.safetensors`).

**What to Do:** Either:
- Add a `vismatch.yaml` config file to your model repos, or
- Change the rule to `path_extension:"safetensors"` to match existing files

**Why this matters:** The download counter needs an actual file to track.

---

This review was AI-generated and manually reviewed before posting.

## Quick Path Forward

1. Let us know if you're renaming or keeping both libraries
2. Add `library_name: vismatch` to your model cards
3. Fix the countDownloads rule
4. We'll get this merged!

Happy to help if you have questions about any of this!

gmberton:
Hi @Wauplin, thank you for your reply! I am aware of those 3 issues, I was waiting for the merge to fix them :)

image-matching-models is the old name of vismatch, and we're renaming all instances of image-matching-models to vismatch.
Right now we're tracking download counts through .safetensors files (in the image-matching-models), but we want to track custom `vismatch.yaml` files to offer more flexibility. Specifically there are some models that require to download N .safetensors files (while most models require only 1) which was increasing the count by N for each download, making it unfair towards models using only 1 .safetensors file.

So these are the changes I'm planning, as you said:
1. edit the README of the models to change the library name to vismatch
2. add a `vismatch.yaml` file to every model

Let me know if you would like me to make the changes before the merge, instead of after it

Wauplin (approved):
Perfect! Then this is the good way forward yes. I'd prefer to merge this PR only when the models are updated on the Hub so we don't merge and forget about them. Let me know once it's done :)

gmberton:
Thank you all! I have updated [all 68 models](https://huggingface.co/vismatch/models) with the tag: `library_name: vismatch
`.

We are planning to keep the [image-matching-models](https://huggingface.co/image-matching-models) for a few weeks (people are still using it). We didn't rename the image-matching-models to vismatch because we wanted to reset the models download counts (for fairness to models that we added later), hence we create [vismatch](https://huggingface.co/vismatch) from scratch.
So if that's okay for you, it would be nice to keep also this for some time
https://github.com/huggingface/huggingface.js/blob/f3d12aa5c98e3801fa0e7123a32236204813d9a3/packages/tasks/src/model-libraries.ts#L598
Otherwise we can remove the image-matching-models entry and leave only the vismatch one :)

Wauplin (approved):
All good, thanks for the heads-up! 

It's not a problem if you want to keep https://huggingface.co/models?other=image-matching-models to be honest. You can just add a disclaimer in there readmes saying it's not maintained anymore + redirect to the vismatch equivalent and still keep them on the Hub. 

Note that you've mentioned 68 models but:
- https://huggingface.co/models?other=vismatch has 67 models
- https://huggingface.co/vismatch has 69 models

Mentioning it in case there is a mismatch you haven't spotted. It's not a blocker to merge this PR. Expect 2-3 days before getting it live on the Hub 🤗

PR #1974, opened 2026-02-12
Add causal-forcing library

zhuhz22:
Causal Forcing is an improved version of Self Forcing.

### DIFF
--- a/packages/tasks/src/model-libraries.ts
+++ b/packages/tasks/src/model-libraries.ts
@@ -1526,6 +1526,13 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		snippets: snippets.zonos,
 		filter: false,
 	},
+	"causal-forcing": {
+		prettyLabel: "CausalForcing",
+		repoName: "CausalForcing",
+		repoUrl: "https://github.com/thu-ml/Causal-Forcing",
+		filter: false,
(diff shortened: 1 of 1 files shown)


Wauplin:
Thanks for the PR! :hugs:

A few things to address before we can merge:

1. **No models on Hub yet**: We need at least one model with `library_name: causal-forcing` in its model card before merging. Currently https://huggingface.co/models?other=causal-forcing shows 0 results. You can add it to the model card metadata like this:
```yaml
   library_name: causal-forcing
```
2. **Alphabetical order**: The entry should be placed between `cartesia_mlx` and `champ` (not at the end of the file). Libraries are sorted alphabetically by key.
3. **Missing trailing comma**: There's a missing `,` after the closing `}` of the entry — this will break the build.
4. **Download count rule**: `path_extension:"pt"` will count every `.pt` file in a repo. Since your models seem to have multiple `.pt` files (`ar_diffusion.pt`, `causal_ode.pt`, `causal_forcing.pt`), this would over-count downloads. I'd suggest picking one specific file, e.g. `path:"causal_forcing.pt"`, so each model download is counted once.

Happy to help if you have any questions! Let me know once there's at least one model tagged on the Hub and we'll get this merged.

---
*This review was AI-generated and manually reviewed before posting.*

PR #1975, opened 2026-02-12
Fix GGUF quant label parsing for Unsloth Dynamic (UD) prefixed filenames

mishig25:
## Summary

- GGUF filenames with the [Unsloth Dynamic "UD-" prefix ](https://unsloth.ai/docs/basics/unsloth-dynamic-2.0-ggufs)(e.g. `Qwen3-4B-UD-Q2_K_XL.gguf`) have their quant labels parsed, but the "UD-" prefix gets dropped — so `parseGGUFQuantLabel` returns `"Q2_K_XL"` instead of `"UD-Q2_K_XL"`. This means the "Hardware compatibility" section on model pages like https://huggingface.co/unsloth/Qwen3-4B-GGUF doesn't distinguish UD quants from regular ones.
- Updates `GGUF_QUANT_RE` to capture an optional `(?<prefix>UD-)?` named group before the quant type, so `parseGGUFQuantLabel` now returns `"UD-Q2_K_XL"` preserving the full label.
- Adds a test case in `gguf.spec.ts` for the UD-prefixed filename pattern.

image


mishig25:
cc: @danielhanchen @shimmyshimmer

danielhanchen:
Wonderful this works thanks!

PR #1982, opened 2026-02-17
fix bucket download file

coyotte508:
> [!NOTE]
> **Medium Risk**
> Changes download URL construction (including buckets), which can affect file retrieval behavior across repo types. Adds a new optional auth token flow for Xet downloads; incorrect token/caching inputs could cause download failures.
> 
> **Overview**
> Fixes `fileDownloadInfo` URL generation by removing the special-case `/api/buckets/...` path so bucket downloads use the same resolve/raw URL pattern as other repo types.
> 
> Adds support for passing a pre-fetched Xet JWT via `downloadFile` (`xet: { readToken }`), wiring it into `XetBlob` (new exported `XetReadToken` type) to seed the internal JWT cache and avoid an extra refresh request.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 9d8b73865d6e8f527bc2d88cb3387aeb57262732. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/hub/src/lib/download-file.ts
+++ b/packages/hub/src/lib/download-file.ts
@@ -2,6 +2,7 @@ import type { CredentialsParams, RepoDesignation } from "../types/public";
 import { checkCredentials } from "../utils/checkCredentials";
 import { WebBlob } from "../utils/WebBlob";
 import { XetBlob } from "../utils/XetBlob";
+import type { XetReadToken } from "../utils/XetBlob";
 import type { FileDownloadInfoOutput } from "./file-download-info";
 import { fileDownloadInfo } from "./file-download-info";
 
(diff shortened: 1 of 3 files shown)


packages/hub/src/lib/download-file.ts
```diff
@@ -32,9 +33,12 @@ export async function downloadFile(
 		/**
 		 * Whether to use the xet protocol to download the file (if applicable).
 		 *
+		 * When an object with `readToken` is provided along with `downloadInfo`,
+		 * the xet download can skip the token refresh roundtrip.
+		 *
 		 * @default true
 		 */
-		xet?: boolean;
+		xet?: boolean | { readToken: XetReadToken };
```
coyotte508, line 41: cc @mishig25 - just in case it's relevant w/ safetensors - if a token is already present (eg called from the hub's backend, where it can be generated), skips one roundtrip

(and manually computing the `downloadInfo` skips another)

not relevant in most cases when called from frontend

PR #1993, opened 2026-02-20
Fix workflow name in documentation.yml

kerrrang9214-tech:
Merge

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -1,4 +1,4 @@
-name: Build documentation
+kerrrangs pushname: Build documentation
 
 on:
   workflow_dispatch:


.github/workflows/documentation.yml
```diff
@@ -1,4 +1,4 @@
-name: Build documentation
+kerrrangs pushname: Build documentation
```
kerrrang9214-tech, line 1: Inference provider

kerrrang9214-tech (commented):
Approved

.github/workflows/documentation.yml
kerrrang9214-tech: Open

PR #2001, opened 2026-02-24
:wrench: Add Ryzen AI Max 395 hardware specifications

martinmarkwart:
tflops value is taken from this post on reddit:

https://www.reddit.com/r/LocalLLaMA/comments/1kmi3ra/amd_strix_halo_ryzen_ai_max_395_gpu_llm/

### DIFF
--- a/packages/tasks/src/hardware.ts
+++ b/packages/tasks/src/hardware.ts
@@ -549,7 +549,7 @@ export const SKUS = {
 				memory: [16, 32],
 			},
 			"Ryzen AI Max 395+": {
-				tflops: 36.9,
+				tflops: 59.4,
 				memory: [64, 96, 128],
 			},
 		},


packages/tasks/src/hardware.ts
```diff
@@ -548,6 +548,10 @@ export const SKUS = {
 				tflops: 26.11,
 				memory: [16, 32],
 			},
+			"Ryzen AI Max 395+": {
+				tflops: 36.9,
+				memory: [64, 96, 128],
+			},
```
pcuenca, line 554: Wouldn't this belong in the CPU section better?

pcuenca: Actually, it should be fine!

packages/tasks/src/hardware.ts
```diff
@@ -548,6 +548,10 @@ export const SKUS = {
 				tflops: 26.11,
 				memory: [16, 32],
 			},
+			"Ryzen AI Max 395+": {
+				tflops: 36.9,
```
pcuenca: I think we can go with the 59.4 theoretical number. Most of the other numbers in this page don't represent actual tested performance.

martinmarkwart: I'm totally fine with that - it doesn't seem to be a very good indicator for LLM Performance anyway.

pcuenca:
```suggestion
				tflops: 59.4,
```

martinmarkwart:
I just saw that there is a little Typo in the description...

it should be: Ryzen AI Max+ 395
instead of: Ryzen AI Max 395+

shall i create another pull-request for this?

pcuenca:
Ah yes, you are right, I missed it too! A new PR would be awesome!

martinmarkwart:
done:
[https://github.com/huggingface/huggingface.js/pull/2007](url)

PR #2025, opened 2026-03-08
feat(jinja): `mapping`, `sequence`, `safe`

giladgd:
Add missing features in the Jinja package to support [Qwen 3.5 function calling](https://huggingface.co/Qwen/Qwen3.5-4B/blob/main/chat_template.jinja#L122)

---

> [!NOTE]
> **Medium Risk**
> Changes template evaluation semantics (tests/filters and type detection), which could subtly affect rendering outcomes for existing templates even though the surface area is small and covered by new tests.
> 
> **Overview**
> Adds missing Jinja runtime behavior needed by Qwen 3.5 templates: introduces the `sequence` test, changes `mapping` to use `instanceof ObjectValue`, and implements the `safe` filter as a no-op.
> 
> Extends filter support by allowing numeric values to be converted via the `string` filter, and updates tests to cover `safe`/numeric `string` plus a new `Qwen/Qwen3.5-4B` e2e chat-template fixture and expected output.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit df80242778ef493f1859e9f9d5afe01cb7fa6c41. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/tasks/src/hardware.ts
+++ b/packages/tasks/src/hardware.ts
@@ -548,6 +548,10 @@ export const SKUS = {
 				tflops: 17.86,
 				memory: [8],
 			},
+			"RX 5500 XT": {
+				tflops: 10.39,
+				memory: [4, 8],
+			},
 			"Radeon Pro VII": {
 				tflops: 26.11,
 				memory: [16, 32],
@@ -706,6 +710,9 @@ export const SKUS = {
 			},
 			"Ryzen Threadripper Zen 1000 (Whitehaven)": {
 				tflops: 0.6,
+			},
+			"Ryzen 7 3800X (16)": {
+				tflops: 1.73,
 			},
 			"Ryzen Zen 5 9000 (Ryzen 9)": {
 				tflops: 0.56,


xenova (changes requested):
Thanks for the PR! Looks like cursor has an issue with the ObjectValue being a sequence, but other than that the PR looks great!

giladgd:
@xenova I think Cursor is wrong due to the lack of context, it thinks `ObjectValue` is a python-equivalent runtime object, but it actually represents a dict.

I've done a quick test in python:
```python
from jinja2 import Environment

env = Environment()

class PlainObject:
    pass

values = [
    ("dict", {"a": 1}),
    ("plain_object", PlainObject()),
]

for name, value in values:
    print(name, repr(value))
    print("  sequence =", env.tests["sequence"](value))
```

Which outputs:
```
dict {'a': 1}
  sequence = True
plain_object <__main__.PlainObject object at 0xffffb4871bd0>
  sequence = False
```

xenova:
Thanks for the explanation! You're spot on.
```py
from jinja2 import Environment
jinja_env = Environment(
    trim_blocks=True,
    lstrip_blocks=True,
)
template = jinja_env.from_string("""\
{{ 'hi' is sequence }}
{{ 1 is sequence }}
{{ True is sequence }}
{{ [1,2,3] is sequence }}
{{ {'a': 'b'} is sequence }}
""")
print(template.render())
```

outputs
```
True
False
False
True
True
```

running tests, then will merge

xenova:
minor formatting issue (fixable with `pnpm --filter @huggingface/jinja format`). fixed myself

xenova:
thanks again! :) new version out with this too: https://www.npmjs.com/package/@huggingface/jinja/v/0.5.6 🙌

giladgd:
@xenova Thanks for the quick review and release :)

xenova:
You're welcome! 🤗 thanks for contributing!

PR #2026, opened 2026-03-10
🛡️ safetensors: harden parser against crafted files (ReDoS, path traversal, resource exhaustion)

mishig25:
## Summary

Follow-up to #2020 (GGUF CWE-770 hardening) — same audit applied to `parse-safetensors-metadata.ts`.

- **Fix ReDoS in `isQuantizedTensor`** (High): `modules_to_not_convert` patterns from `config.json` were passed raw into `new RegExp(pattern)`, letting a malicious model author trigger catastrophic backtracking or an uncaught `SyntaxError`. Fixed by escaping all regex metacharacters in each literal glob segment before joining with `.*`.
- **Cap shard count** (Medium): `fetchAllHeaders` had no limit on unique shard filenames from `weight_map`. A crafted index.json within the 25 MB limit could list thousands of unique filenames, each triggering an HTTP request. Capped at `MAX_SHARD_COUNT = 10_000` (well above any real sharded model).
- **Validate shard filenames against path traversal** (Medium): filenames from `weight_map` were concatenated into request paths without sanitization. Now rejects any filename containing `..`, starting with `/`, or containing `://`.
- **Cap `config.json` reads** (Low): unlike safetensors headers and index files (both capped at 25 MB), `config.json` was fetched with no size limit. Capped at `MAX_CONFIG_LENGTH = 10_000_000` (10 MB).
- **Guard against `Infinity` shape products** (Low): `v.shape.reduce((a, b) => a * b)` on crafted large shape values (e.g. `[1e308, 2]`) overflows to `Infinity`, poisoning `parameterCount` results. Now skipped via `Number.isFinite` check.

## Test plan

- [x] Existing tests pass: `pnpm -C packages/hub test`

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it changes how sharded filenames and quantization exclusion patterns are interpreted/validated, which could reject some previously-accepted (but unusual) model repos. Core logic remains the same, but adds new error paths and limits around downloads and parameter counting.
> 
> **Overview**
> **Hardens safetensors metadata parsing against crafted inputs.** `config.json` reads are now capped (`MAX_CONFIG_LENGTH`) to avoid large memory use, and sharded `index.weight_map` processing enforces a maximum shard count plus basic filename sanitization to block path traversal/URL-like entries.
> 
> **Removes attacker-controlled RegExp usage** by replacing `modules_to_not_convert` matching with a non-RegExp glob matcher, avoiding ReDoS and invalid-pattern crashes. Parameter counting also skips tensors whose `shape` product overflows to non-finite values.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d627140d7c6a3ffdd74d778eec0fed228fce28e0. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/hub/src/lib/parse-safetensors-metadata.ts
+++ b/packages/hub/src/lib/parse-safetensors-metadata.ts
@@ -36,6 +36,8 @@ export function parseSafetensorsShardFilename(filename: string): SafetensorsShar
 
 const PARALLEL_DOWNLOADS = 20;
 const MAX_HEADER_LENGTH = 25_000_000; // 25MB
+const MAX_CONFIG_LENGTH = 10_000_000; // 10MB — config.json is typically small; cap to avoid large memory use
+const MAX_SHARD_COUNT = 10_000; // well above any real sharded model; blocks crafted index with millions of entries
 const GPTQ_QWEIGHT_SUFFIX = "qweight";
 const GPTQ_AWQ_AUXILIARY_SUFFIXES = ["qzeros", "g_idx", "scales"];
 
@@ -121,7 +123,7 @@ async function fetchModelConfig(
 			return null;
 		}
 
-		const config = JSON.parse(await configBlob.text());
+		const config = JSON.parse(await configBlob.slice(0, MAX_CONFIG_LENGTH).text());
 		return config as ModelConfig;
 	} catch (error) {
 		// Config file might not exist or be inaccessible
@@ -213,6 +215,16 @@ async function fetchAllHeaders(
 ): Promise<SafetensorsShardedHeaders> {
 	const pathPrefix = path.slice(0, path.lastIndexOf("/") + 1);
 	const filenames = [...new Set(Object.values(index.weight_map))];
+	if (filenames.length > MAX_SHARD_COUNT) {
+		throw new SafetensorParseError(
+			`Too many shard files (${filenames.length}). Maximum supported is ${MAX_SHARD_COUNT}.`,
+		);
+	}
+	for (const filename of filenames) {
+		if (filename.includes("..") || filename.startsWith("/") || filename.includes("://")) {
+			throw new SafetensorParseError(`Unsafe shard filename in weight_map: "${filename}"`);
+		}
+	}
 	const shardedMap: SafetensorsShardedHeaders = Object.fromEntries(
 		await promisesQueue(
 			filenames.map(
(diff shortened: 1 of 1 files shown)


packages/hub/src/lib/parse-safetensors-metadata.ts
```diff
@@ -356,7 +368,15 @@ function isQuantizedTensor(tensorName: string, quantConfig?: QuantizationConfig)
...
 	if (!patterns?.length) return true;
-	return !patterns.some((pattern) => new RegExp(pattern.replace(/\*/g, ".*")).test(tensorName));
+	return !patterns.some((pattern) => {
+		// Build a safe regex: escape all metacharacters in each literal segment, then join with .*
+		// This prevents ReDoS and SyntaxError from attacker-controlled patterns in config.json.
+		const escaped = pattern
+			.split("*")
+			.map((part) => part.replace(/[.+?^${}()|[\]\\]/g, "\\$&"))
+			.join(".*");
+		return new RegExp(escaped).test(tensorName);
```
coyotte508: maybe we could avoid regexps altogether or simplify regex use

mishig25: handled in https://github.com/huggingface/huggingface.js/blob/d627140d7c6a3ffdd74d778eec0fed228fce28e0/packages/hub/src/lib/parse-safetensors-metadata.ts#L369-L378

packages/hub/src/lib/parse-safetensors-metadata.ts
```diff
@@ -349,14 +361,30 @@ export interface ModelConfig {
...
+function globMatch(pattern: string, str: string): boolean {
+	const parts = pattern.split("*");
+	let pos = 0;
+	for (const part of parts) {
+		const idx = str.indexOf(part, pos);
+		if (idx === -1) return false;
+		pos = idx + part.length;
+	}
+	return true;
+}
```
coyotte508, line 378: @cursor I think there may be a bug for the first/last parts, can you double check?

Eg if glob is `abc_*_def` feels like `XXabc_x_def` could match when it shouldn't

feel free to correct me (and don't do any change to this PR, it's already merged)

coyotte508: cursor doesn't like me

coyotte508: anyway locally, cc @mishig25 :

```
Bug 1 — No start anchoring. The first segment (before the first *) must appear at position 0, but the function accepts it anywhere.

globMatch("abc", "xabc") → true (should be false)
globMatch("a*c", "xa1c") → true (should be false)
Bug 2 — No end anchoring. The last segment (after the last *) must align with the end of the string, but the function accepts the first occurrence it finds.

globMatch("*a", "bac") → true (should be false, string doesn't end with "a")
globMatch("abc", "abcx") → true (should be false)
Bug 3 — Consequence of Bug 2: first-match greediness for the last part. Even if you added a pos === str.length check at the end, it would still fail because indexOf finds the first occurrence rather than one aligned to the end. For example, globMatch("*ab", "aab"): indexOf("ab", 0) returns 0, giving pos = 2 ≠ 3, so it would incorrectly return false even though "aab" ends with "ab".
``` 

and suggested fix

```ts
function globMatch(pattern: string, str: string): boolean {
	const parts = pattern.split("*");

	if (parts.length === 1) {
		return pattern === str;
	}

	// First part must be anchored to the start
	if (!str.startsWith(parts[0])) return false;
	let pos = parts[0].length;

	// Last part must be anchored to the end
	const lastPart = parts[parts.length - 1];
	if (!str.endsWith(lastPart)) return false;
	const end = str.length - lastPart.length;

	// Middle parts must appear in order between pos and end
	for (let i = 1; i < parts.length - 1; i++) {
		const idx = str.indexOf(parts[i], pos);
		if (idx === -1 || idx + parts[i].length > end) return false;
		pos = idx + parts[i].length;
	}

	return pos <= end;
}
```


(note double check the fix)

mishig25: cursor bugbot ci didn't find any issues in https://github.com/huggingface/huggingface.js/runs/66439114867

handled the issues in https://github.com/huggingface/huggingface.js/pull/2031

PR #2036, opened 2026-03-13
Add BioNeMo and NV-Medtech libraries

kyle-gion:
This PR is to add two new libraries to enable download count for the model repos.

Librares:
- BioNeMo
- NV-Medtech

---

> [!NOTE]
> **Low Risk**
> Low risk: only adds two new entries to the model library UI/download-count mapping, with no changes to runtime logic beyond additional keys and Elastic download queries.
> 
> **Overview**
> Adds support for two additional model libraries in `MODEL_LIBRARIES_UI_ELEMENTS`: **`bionemo`** (BioNeMo) and **`nv-medtech`** (NV-MedTech).
> 
> Each entry includes repo metadata and a `countDownloads` Elastic query so models tagged with these library names can have download counts computed (without enabling the models page filter for either).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 892838399c0210cf2f92f629b38b36242b5ed441. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/tasks/src/model-libraries.ts
+++ b/packages/tasks/src/model-libraries.ts
@@ -899,7 +899,7 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		filter: true,
 		countDownloads: `path_extension:"nemo" OR path:"model_config.yaml" OR path_extension:"json"`,
 	},
-	nv-medtech: {
+	"nv-medtech": {
 		prettyLabel: "NV-MedTech",
 		repoName: "NV-MedTech",
 		filter: false,


packages/tasks/src/model-libraries.ts
```diff
@@ -892,6 +899,13 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		filter: true,
 		countDownloads: `path_extension:"nemo" OR path:"model_config.yaml" OR path_extension:"json"`,
 	},
+	nv-medtech: {
```
pcuenca:
```suggestion
	"nv-medtech": {
```

packages/tasks/src/model-libraries.ts
```diff
@@ -892,6 +899,13 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		filter: true,
 		countDownloads: `path_extension:"nemo" OR path:"model_config.yaml" OR path_extension:"json"`,
 	},
+	"nv-medtech": {
+		prettyLabel: "NV-MedTech",
+		repoName: "NV-MedTech",
+		filter: false,
+		repoUrl: "https://github.com/nvidia-medtech",
```
pcuenca, line 906: Should this be a specific repo?

kyle-gion: @pcuenca It's intentionally pointing to a github org. The org will house models that go with the "library".

packages/tasks/src/model-libraries.ts
```diff
@@ -166,6 +166,13 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		filter: false,
 		countDownloads: `path_extension:"npz"`,
 	},
+	bionemo: {
+		prettyLabel: "BioNeMo",
+		repoName: "BioNeMo",
+		filter: false,
+		repoUrl: "https://github.com/nvidia/BioNeMo",
+		countDownloads: `path_extension:"ckpt" OR path:"config.json"`,
```
Wauplin, line 174: If the model on the Hub always have a `config.json` then the good practice is to count only by this one. Otherwise you will end up double-counting downloads if user downloads both a config.json and a xxx.ckpt.

And since counting by `config.json` is the default rule on the Hub, it means you can completely remove the download count rule and the counter will work out of the box :) 


```suggestion
```

Wauplin: This is only on the assumption that all `bionemo` models will have a config.json. If it's not the case, you can ignore my comment

packages/tasks/src/model-libraries.ts
```diff
@@ -892,6 +899,13 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		filter: true,
 		countDownloads: `path_extension:"nemo" OR path:"model_config.yaml" OR path_extension:"json"`,
 	},
+	"nv-medtech": {
+		prettyLabel: "NV-MedTech",
+		repoName: "NV-MedTech",
+		filter: false,
+		repoUrl: "https://github.com/nvidia-medtech",
+		countDownloads: `path_extension:"pt" OR path_extension:"safetensors" OR path:"config.json"`,
```
Wauplin, line 907: Same comment here about the `config.json` file

```suggestion
```

Wauplin:
> Please, remember to flag models with `library_name: bionemo` and `library_name: nv-medtech` when you release them.

And yes, as @pcuenca mentioned you have to tag the models correctly on the Hub. You can already do it actually to avoid forgetting about it (even if models are still in private and if this PR is not yet merged)

packages/tasks/src/model-libraries.ts
kyle-gion: Hi @Wauplin, we typically only include a .ckpt file. We have the config.json option in there in case we use a different file type for the checkpoint. 

Examples: https://huggingface.co/collections/nvidia/clara-biology

Based on what you're saying, since `config.json` is the default rule, I'll remove it since adding it is redundant

Wauplin: Great! Then no need to change anything (hence why I've merged before you could push new changes). If you define a rule then it'll overwrite the default `config.json` rule, you if you want it applied your inital PR is exactly what's needed.

Wauplin:
Given https://github.com/huggingface/huggingface.js/pull/2036#discussion_r2933300972, I took the liberty to merge the PR :) Expect it to be live likely on Monday.

PR #2088, opened 2026-04-09
Update Use This Model instructions for Lemonade v10.2

jeremyfowers:
I'm one of the maintainers of Lemonade, which has been featured in the `Use this model` button on model cards ever since my teammate @danielholanda added it last year.

In Lemonade v10.2, we overhauled model pulling: the CLI was renamed from `lemonade-server` to `lemonade` and the HF happy path is now a single `lemonade pull owner/repo:VARIANT`. This makes usage a lot simpler and I'd like to merge the change here.

Old way, from the current `Use this model` button on `Tesslate/OmniCoder-9B-GGUF`:

```
lemonade-server pull user.OmniCoder-9B-GGUF --checkpoint Tesslate/OmniCoder-9B-GGUF:Q4_K_M --recipe llamacpp
```

New way, as of Lemonade v10.2:

```
lemonade pull Tesslate/OmniCoder-9B-GGUF:Q4_K_M
``` 

Thanks for your consideration!

### DIFF
--- a/packages/tasks/src/local-apps.ts
+++ b/packages/tasks/src/local-apps.ts
@@ -513,43 +513,39 @@ const snippetDockerModelRunner = (model: ModelData, filepath?: string): string =
 };
 
 const snippetLemonade = (model: ModelData, filepath?: string): LocalAppSnippet[] => {
-	const tagName = getQuantTag(filepath);
 	const modelName = model.id.includes("/") ? model.id.split("/")[1] : model.id;
+	const isRyzenAI = model.tags.some((tag) => ["ryzenai-npu", "ryzenai-hybrid"].includes(tag));
 
-	// Get recipe according to model type
-	let simplifiedModelName: string;
-	let recipe: string;
-	let checkpoint: string;
+	// Lemonade auto-registers pulled models as `user.<suggested_name>[-<variant>]`.
+	// For GGUF/llamacpp: suggested_name is the repo name and variant is the quant tag.
+	// For RyzenAI ONNX: there is no per-variant suffix.
+	let pullArg: string;
+	let runName: string;
 	let requirements: string;
-	if (model.tags.some((tag) => ["ryzenai-npu", "ryzenai-hybrid"].includes(tag))) {
-		recipe = model.tags.includes("ryzenai-npu") ? "oga-npu" : "oga-hybrid";
-		checkpoint = model.id;
-		requirements = " (requires RyzenAI 300 series)";
-		simplifiedModelName = modelName.split("-awq-")[0];
-		simplifiedModelName += recipe === "oga-npu" ? "-NPU" : "-Hybrid";
+	if (isRyzenAI) {
+		pullArg = model.id;
+		runName = `user.${modelName}`;
+		requirements = " (requires XDNA 2 NPU)";
 	} else {
-		recipe = "llamacpp";
-		checkpoint = `${model.id}${tagName}`;
+		const tagName = getQuantTag(filepath);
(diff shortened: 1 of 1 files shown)


julien-c (approved):
Cool, this looks like a simpler UX!

jeremyfowers:
Is there anything I need to do to get that test passing?

packages/tasks/src/local-apps.ts
```diff
@@ -513,43 +513,39 @@ const snippetDockerModelRunner = (model: ModelData, filepath?: string): string =
...
+	if (isRyzenAI) {
+		pullArg = model.id;
+		runName = `user.${modelName}`;
+		requirements = " (requires XDNA 2 NPU)";
 	} else {
-		recipe = "llamacpp";
-		checkpoint = `${model.id}${tagName}`;
+		const tagName = getQuantTag(filepath);
+		pullArg = `${model.id}${tagName}`;
+		runName = `user.${modelName}${tagName.replace(":", "-")}`;
```
pcuenca, line 532: So pull would be something like:

```bash
lemonade pull unsloth/gemma-4-26B-A4B-it-GGUF:Q8_0
```

but run would be:

```bash
lemonade run user.gemma-4-26B-A4B-it-GGUF-Q8_0
```

Are models pulled from HF moved to the `user.` namespace? Or should `user.` be inferred from the model Id rather than a constant in the snippet above? I quickly checked [the docs](https://lemonade-server.ai/docs/lemonade-cli/#options-for-load) but saw no examples of using `load` or `run` with `user.*`. 

What happens if the user pulls two different gguf files with the same name from different repos?

pcuenca:
> Is there anything I need to do to get that test passing?

The test failure is unrelated, sorry for the confusion!

packages/tasks/src/local-apps.ts
jeremyfowers: > So pull would be something like:
> lemonade pull unsloth/gemma-4-26B-A4B-it-GGUF:Q8_0
> but run would be:
> lemonade run user.gemma-4-26B-A4B-it-GGUF-Q8_0

Yes

> Are models pulled from HF moved to the user. namespace?

Lemonade adds the `user.` prefix to any "custom" models the user is pulling that aren't in the built-in models list. This guarantees no naming collision with the built-in list.

We curate the built-in list for our users, you can check it out here: https://lemonade-server.ai/models.html

> What happens if the user pulls two different gguf files with the same name from different repos?

This should generate an error a pull time due to a naming collision.

I think I see where you're going with this... it would be nice if `lemonade` would just resolve `lemonade run unsloth/gemma-4-26B-A4B-it-GGUF:Q8_0`. And if we had some logic to gracefully handle collisions on custom pulls. Could do that in a future Lemonade update?

pcuenca: > it would be nice if lemonade would just resolve `lemonade run unsloth/gemma-4-26B-A4B-it-GGUF:Q8_0`

Exactly, that'd be awesome!

> Could do that in a future Lemonade update?

Certainly! This is already an improvement (thanks!) and good to merge for me, cc @krampstudio as I think he reviewed your previous PRs.

pcuenca (approved):

packages/tasks/src/local-apps.ts
jeremyfowers: Awesome, thanks very much @pcuenca !

krampstudio (approved):
Cool, this will look like this @jeremyfowers 

Image

jeremyfowers:
Thanks all!

PR #2118, opened 2026-04-22
fix "browser" test

krampstudio:
> [!NOTE]
> **Medium Risk**
> Touches `WebBlob.create` header parsing and test fetch behavior; failures would affect streamed/ranged downloads, but changes are small and add validation to reduce silent misbehavior.
> 
> **Overview**
> Fixes browser-side `WebBlob` creation when `Content-Length` is missing or not usable by **validating the `content-length` header** and throwing a clearer error (including `x-` headers) instead of silently producing an invalid size.
> 
> Updates `WebBlob` tests to avoid relying on CORS-exposed `HEAD` headers by deriving size from a `GET` body and injecting a custom `fetch` that supplies `content-length`/`accept-ranges` for the `HEAD` request. Adds a root `lint-staged` config to run `oxfmt` on staged `*.{cjs,ts}` files.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 015531ae7d619b5747c96ad181d25b7f2b8d8135. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


coyotte508 (commented):
this fixes the test, but does the problem happen in prod or not?

From a different origin

krampstudio:
> this fixes the test, but does the problem happen in prod or not?
> 
> From a different origin

My understanding was that it has already a fallback to a `GET` in prod when the `HEAD` was failing or the Content-Length at 0: 
https://github.com/huggingface/huggingface.js/blob/1b32a04620cfab1d074a265a03869e19ed92201d/packages/hub/src/utils/WebBlob.ts#L34-L41

(I will prevent a NaN here btw to be 100% sure)

coyotte508:
> > this fixes the test, but does the problem happen in prod or not?
> > From a different origin
> 
> My understanding was that it has already a fallback to a `GET` in prod when the `HEAD` was failing or the Content-Length at 0:
> 
> https://github.com/huggingface/huggingface.js/blob/1b32a04620cfab1d074a265a03869e19ed92201d/packages/hub/src/utils/WebBlob.ts#L34-L41
> 
> (I will prevent a NaN here btw to be 100% sure)

Ideally though, to avoid downloading the whole blob in memory, CORS issue should be fixed / mitigated in a better way

(finding the root cause is already very nice)

krampstudio:
> Ideally though, to avoid downloading the whole blob in memory, CORS issue should be fixed / mitigated in a better way
> 
> (finding the root cause is already very nice)

_back on this_

The only solution I see is to always use GET and cancel reading the body if the server send the accept-ranges header. The downside is the cancel that can be late and we'll start receiving data.  
Something like this:

```ts
  const size = Number(response.headers.get("content-length") || 0);
  const contentType = response.headers.get("content-type") || "";
  const supportRange = response.headers.get("accept-ranges") === "bytes";

  if (supportRange && size >= (opts?.cacheBelow ?? 1_000_000)) {
    await response.body?.cancel();
    return new WebBlob(url, 0, size, contentType, true, customFetch, opts?.accessToken);
  }

  return response.blob();
```

coyotte508 (approved):

PR #2120, opened 2026-04-24
Add model download stats for encoderfile

besaleli:
This PR adds download stats tracking for encoderfile, a framework that packages encoder transformers into single binary executables. Hoping I added everything I needed to! You can learn more about the project here: https://github.com/mozilla-ai/encoderfile.

Thanks from Mozilla.ai! ❤️🤗

---

> [!NOTE]
> **Low Risk**
> Low risk: adds a single new model-library entry and an ElasticSearch query for download counting, with no changes to core logic or security-sensitive code.
> 
> **Overview**
> Adds `encoderfile` to `MODEL_LIBRARIES_UI_ELEMENTS`, including repo metadata and a `countDownloads` ElasticSearch query (`path_extension:"encoderfile"`) so Hub download stats can be tracked for encoderfile-packaged models.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit acb532c87f8afcdc468d692d73ac8dc4f2a95398. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/tasks/src/model-libraries.ts
+++ b/packages/tasks/src/model-libraries.ts
@@ -434,6 +434,13 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		filter: false,
 		countDownloads: `path_filename:"slicex_elm_config" AND path_extension:"json"`,
 	},
+	encoderfile: {
+		prettyLabel: "encoderfile",
+		repoName: "encoderfile",
+		repoUrl: "https://github.com/mozilla-ai/encoderfile",
+		filter: false,
+		countDownloads: `path_extension: "encoderfile"`,
+	},
 	espnet: {
 		prettyLabel: "ESPnet",
 		repoName: "ESPnet",
--- a/packages/tasks/src/model-libraries.ts
+++ b/packages/tasks/src/model-libraries.ts
@@ -439,7 +439,7 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		repoName: "encoderfile",
 		repoUrl: "https://github.com/mozilla-ai/encoderfile",
 		filter: false,
-		countDownloads: `path_extension: "encoderfile"`,
+		countDownloads: `path_extension:"encoderfile"`,
 	},
 	espnet: {
 		prettyLabel: "ESPnet",


Wauplin:
Hey, thanks for your PR! Review below is AI-generated but perfectly accurate (reviewed it myself). Biggest isssue is related to how downloads will be counted. Currently you'll get a single download counter for all models which is suboptimal. Best solution is to split in several repos. Let me know what you prefer to do :)

---

Thanks for the PR! 🤗

The entry itself is well-formed (alphabetical placement is correct, public GitHub repo, clean `countDownloads` rule). A few things worth discussing before we merge, though — mostly around how the models are laid out on the Hub:

1. **One Hub repo containing many models** — right now https://huggingface.co/mozilla-ai/encoderfile is a single repo holding distinct models (`DuoGuard/`, `JasperLS/`, `deepset/`, `protectai/`, `qualifire/`, `sentence-transformers/…`) as subdirectories. The Hub's library integration is designed around one repo per model: the filter page, the "models using this library" card, and the download stats all aggregate at the repo level. With the current layout, https://huggingface.co/models?other=encoderfile will always show exactly one tile regardless of how many underlying models you ship, and per-model stats aren't recoverable. Strong recommendation to split each subdirectory into its own repo under `mozilla-ai/…` (e.g. `mozilla-ai/DuoGuard-0.5B-encoderfile`) before merging — this is the single biggest thing that will make the integration actually useful to users.

2. **Use `library_name: encoderfile` in the modelcard metadata** — the current card on `mozilla-ai/encoderfile` sets `tags: [encoderfile]`, which is why it shows up in the filter, but the canonical way to register a model with a Hub library is the `library_name` key in the YAML block at the top of the README. This is what the library badge on the model page keys off. See https://huggingface.co/docs/hub/models-adding-libraries. Easy to add when you (hopefully) split the repos.

3. **`countDownloads` — worth sanity-checking** — each model ships 4 per-arch `.encoderfile` binaries (`aarch64-apple-darwin`, `aarch64-linux-gnu`, `x86_64-apple-darwin`, `x86_64-linux-gnu`). If a typical user downloads one binary for their platform, the current rule `path_extension:"encoderfile"` is accurate. If anything downloads all four (mirrors, CI, the encoderfile CLI itself pulling all variants), one "install" would register as 4 downloads. If the latter is common, consider something more unique per model, e.g. a small manifest file that every consumer reads exactly once. If the former is the usual case, the current rule is fine as-is.

Happy to merge once the repo layout is sorted (item 1) — the other two are easy follow-ups. The download count will be retroactive, so there's no urgency from a stats perspective. Let me know what you think!

---
_This review was drafted by Claude based on the PR diff and a fetch of the referenced Hub repos, then reviewed by a human before posting. If a specific point doesn't match reality for your library, please push back — the automation can be wrong._

besaleli:
@Wauplin Thanks so much for the quick response! More than happy to split into separate repositories. We will also do the same for our official Llamafile distributions (happy to add that to the model libraries as well! Just let me know if you want me to do that in a separate PR).

For point 2, happy to do that as well. For 3, we expect people to only download one model as encoderfiles are built per-platform, but we can provide the `encoderfile.yml` manifests used to build them as well.

I'll follow up with another comment once that's done!

besaleli:
Hi @Wauplin, thanks for your patience on this! We've deprecated the old encoderfiles repo and split into separate ones (they now live in this collection: https://huggingface.co/collections/mozilla-ai/encoderfiles). We've added `library_name` to all of the model cards. Just let me know if there's anything else you need! :)

Wauplin (approved):
That's really great! Thanks for splitting up the repos, it's definitely the best solution to get good analytics. I'll merge this PR now. Expect 2-3 days before getting it live on the Hub :)

PR #2126, opened 2026-04-25
Add time-series-forecasting task schema

pjhul:
Scaffolds the new `time-series-forecasting` task with input/output JSON schemas, generated TypeScript types, task metadata, and registry wiring. `pnpm check`, `test`, and `lint:check` on the tasks package all pass locally.

## Design decisions worth flagging

1. **`series` is a list of structured series objects**, not a string or binary payload. This diverges from existing HF tasks but mirrors how mature time-series libraries represent a batch: GluonTS datasets are iterables of `{target, start, …}` entries, Darts uses `List[TimeSeries]`, and Chronos takes a list of series. Parallel-array encodings don't survive contact with ragged lengths or per-series optional fields (timestamps, covariates, metadata).

2. **`target` is always 2D `[num_timesteps][num_channels]`.** Univariate series use a length-1 channel axis. This is a deliberate normalization, not a shape any single library uses — GluonTS/DeepAR are 1D for univariate, Darts is 3D `(time, component, sample)`, and Chronos-1 is univariate-1D (**Chronos-2** is the one that adds multivariate). Fixing `target` at 2D avoids `oneOf` 1D/2D ambiguity at length-1 inputs and keeps output indexing uniform (`out[t][c]`).

3. **Missing observations are `null` inline** inside `target`, not a separate `observed_mask`. Matches pandas, R, SQL, Arrow convention. HF's TimeSeriesTransformer uses `past_observed_mask` because PyTorch tensors can't hold null — but for API consumers dealing with a separate mask is much more cumbersome than in-line null's.

4. **Input uses `start` + `parameters.frequency`** rather than an explicit timestamps array. Matches the Darts `TimeSeries` model.

5. **`quantile_predictions` is an array of `{level, values}` objects** rather than a dict keyed by string-float. Matches HF's pattern for enumerated scored outputs (`text-generation` logprobs, classification scores, fill-mask tokens). AWS Chronos uses a dict-keyed-by-string; we diverge intentionally for HF-ecosystem consistency.

6. **Three uncertainty channels on output** — `mean` (required), `quantile_predictions`, `samples`. All optional; they may coexist. No parametric distribution parameters. `samples` is the universal fallback for any distributional output.

## Validation

- [x] `pnpm --filter tasks-gen inference-codegen` regenerates clean `inference.ts` from the JSON schemas
- [x] `pnpm --filter @huggingface/tasks check` (tsc)
- [x] `pnpm --filter @huggingface/tasks test`
- [x] `pnpm --filter @huggingface/tasks lint:check`
- [x] Registered in `packages/tasks/src/tasks/index.ts`

---

> [!NOTE]
> **Low Risk**
> Additive task-package schemas and registry wiring only; no auth, runtime inference, or existing task behavior changes.
> 
> **Overview**
> Introduces a full **`time-series-forecasting`** task definition in `@huggingface/tasks`: JSON Schema input/output specs, codegen’d TypeScript inference types, task metadata (`data.ts`), and docs (`about.md`).
> 
> **Registry:** `time-series-forecasting` moves from a placeholder (`undefined`) to a real task page via `getData(...)`, with public exports for `TimeSeriesForecastingInput` / `Output` and related types.
> 
> **API shape (new contract):** Requests use a **`series`** array of structured objects (not a flat tensor). Each item requires **`target`** as 2D `[timesteps][channels]` (univariate = single channel per step), optional **`start`**, covariates, and echoed **`metadata`**. Parameters cover **`prediction_length`**, **`frequency`**, **`quantile_levels`**, **`num_samples`**, and **`seed`**. Responses are **`outputs`** aligned 1:1 with input order: required **`mean`**, optional **`quantile_predictions`** (`{level, values}`), **`samples`**, and **`timestamps`** when `start` + `frequency` are set.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 93170928f0dac158597838858223235f473a7919. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/tasks/src/tasks/time-series-forecasting/spec/input.json
```diff
@@ -0,0 +1,93 @@
...
+					"type": "array",
+					"items": {
+						"type": "array",
+						"items": { "type": ["number", "null"] }
+					}
+				},
+				"start": {
+					"description": "Optional ISO-8601 timestamp of the first observation. Combined with `parameters.frequency` it defines the time index for both the input and the returned forecast timestamps. If omitted, observations are treated as an integer sequence and the response has no `timestamps`.",
+					"type": "string",
+					"format": "date-time"
```
hanouticelina: `format: "date-time"` makes quicktype generate `start?: Date` but but the API will return raw JSON and `@huggingface/inference`  does not hydrate response fields into `Date` objects. let's make it plain string instead
```suggestion
```

packages/tasks/src/tasks/time-series-forecasting/spec/output.json
```diff
@@ -0,0 +1,78 @@
...
+							"items": { "type": "number" }
+						}
+					}
+				},
+				"timestamps": {
+					"description": "Optional ISO-8601 timestamps for each forecast step. Present when the input item had `start` and `parameters.frequency` was supplied.",
+					"type": "array",
+					"items": {
+						"type": "string",
+						"format": "date-time"
```
hanouticelina: same comment as https://github.com/huggingface/huggingface.js/pull/2126/changes#r3255296460
```suggestion
```

packages/tasks/src/tasks/time-series-forecasting/spec/input.json
```diff
@@ -0,0 +1,93 @@
...
+			"type": "object",
+			"required": ["target"],
+			"properties": {
+				"target": {
+					"description": "Historical values on a regular grid. Always a 2D array of shape [num_timesteps][num_channels]. Univariate series use a length-1 channel axis. Missing observations are encoded as `null`.",
+					"type": "array",
+					"items": {
+						"type": "array",
+						"items": { "type": ["number", "null"] }
+					}
```
hanouticelina, line 31: the schema currently accepts `target: []`, `target: [[]]`, is that intentional? 
```suggestion
					"type": "array",
					"minItems": 1,
					"items": {
						"type": "array",
						"minItems": 1,
						"items": { "type": ["number", "null"] }
					}
```

packages/tasks/src/tasks/time-series-forecasting/spec/output.json
```diff
@@ -0,0 +1,78 @@
...
+			"type": "object",
+			"required": ["mean"],
+			"properties": {
+				"mean": {
+					"description": "Point forecast. Same shape as the corresponding input's `target`: outer dimension is time (length = prediction_length), inner dimension is channels.",
+					"type": "array",
+					"items": {
+						"type": "array",
+						"items": { "type": "number" }
+					}
```
hanouticelina, line 29: empty array is expected here? if not, we should add `"minItems": 1`
```suggestion
				"mean": {
					"description": "Point forecast. Same shape as the corresponding input's `target`: outer dimension is time (length = prediction_length), inner dimension is channels.",
					"type": "array",
					"minItems": 1,
					"items": {
						"type": "array",
						"minItems": 1,
						"items": { "type": "number" }
					}
```

packages/tasks/src/tasks/time-series-forecasting/data.ts
```diff
@@ -0,0 +1,87 @@
...
+			description:
+				"The GluonTS reference datasets for electricity load, traffic, and exchange rates at multiple frequencies.",
+			id: "autogluon/chronos_datasets",
+		},
+	],
+	demo: {
+		inputs: [
+			{
+				label: "Input series (univariate, 10 observations)",
+				content: "[120, 135, 142, 158, 163, 170, 185, 190, 198, 210]",
```
hanouticelina: The PR description says that `target` is always 2D but the demo embedded in here is 1D

packages/tasks/src/tasks/time-series-forecasting/data.ts
```diff
@@ -0,0 +1,87 @@
+import type { TaskDataCustom } from "../index.js";
+
+const taskData: TaskDataCustom = {
+	datasets: [
+		{
+			description:
+				"A curated collection of ~30 time-series datasets from diverse domains used to pretrain and evaluate foundation models.",
+			id: "monash_tsf",
```
hanouticelina: please verify that the datasets id are correct 
```suggestion
			id: "Monash-University/monash_tsf",
```

packages/tasks/src/tasks/time-series-forecasting/data.ts
```diff
@@ -0,0 +1,87 @@
...
+		},
+		{
+			description:
+				"A large-scale benchmark for evaluating probabilistic time-series forecasters across many domains and frequencies.",
+			id: "Salesforce/gift-eval",
```
hanouticelina:
```suggestion
			id: "Salesforce/GiftEval",
```

packages/tasks/src/tasks/time-series-forecasting/spec/input.json
```diff
@@ -0,0 +1,93 @@
...
+				},
+				"quantile_levels": {
+					"type": "array",
+					"items": {
+						"type": "number",
+						"exclusiveMinimum": 0,
+						"exclusiveMaximum": 1
+					},
+					"description": "Quantile levels in (0, 1) to return alongside the mean. Omit to receive only the mean."
+				},
```
hanouticelina, line 86: I think quantile levels are floats with semantically meaningful ordering and uniqueness. if it's the case, should we require unique quantile levels here?

packages/tasks/src/tasks/time-series-forecasting/spec/input.json
```diff
@@ -0,0 +1,93 @@
+{
+	"$id": "/inference/schemas/time-series-forecasting/input.json",
+	"$schema": "http://json-schema.org/draft-06/schema#",
+	"title": "TimeSeriesForecastingInput",
+	"description": "Inputs for Time Series Forecasting inference. Each input series lies on a regular grid defined by its `start` and `parameters.frequency`. Missing observations are encoded as `null` inside `target`; clients with irregular data should resample before calling.",
+	"type": "object",
+	"required": ["inputs"],
```
hanouticelina: every other task uses `inputs` as the data, not a batch wrapper so I'd be in favor to rename it to `series` or any other name that signals "this is a batch container and each element is one series"

packages/tasks/src/tasks/time-series-forecasting/spec/input.json
```diff
@@ -0,0 +1,93 @@
...
+					"type": "object",
+					"additionalProperties": { "type": "number" }
+				},
+				"metadata": {
+					"description": "Opaque per-series metadata. Echoed verbatim in the corresponding output item. Not consumed by any model. Maximum 4KB per item.",
+					"type": "object"
+				}
+			}
+		},
+		"TimeSeriesForecastingParameters": {
```
hanouticelina, line 64: we should probably add a `seed` parameter here to make samples reproducible across retries

hanouticelina (commented):
Hi @pjhul, did a first pass to the PR, thanks a lot for working at this! 

could you share links to the schemas/docs you used as reference?
When I asked an AI agent to cross-check, some of the claims didn't quite hold against current docs:

  - Darts is 3D, not 2D.  docs say "Return a 3-D array of dimension (time,
  component, sample)"
  - GluonTS is 1D for univariate and 2D `(dim, num_time_steps)` for multivariate
  - **Chronos-1** is 1D-only, `predict` accepts "a 1D tensor, or a list of 1D
  tensors, or a 2D tensor whose first dimension is batch"* Chronos-2 adds
  multivariate, so if you used Chronos-2 the prose just needs narrowing.
  - DeepAR is 1D with `"NaN"` strings: `"target": [4.0, 10.0, "NaN", 100.0,
  113.0]`

  want to make sure it's comparing against the same references you used.

packages/tasks/src/tasks/time-series-forecasting/about.md
```diff
@@ -0,0 +1,86 @@
...
+
+for output in result["outputs"]:
+    print("mean:", [row[0] for row in output["mean"]])
+    for qp in output.get("quantile_predictions", []):
+        print(f"  q{qp['level']}:", [row[0] for row in qp["values"]])
+```
+
+## Direct Inference
+
+You can also run time series foundation models locally using their native libraries. Examples include `chronos-forecasting`, `uni2ts` (Moirai), `timesfm`, `lag-llama`, `darts`, and `gluonts`. Most of these follow the input shape described above — a 2D array of observations per series — with library-specific helpers for wrapping pandas DataFrames.
```
julien-c: mention transfomers (@kashif) and transformers.js (@xenova) as well, no?

packages/tasks/src/tasks/time-series-forecasting/spec/input.json
pjhul: Got it - just removed these two lines

packages/tasks/src/tasks/time-series-forecasting/spec/input.json
pjhul: Great catch - added that validation

packages/tasks/src/tasks/time-series-forecasting/about.md
pjhul: Great flag - added a reference to those libs

pjhul:
Hey @hanouticelina!

Thank you for the detailed review 🙏 I think I addressed all the comments you left on the PR, but let me know if I missed anything.

In regards to your comment, happy to expand more here! 

> Darts is 3D, not 2D. docs say "Return a 3-D array of dimension (time,
component, sample)"

Yup this is true - I should have been more specific here when I referenced Darts. The reason for the third axis is to support stochastic time series [[ref](https://unit8co.github.io/darts/generated_api/darts.timeseries.html#darts.timeseries.TimeSeries.is_stochastic)]. While this makes sense for time series output, it doesn't really apply for input time series. Since we already have an output structure to wrap quantiles/samples there wasn't a need for this axis. 

When calling `.values()` the return is a 2-D (time, component) array [[ref](https://unit8co.github.io/darts/generated_api/darts.timeseries.html#darts.timeseries.TimeSeries.values)].

> GluonTS is 1D for univariate and 2D (dim, num_time_steps) for multivariate

Great callout - this was actually something I went back and forth on: whether or not to use a 1D array for univariate. I eventually landed on 2D for everything to avoid the need for consumers to deal with 'polymorphic' return types. Basically to avoid calling something like `Array.isArray(series[0])` to narrow types.

Honestly I'd love to get your take on how big of a deal something like is in your experience. There's also the counter argument that needing to marshal 1D series into a 2D array is painful, so I could go either way here.

> Chronos-1 is 1D-only, predict accepts "a 1D tensor, or a list of 1D
tensors, or a 2D tensor whose first dimension is batch"* Chronos-2 adds
multivariate, so if you used Chronos-2 the prose just needs narrowing.

Great flag - I updated that reference. 

> DeepAR is 1D with "NaN" strings: "target": [4.0, 10.0, "NaN", 100.0,   113.0]

Yup went back and forth a bit on this one as well. Unfortunately because `NaN` isn't a valid token in strict JSON that didn't seem like an option. I decided to go with `null` instead, once again to avoid the polymorphic array element type. Along with that, I couldn't think of a need to distinguish between `NaN` and `null` as they both typically represent empty/missing values. 

Hopefully this helps out and please let me know if you have any follow up questions!

kashif:
yes all seem like sensible choices. one can think of univariate a multivariate with just 1 time series so that abstraction also works out I believe. I am checking in regards to the probabilisitic forecasts and quantile based forecasts next and will report back

packages/tasks/src/tasks/time-series-forecasting/data.ts
```diff
@@ -0,0 +1,88 @@
...
+	metrics: [
+		{
+			description:
+				"Mean Absolute Scaled Error, which scales MAE by an in-sample naive forecast so it is comparable across series with different magnitudes.",
+			id: "mase",
+		},
+		{
+			description:
+				"Continuous Ranked Probability Score, which evaluates probabilistic forecasts by measuring the difference between forecast and observation CDFs.",
+			id: "crps",
```
kashif, line 47: Minor: crps and wql end up being essentially the same number once you're scoring at fixed quantile levels. GIFT-Eval pairs MASE with MSIS as the interval metric — might be worth msis here instead so the three aren't redundant. No strong opinion though.

packages/tasks/src/tasks/time-series-forecasting/spec/input.json
```diff
@@ -0,0 +1,99 @@
...
+				},
+				"future_covariates": {
+					"description": "Optional named covariates known over both the historical window and the forecast horizon. Each key maps to a 1D array spanning the full timeline: the first `num_timesteps` values align 1:1 with `target`, followed by `parameters.prediction_length` values over the horizon (total length `num_timesteps + prediction_length`). Missing values in the historical portion are encoded as `null`.",
+					"type": "object",
+					"additionalProperties": {
+						"type": "array",
+						"items": { "type": ["number", "null"] }
+					}
+				},
+				"static_covariates": {
```
kashif, line 53: Should this allow categorical statics too, or is numeric-only intentional for v1? `transformers`' own TimeSeriesTransformer splits these into static_categorical + static_real, and item/store-id type features are pretty common. Totally fine to punt — just flagging it has nowhere to go right now.

packages/tasks/src/tasks/time-series-forecasting/spec/input.json
```diff
@@ -0,0 +1,99 @@
...
+				"past_covariates": {
+					"description": "Optional named covariates observed over the historical window. Each key maps to a 1D array aligned 1:1 with the time axis of `target`. Missing values are encoded as `null`. Models that don't support covariates silently ignore this field.",
+					"type": "object",
+					"additionalProperties": {
+						"type": "array",
+						"items": { "type": ["number", "null"] }
+					}
+				},
+				"future_covariates": {
+					"description": "Optional named covariates known over both the historical window and the forecast horizon. Each key maps to a 1D array spanning the full timeline: the first `num_timesteps` values align 1:1 with `target`, followed by `parameters.prediction_length` values over the horizon (total length `num_timesteps + prediction_length`). Missing values in the historical portion are encoded as `null`.",
```
kashif, line 46: Small ambiguity: if prediction_length is omitted (native horizon), the client can't tell how long the future portion here needs to be. Maybe require prediction_length when future_covariates is present?

packages/tasks/src/tasks/time-series-forecasting/spec/output.json
```diff
@@ -0,0 +1,85 @@
...
+				},
+				"quantile_predictions": {
+					"description": "Optional per-quantile forecasts, sorted ascending by level. Each entry has the same shape as `mean`.",
+					"type": "array",
+					"items": {
+						"type": "object",
+						"required": ["level", "values"],
+						"properties": {
+							"level": {
```
kashif, line 38: Nice — numeric level reads cleaner than the stringified-float keys gluonts/chronos use internally. Avoids the 0.1 → "0.1" round-trip.

PR #2134, opened 2026-04-29
[Hub] Dedupe file entries by xet hash within a shard

assafvayner:
## Summary
- Track xet file hashes already written into the current shard's file-info section and skip duplicates so the shard's file list behaves like a set rather than a list.
- The dedup set is scoped to the entire `uploadShards` call so it persists across shard flushes — a duplicate in a later shard is also skipped.
- File events are still yielded to callers for every input path, so consumers (e.g. `commit.ts`) keep getting per-path progress and `path -> xet hash` mapping for duplicates.

Today the only dedup that happens is sha256-based and lives in `createXorbs` — if a file has no sha256 (or only one of the duplicates does), both entries currently get written into the shard's file list. This change handles that case.

## Test plan
- [x] `npx vitest run src/utils/uploadShards.spec.ts` — added a case that feeds three paths through `uploadShards` with the mocked `createXorbs` yielding the same xet hash, and asserts: each path still gets a `file` event, but only one file entry ends up in the resulting shard.

---

> [!NOTE]
> **Medium Risk**
> Changes shard serialization behavior by skipping duplicate file entries based on Xet hash, which could impact upload correctness and downstream consumers if the dedupe scope/ordering assumptions are wrong.
> 
> **Overview**
> Makes `uploadShards` treat the shard file-info section like a set by tracking seen Xet file hashes and **skipping writing duplicate file entries** while still yielding `file` events for every input path.
> 
> Adds a new unit test that feeds multiple paths producing the same Xet hash and asserts only one file entry is stored in the uploaded shard, while all paths still receive `file` events.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 24884a173dd1a304ccfcc94d48e051b62a83fa6a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/hub/src/utils/uploadShards.spec.ts
+++ b/packages/hub/src/utils/uploadShards.spec.ts
@@ -91,6 +91,17 @@ function toSource(sha256?: string): AsyncGenerator<{ content: Blob; path: string
 	})();
 }
 
+function toMultiSource(paths: string[]): AsyncGenerator<{ content: Blob; path: string; sha256?: string }> {
+	return (async function* () {
+		for (const path of paths) {
+			yield {
+				content: new Blob(["content"]),
+				path,
+			};
+		}
+	})();
+}
+
 describe("uploadShards", () => {
 	it("omits metadata flag and metadata section when sha256 is missing", async () => {
 		const uploadedShards: Uint8Array[] = [];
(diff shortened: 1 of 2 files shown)


coyotte508:
Thanks!

To offer more context for future us, the root issue is that for buckets, there's no sha256 to dedup on, so the JS lib can send duplicate files in a shard call. That's why duplicate on xetHashes is needed in addition

PR #2140, opened 2026-05-05
fix: support numeric member access in jinja parser (obj.0 -> obj[0])

koush:
Python Jinja2 supports numeric member access like obj.0 (equivalent to obj[0]), but the JS implementation only allowed Identifier after a dot. This fixes parsing of GLM-5.1 chat templates that use m.content.0.type.

https://huggingface.co/zai-org/GLM-5.1/blob/main/chat_template.jinja#L97

---

> [!NOTE]
> **Medium Risk**
> Changes core parsing of member expressions; while small, it can alter how dotted expressions are interpreted and could impact template parsing edge cases.
> 
> **Overview**
> Adds support for Jinja-style numeric member access in `parseMemberExpression`, allowing `obj.0` to parse as a computed member lookup equivalent to `obj[0]`.
> 
> Non-numeric dot access behavior remains the same, but dot access now accepts an `IntegerLiteral` after `.` and immediately emits a computed `MemberExpression` for that segment.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1b316c12a3f17dcae004f506d764d48a0640fb36. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/jinja/src/parser.ts
+++ b/packages/jinja/src/parser.ts
@@ -541,10 +541,17 @@ export function parse(tokens: Token[]): Program {
 				// computed (i.e., bracket notation: obj[expr])
 				property = parseMemberExpressionArgumentsList();
 				expect(TOKEN_TYPES.CloseSquareBracket, "Expected closing square bracket");
-			} else {
+				} else {
 				// non-computed (i.e., dot notation: obj.expr)
-				property = parsePrimaryExpression(); // should be an identifier
-				if (property.type !== "Identifier") {
+				// Python Jinja2 also allows numeric access like obj.0 (equivalent to obj[0])
+				property = parsePrimaryExpression();
+				if (property.type === "Identifier") {
+					// Standard dot access: obj.prop
+				} else if (property.type === "IntegerLiteral") {
+					// Numeric dot access: obj.0 -> obj[0] (computed)
+					object = new MemberExpression(object, property, true);
+					continue;
+				} else {
 					throw new SyntaxError(`Expected identifier following dot operator`);
 				}
 			}


xenova:
Hi there 👋 thanks for the PR. As pointed out [here](https://github.com/huggingface/huggingface.js/pull/2140#discussion_r3189991440), I think we need some additional parsing logic for ensuring `obj.0.1` is not interpreted with `0.1` as a float.

Also, if you could add some unit tests, that would be great.

koush:
As it stands, FloatLiteral fails with the current code, and the making it work is going to be a bit of a pain. I do not think this is the appropriate place to do that is a more involved change.
Splitting on the period to parse the whole/fraction out, and then double indexing. Furthermore that change would not address multiple dot indexes either, which also affects the current code and this change. FloatLiteral parsing does not fix this, and will not fix chained numeric indexing.

The goal of this change is for single dot indexing to work as that covers basically most scenarios for chat template parsing.

xenova:
No worries. I added support for it in https://github.com/huggingface/huggingface.js/pull/2146. Let me know if you can review that PR 👍 I'll add you as a co-author.

The majority of that is just adding unit tests. The core fix is only `+11 -5` lines

xenova:
Thanks again for raising the issue and providing an initial fix. Commit made [here](https://github.com/huggingface/huggingface.js/commit/ee92cdbe6f08eddd24888204853c2510028c588e) 
image


new jinja.js release is also out too https://www.npmjs.com/package/@huggingface/jinja/v/0.5.9

PR #2143, opened 2026-05-06
Add Hermes Agent to local apps

gary149:
image

## Summary
- Add [Hermes Agent](https://hermes-agent.nousresearch.com/) as a local app for tool-calling models served via llama.cpp or MLX (same eligibility as Pi: GGUF or MLX, `conversational`, chat template references `tools`)
- Snippet uses `hermes config set` CLI commands (verified to work) instead of editing `~/.hermes/config.yaml` by hand
- Refactor: extract `isToolCallingLocalAgentModel` and `getLocalServerStep` so Pi and Hermes share the predicate and the server-start step (Pi behavior unchanged byte-for-byte)

## Test plan
- [x] `pnpm --filter @huggingface/tasks test` — 19/19 pass
- [x] Verified end-to-end with `unsloth/Qwen3.6-27B-GGUF:UD-Q6_K_XL` running on llama.cpp

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it changes the generated Pi configuration `model.id` for GGUF models (now includes the quant tag) and adds new snippet logic that affects what users run/configure, though it’s isolated to local-app guidance.
> 
> **Overview**
> Adds **Hermes Agent** as a new entry in `LOCAL_APPS`, available for the same tool-calling GGUF/MLX conversational models as Pi, and provides an install/config/run snippet using `hermes config set ...` against a local OpenAI-compatible server.
> 
> Refactors the Pi/Hermes flow by extracting a shared `isToolCallingLocalAgentModel` predicate and `getLocalServerStep` snippet generator, and updates Pi’s generated `models.json` to use the full GGUF model id including `:{{QUANT_TAG}}` (instead of the repo basename). Tests are expanded to cover Hermes (GGUF + MLX) and updated for the Pi id change.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c9c409d41d8a337ef25c68b276b702c59794d15b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/tasks/src/local-apps.spec.ts
+++ b/packages/tasks/src/local-apps.spec.ts
@@ -3,6 +3,10 @@ import { LOCAL_APPS } from "./local-apps.js";
 import type { ModelData } from "./model-data.js";
 
 describe("local-apps", () => {
+	it("shares the same model-page predicate for tool-calling local agents", async () => {
+		expect(LOCAL_APPS.pi.displayOnModelPage).toBe(LOCAL_APPS["hermes-agent"].displayOnModelPage);
+	});
+
 	it("llama.cpp conversational", async () => {
 		const { snippet: snippetFunc } = LOCAL_APPS["llama.cpp"];
 		const model: ModelData = {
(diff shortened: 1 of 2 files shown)


packages/tasks/src/local-apps.ts
```diff
@@ -510,6 +519,33 @@ const snippetPi = (model: ModelData, filepath?: string): LocalAppSnippet[] => {
 	];
 };
 
+const snippetHermesAgent = (model: ModelData, filepath?: string): LocalAppSnippet[] => {
+	const modelId = isMlxModel(model) ? model.id : `${model.id}${getQuantTag(filepath)}`;
```
pcuenca, line 523: Nice. Shouldn't we optionally do the same for `pi` btw? I have stuff like `"id": "unsloth/Qwen3.6-27B-GGUF:Q8_0"` in my `models.json`, but we are just using the `modelName` in pi:

Image

pcuenca (approved):
🙌

packages/tasks/src/local-apps.ts
gary149: updated

packages/tasks/src/local-apps.spec.ts
```diff
@@ -3,6 +3,10 @@ import { LOCAL_APPS } from "./local-apps.js";
 import type { ModelData } from "./model-data.js";
 
 describe("local-apps", () => {
+	it("shares the same model-page predicate for tool-calling local agents", async () => {
+		expect(LOCAL_APPS.pi.displayOnModelPage).toBe(LOCAL_APPS["hermes-agent"].displayOnModelPage);
+	});
```
julien-c:
```suggestion
```
no neeed imo

julien-c (approved):

teknium1 (approved):
Looks good from the Hermes side — verified the snippet end-to-end against current `main`:

- `hermes config set model.provider custom` / `model.base_url` / `model.default` are all valid config paths; `custom` is a recognized provider and is in the allowlist for vendor-prefixed model IDs, so `org/repo:QUANT_TAG` won't trigger a doctor warning.
- `curl | bash` installer URL responds 200, and `hermes setup` + bare `hermes` launch flow is correct.
- No changes needed on the hermes-agent side to support this.

Thanks for wiring us up alongside Pi!

Issue #2145, opened 2026-05-07
Adding radeon igpus to hardware list


BlivionIaG:
Hi, 
After looking at some comments online on recent hf hardware reports/stats, it was reported that radeon igpu users were not represented. Issue for radeon igpus (maybe it's the same for other brands, i am not familiar with them) is that the maximum usable memory for the igpu for lets say llama.cpp in vulkan mode can be variable depending on the ttm pages_limit set on the machine.
Exemple here for a 610m igpu by default set to total_ram_pages/2:
```shell
llama-cli --list-devices | grep -i amd 
  Vulkan0: AMD Radeon 610M (RADV RAPHAEL_MENDOCINO) (49166 MiB, 48600 MiB free)
```
Based on https://www.techpowerup.com/gpu-specs/radeon-610m.c3994
While we could say tflops is 0.97, memory can range from a lot of values (lets say 1 to almost 96GB).
What would be the best way to add them ? Power of 2 + half power of 2 increments like [16,24,32,48,64,96] ?

PR #2179, opened 2026-05-19
Add Radeon 610M hardware SKU

YBJ0000:
## Summary

- add Radeon 610M to the AMD GPU hardware compatibility list
- model it as an RDNA2 iGPU with shared-memory options for llama.cpp/Vulkan-style reporting

Refs #2145

## Tests

- pnpm --filter tasks check
- pnpm --filter tasks format:check
- pnpm --filter tasks lint:check
- pnpm --filter tasks test

---

> [!NOTE]
> **Low Risk**
> Low risk: adds a single new static AMD GPU/iGPU SKU entry with no behavioral or algorithmic changes.
> 
> **Overview**
> Adds **Radeon 610M** to the `AMD_GPU_SKUS` compatibility list in `hardware-amd.ts`, including TFLOPS, shared-memory size options, `gfxVersion` (`gfx1037`), MSRP, power, and release year so it can be recognized in hardware lookups.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c445d1b5b10b7cc072249ff766659db02092fa1f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/tasks/src/hardware-amd.ts
+++ b/packages/tasks/src/hardware-amd.ts
@@ -234,6 +234,14 @@ export const AMD_GPU_SKUS: Record<string, AmdGpuHardwareSpec> = {
 		power: 250,
 		releaseYear: 2020,
 	},
+	"Radeon 610M": {
+		tflops: 0.97,
+		memory: [16, 24, 32, 48, 64, 96],
+		gfxVersion: "gfx1037",
+		msrp: 300,
+		power: 15,
+		releaseYear: 2022,
+	},
 	"Ryzen AI Max+ 395": {
 		tflops: 59.4,
 		memory: [64, 96, 128],


BlivionIaG:
Hi, issue with that draft is that shared memory option range can change depending on the platform, i guess for rdna2 based igpus its somewhat the same. Also its gfx1037.

YBJ0000:
@BlivionIaG Thanks for the catch! I updated the Radeon 610M `gfxVersion` to `gfx1037` in the PR.

For the shared-memory options, I kept the current conservative range for now since it matches the existing hardware selector shape and can cover the platform-dependent llama.cpp/Vulkan reports reasonably well.

packages/tasks/src/hardware-amd.ts
```diff
@@ -9,6 +9,8 @@ export interface AmdGpuHardwareSpec extends HardwareSpec {
 	gfxVersion: string;
 }
 
+const SHARED_MEMORY_OPTIONS = [16, 24, 32, 48, 64, 96];
+
```
pcuenca:
```suggestion
```

packages/tasks/src/hardware-amd.ts
```diff
@@ -234,6 +236,14 @@ export const AMD_GPU_SKUS: Record<string, AmdGpuHardwareSpec> = {
 		power: 250,
 		releaseYear: 2020,
 	},
+	"Radeon 610M": {
+		tflops: 0.97,
+		memory: SHARED_MEMORY_OPTIONS,
```
pcuenca:
```suggestion
		memory: [16, 24, 32, 48, 64, 96],
```

packages/tasks/src/hardware.spec.ts
```diff
@@ -0,0 +1,15 @@
...
+		expect(SKUS.GPU.AMD["Radeon 610M"]).toEqual({
+			tflops: 0.97,
+			memory: [16, 24, 32, 48, 64, 96],
+			gfxVersion: "gfx1037",
+			msrp: 300,
+			power: 15,
+			releaseYear: 2022,
+		});
+	});
+});
```
pcuenca:
```suggestion
```

Thanks for being exhaustive, but we usually don't add tests for each new item.

pcuenca (commented):
Suggested to remove the `const` as it's only used in the new entry for now. We can revisit if additional shared-memory chips are added.

pcuenca:
Marked it "ready for review", is there anything else you were planning to add @YBJ0000?

packages/tasks/src/hardware-amd.ts
BlivionIaG: I can give a list of igpus, i think the one that would be the most interesting for llm use are probably the strix point mobile ones (HX 370 and stuff gfx1151), so method going forward is that kind of array ?

pcuenca: I think so, yes; the Strix Halo for instance has already been added as `[64, 96, 128]`. We can converge on something more standard if more of these pop up?

BlivionIaG: Perfect that answers #2145 :)

BlivionIaG: For Strix halo since its soldered i guess the reasoning was to add the possible max ram dispositions (actually there is also a 32GB config possible for 395+)

YBJ0000:
Thanks, addressed the suggestions: removed the single-use constant, inlined the memory options on the Radeon 610M entry, and dropped the per-SKU test file to match the existing pattern. I also updated the PR test plan accordingly.

pcuenca (approved):

pcuenca:
Thank you!

PR #2180, opened 2026-05-20
[inference] Use /v1/completions for hf-inference text-generation

oOraph:
context: https://huggingface.slack.com/archives/C0AA7DZFGVC/p1779282978979449?thread_ts=1779271635.539769&cid=C0AA7DZFGVC

HFInferenceTextGenerationTask now routes to /v1/completions with an OpenAI-compatible payload (prompt, max_tokens, …) instead of the legacy POST /models/{id} format with {inputs, parameters}.

This matches how HFInferenceConversationalTask already routes to /v1/chat/completions, and fixes text-generation models served via vLLM on hf-inference which only expose the OpenAI-compatible endpoints.

---

> [!NOTE]
> **Medium Risk**
> Changes the `hf-inference` text-generation request URL, payload shape, and response parsing, which could break compatibility with models/endpoints still expecting the legacy `POST /models/{id}` format. Conversational routing is also adjusted to rely on route building rather than custom URL logic.
> 
> **Overview**
> Updates the `hf-inference` provider’s **text generation** to call the OpenAI-compatible `models/{model}/v1/completions` endpoint, translating the payload from `{inputs, parameters}` to `{prompt, max_tokens, ...}` and mapping the new `{choices: [{text}]}` response back to `generated_text`.
> 
> Aligns **conversational** requests with the same route-based approach by routing to `models/{model}/v1/chat/completions`, removing bespoke URL construction logic.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c1c2d66cb8ef96410b4a709ed93e681023a1e49f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/inference/src/providers/hf-inference.ts
+++ b/packages/inference/src/providers/hf-inference.ts
@@ -27,6 +27,7 @@ import type {
 	SummarizationOutput,
 	TableQuestionAnsweringOutput,
 	TextClassificationOutput,
+	TextGenerationInput,
 	TextGenerationOutput,
 	TokenClassificationOutput,
 	TranslationOutput,
@@ -39,7 +40,6 @@ import { HF_ROUTER_URL } from "../config.js";
 import { InferenceClientInputError, InferenceClientProviderOutputError } from "../errors.js";
 import type { TabularClassificationOutput } from "../tasks/tabular/tabularClassification.js";
 import type { BodyParams, OutputType, RequestArgs, UrlParams } from "../types.js";
-import { toArray } from "../utils/toArray.js";
 import type {
 	AudioClassificationTaskHelper,
 	AudioToAudioTaskHelper,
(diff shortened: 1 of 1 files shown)


oOraph:
this is a breaking change but it should actually be ok:  HF-inference provider currently doesn't serve any non conversational text generation models so this is the right moment to change it and make it openAI compatible (so that for example, any vllm instance can be plugged behind)

(or do I miss some side effect like anything else using the modified file but HF inference provider ?)

packages/inference/src/providers/hf-inference.ts
```diff
@@ -210,14 +210,54 @@ export class HFInferenceConversationalTask extends HFInferenceTask implements Co
...
+interface HFInferenceTextCompletionOutput {
+	choices: Array<{ text: string }>;
+}
+
 export class HFInferenceTextGenerationTask extends HFInferenceTask implements TextGenerationTaskHelper {
-	override async getResponse(response: TextGenerationOutput | TextGenerationOutput[]): Promise<TextGenerationOutput> {
-		const res = toArray(response);
-		if (Array.isArray(res) && res.every((x) => "generated_text" in x && typeof x?.generated_text === "string")) {
-			return (res as TextGenerationOutput[])?.[0];
+	override makeUrl(params: UrlParams): string {
```
SBrandeis: Should probably override makeRoute instead

julien-c (commented):
nice, thanks for fixing this

packages/inference/src/providers/hf-inference.ts
oOraph: done (on both textGeneration and conversational sides, for consistency), edit: reverted, breaks some tests on conversational side

hanouticelina (approved):
thanks!

hanouticelina:
@oOraph `@huggingface/inference` patch released (https://github.com/huggingface/huggingface.js/actions/runs/26177823355/job/77012499058) with your changes. the version needs to be bumped in `moon-landing`

PR #2199, opened 2026-05-28
update longcat-video-avatar-1.5 to model libraries

lile839:
update longcat-video-avatar-1.5 to model libraries

---

> [!NOTE]
> **Low Risk**
> Single registry entry with no runtime or security-sensitive logic changes.
> 
> **Overview**
> Registers **LongCat-Video-Avatar-1.5** as a Hub modeling library in `MODEL_LIBRARIES_UI_ELEMENTS`, keyed by `longcat-video-avatar-1.5` (matching `library_name` on models).
> 
> The entry points to the [LongCat-Video](https://github.com/meituan-longcat/LongCat-Video) repo, keeps the library off the main models filter (`filter: false`), and defines download counting via `.safetensors` files or `config.json`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ac18cb1b5ca3657b6bee1267e3eb54f56801efae. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/tasks/src/model-libraries.ts
+++ b/packages/tasks/src/model-libraries.ts
@@ -808,6 +808,13 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		filter: false,
 		countDownloads: `path:"liveportrait/landmark.onnx"`,
 	},
+	"longcat-video-avatar-1.5": {
+		prettyLabel: "LongCat-Video-Avatar-1.5",
+		repoName: "LongCat-Video-Avatar-1.5",
+		repoUrl: "https://github.com/meituan-longcat/LongCat-Video",
+		filter: false,
+		countDownloads: `path_extension:"safetensors" OR path:"config.json"`,
+	},
 	"llama-cpp-python": {
 		prettyLabel: "llama-cpp-python",
 		repoName: "llama-cpp-python",


packages/tasks/src/model-libraries.ts
```diff
@@ -808,6 +808,13 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		filter: false,
 		countDownloads: `path:"liveportrait/landmark.onnx"`,
 	},
+	"longcat-video-avatar-1.5": {
+		prettyLabel: "LongCat-Video-Avatar-1.5",
+		repoName: "LongCat-Video-Avatar-1.5",
```
Wauplin, line 813:
```suggestion
	"longcat-video": {
		prettyLabel: "LongCat-Video",
		repoName: "LongCat-Video",
```

Only a suggestion but I feel that since the library is named LongCat Video (according to https://github.com/meituan-longcat/LongCat-Video), then the tag should be this instead of the exact repo name itself. This is only a suggestion, if you prefer `longcat-video-avatar` or even `longcat-video-avatar-1.5` then go for it (though I find them too long/specific). The important part is that you had

```yaml
library_name: longcat-video
```

in the model card metadata of the model. 

Since it's a diffusers model, you can also do 

```yaml
library_name: diffusers
tags:
  - longcat-video
```
or
```yaml
# preferred version if you want longcat video download count rule
library_name: longcat-video
tags:
  - diffusers
```

In both cases, both libraries will be detected but the `library_name` one will be considered as the "main library" which we will use to count download.

packages/tasks/src/model-libraries.ts
```diff
@@ -808,6 +808,13 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
...
+		repoUrl: "https://github.com/meituan-longcat/LongCat-Video",
+		filter: false,
+		countDownloads: `path_extension:"safetensors" OR path:"config.json"`,
```
Wauplin, line 816: Counting both safetensors and config files will lead to double-counting download if someone downloads the full repo. The best in this situation is to only count by `config.json`, which is the default rule. So my advice here is to remove this line and fallback to the default rule 


```suggestion
```

Wauplin (commented):
Thanks for the PR @lile839 ! Made a few comments to address in order to get things done correctly. Let me know if you have any questions 🤗

PR #2210, opened 2026-06-03
Add download stats filter for upcoming model release

chrisdonahue-goog:
Repo contains checkpoints for 3 different pipeline components of the model. This download filter only counts downloads to the core LLM component (in either MLX or Safetensors format)

---

> [!NOTE]
> **Low Risk**
> Single registry entry in `model-libraries.ts` with no auth, billing, or runtime behavior changes.
> 
> **Overview**
> Registers **Magenta RT 2** as a Hub model library (`magenta-realtime-2`) so models tagged with that `library_name` get correct UI labels and repo links.
> 
> Adds a **custom download counter** that only counts pulls of the core LLM weights (`mrt2_base` / `mrt2_small` in `.mlxfn` or `.safetensors`), excluding other Magenta RT 2 pipeline artifacts in the same repos.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 01917d470513a4b2a193d4cda1ebb24c956ce3e4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/tasks/src/model-libraries.ts
+++ b/packages/tasks/src/model-libraries.ts
@@ -838,6 +838,12 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		repoUrl: "https://github.com/magenta/magenta-realtime",
 		countDownloads: `path:"checkpoints/llm_base_x4286_c1860k.tar" OR path:"checkpoints/llm_large_x3047_c1860k.tar" OR path:"checkpoints/llm_large_x3047_c1860k/checkpoint"`,
 	},
+	"magenta-realtime-2": {
+		prettyLabel: "Magenta RT 2",
+		repoName: "Magenta RT 2",
+		repoUrl: "https://github.com/magenta/magenta-realtime",
+		countDownloads: `path:"models/mrt2_base/mrt2_base.mlxfn" OR path:"models/mrt2_small/mrt2_small.mlxfn" OR path:"checkpoints/mrt2_base.safetensors" OR path:"checkpoints/mrt2_small.safetensors"`,
+	},
 	"mamba-ssm": {
 		prettyLabel: "MambaSSM",
 		repoName: "MambaSSM",


packages/tasks/src/model-libraries.ts
```diff
@@ -838,6 +838,12 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		repoUrl: "https://github.com/magenta/magenta-realtime",
 		countDownloads: `path:"checkpoints/llm_base_x4286_c1860k.tar" OR path:"checkpoints/llm_large_x3047_c1860k.tar" OR path:"checkpoints/llm_large_x3047_c1860k/checkpoint"`,
 	},
+	"magenta-realtime-2": {
+		prettyLabel: "Magenta RT 2",
+		repoName: "Magenta RT 2",
+		repoUrl: "https://github.com/magenta/magenta-realtime",
+		countDownloads: `path:"models/mrt2_base/mrt2_base.mlxfn" OR path:"models/mrt2_small/mrt2_small.mlxfn" OR path:"checkpoints/mrt2_base.safetensors" OR path:"checkpoints/mrt2_small.safetensors"`,
```
pcuenca, line 845: Without knowing the exact details of the new checkpoints, I'm assuming it contains various types of checkpoints for different engines / sizes and they are mutually exclusive, is that right? Otherwise, if the inference code downloads multiple copies, you'd be counting multiple downloads.

pcuenca (approved):
In general, our recommendation is to split model versions across different repos so usage tracking is more granular, download counting rules are easier, model cards and snippets are more targeted.

Approving to unblock.

pcuenca:
Context ([internal link](https://huggingface.slack.com/archives/C06AF1RUV8F/p1780429160125239)) cc @Wauplin

packages/tasks/src/model-libraries.ts
Wauplin: agree this feels very specific and prone to double-count download. Is there a single-file in the repo that is mandatory to download to be able to use the model? That would be the best so we count downloads only on that file

Wauplin (approved):
Except for the download count rule, PR looks good to me. @pcuenca can I let you handle it from there (either merge/ship, or discuss further the download count rule since you are in the slack thread)

chrisdonahue-goog:
@pcuenca Yes, repo contains three components: an audio codec, and two different LLM sizes which use the same common audio codec. The LLM weights are in two different formats targeting different users / use cases: safetensors (for researchers) and pre-compiled MLX functions (for creative users).

Our downstream libraries/applications are set up to do targeted downloading of individual LLM checkpoint files from the repo based on use case.

We could track downloads to the audio codec, but this would be undercounting users who explore both LLM checkpoints, or even explore different use cases (research/creative). 

We prefer to keep all three components in one mono-repo to reduce maintenance burden on our end, rather than splitting into three different repos.

Please let us know if this is acceptable! Our launch is cemented for Thursday June 4 and we won't be able to sufficiently test the split repo route in time. Thanks in advance!

packages/tasks/src/model-libraries.ts
KEHANG: they are mutually exclusive. Some context if it helps.

1. mrt2_base and mrt2_small are two different model sizes
2. mlxfn is used by CPP application / UI builders (they are not advised to use safetensors)
3. safetensors are for Python researchers who are interested in model architectures (they are not going to use mlxfn much)

Also, the codebase has `mrt models download` CLI as the default way to download models, which is only downloading mlxfn.

Only when people are really interested in safetensors, they can run `mrt checkpoints download`

KEHANG:
Following the download count rule discussion here in case it's lost in the weeds, they are mutually exclusive. Additional context if it helps.

1. mrt2_base and mrt2_small are two different model sizes
2. mlxfn is used by CPP application / UI builders (they are not advised to use safetensors)
3. safetensors are for Python researchers who are interested in model architectures (they are not going to use mlxfn much)

Also, the codebase has `mrt models download` CLI as the default way to download models, which is only downloading mlxfn.

Only when people are really interested in safetensors, they run `mrt checkpoints download`

pcuenca:
Thanks for the clarifications @chrisdonahue-goog @KEHANG, merging now.

pcuenca:
cc @coyotte508 for info, it'd be amazing if there could be a deployment around tomorrow (only if convenient for the team, of course)

PR #2211, opened 2026-06-03
ci: route package installs through internal HF registry proxies

XciD:
## Summary

Wire the [`setup-hf-registry-proxies`](https://github.com/huggingface/hf-workflows/tree/main/.github/actions/setup-hf-registry-proxies) action into the CI and release workflows so package downloads go through the internal HuggingFace registry proxies, which hold back package versions younger than 3 days as a supply-chain defense.

The action is pinned to commit `21845492ef47d4e46b07a3b2a6e14fd4748c39cc`.

## Changes

**CI** (`test.yml`, `lint.yml`)
- `runs-on: ubuntu-latest` -> `runs-on: { group: aws-general-8-plus }` on every job (`node`, `browser`, `e2e`, `lint`, `format`).
- New `setup-hf-registry-proxies` step inserted right after each `actions/checkout`, before any package install / corepack setup.

**Release** (all 16 `*-publish.yml` workflows for `agents`, `blob`, `dduf`, `gearhash-jit`, `gguf`, `hub`, `inference`, `jinja`, `languages`, `mcp-client`, `ollama-utils`, `space-header`, `splitmix64-wasm`, `tasks`, `tiny-agents`, `xetchunk-wasm`)
- Same runner + proxy step.
- Drop the `ignore-scripts=true` line the proxy script writes into `~/.npmrc`. Almost every published package relies on `prepare` / `prepublishOnly` to build before publish (`pnpm run build`), and pnpm would otherwise skip them. The registry override (the actual supply-chain defense, with the 3-day hold) is kept.
- Add `--registry=https://registry.npmjs.org/` to the first `pnpm publish` call (the one that targets the public npm registry), since `~/.npmrc` now points at the proxy and the proxy does not accept uploads. The second `pnpm publish` (GitHub Packages) is unchanged because the preceding `actions/setup-node@v4` step already sets `registry-url: https://npm.pkg.github.com`.

## Notes for review

- Existing `actions/checkout@v3` / `actions/setup-node@v3` (or v4/v5) pins are kept as-is; this PR does not touch versions.
- The `browser` job in `test.yml` calls `google-chrome --version`. Please confirm the `aws-general-8-plus` runner group has Chrome preinstalled, otherwise an install step will be needed.
- The `e2e` job stands up a local Verdaccio registry on `localhost:4874`; the proxies should not interfere with the publish/install loop against the mock registry.
- Other workflows that touch packages (`update-specs.yml`, `ollama-template-update.yml`, `documentation.yml`, etc.) are intentionally left out of this PR; happy to do a follow-up if you want them protected too.

---

> [!NOTE]
> **Medium Risk**
> Changes how all package installs and public npm publishes run in CI/release; misconfiguration could break builds or releases, and re-enabling lifecycle scripts weakens one proxy script-hardening control while keeping registry filtering.
> 
> **Overview**
> Routes **CI installs and package releases** through Hugging Face’s pinned **`setup-hf-registry-proxies`** action so dependencies are fetched via internal npm proxies (with the usual version holdback), instead of hitting the public registry directly from GitHub-hosted runners.
> 
> **Lint**, **test `node`**, and all **16 `*-publish.yml`** release workflows now use the **`aws-general-8-plus`** runner group and run the proxy setup right after checkout. Because the action sets **`ignore-scripts=true`** in `~/.npmrc`, each affected job adds a step that **removes only that line** so `postinstall` / `prepare` / `prepublishOnly` still run while the proxy registry override stays in place.
> 
> **Release** workflows pass **`--registry=https://registry.npmjs.org/`** on the **first** `pnpm publish` (public npm), since the global registry now points at the proxy, which does not accept uploads; the GitHub Packages publish step is unchanged.
> 
> **`test.yml`** **`browser`** and **`e2e`** jobs **remain on `ubuntu-latest`**, with new comments that Chrome and Verdaccio/local-registry behavior do not fit the self-hosted runners (no proxy steps added there).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cd6a962110c1991a334da01209988d281bb9fdfd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


coyotte508:
we already have:

```yaml
minimumReleaseAge: 4320
minimumReleaseAgeExclude:
  - "@huggingface/*"
```

In `pnpm-workspace.yml` from https://github.com/huggingface-internal/moon-landing/pull/17553 (Apr. 1)

which is already 72h

PR #2215, opened 2026-06-04
Fix safetensors parameter count for mixed-precision compressed-tensors models

mishig25:
## Description

For `compressed-tensors` models with multiple `config_groups` at different bit widths, the pack-quantized multiplier used the **first** group's `num_bits` for **every** packed `I32` tensor — e.g. a model with 2-bit embeddings and 4-bit layers got all tensors unpacked at ×16 instead of ×8, inflating the parameter count (~5.5B shown as 9B in one real case).

### Fix

- Match each packed tensor's module name (`X.weight_packed` → `X`) against group `targets` and use that group's `num_bits`/`format`.
- New `matchesCompressedTensorsTarget` helper: exact names + `re:` regex targets, translated to the ReDoS-safe `globMatch` instead of evaluating `RegExp` from attacker-controlled `config.json`.
- No matching targets (e.g. class names like `["Linear"]`, the common single-group case) → previous behavior, so existing models are unaffected.

Verified byte-exactly against ground truth from the model's `weight_shape` tensors. Unit tests added for the matcher.

---

> [!NOTE]
> **Medium Risk**
> Changes how displayed parameter totals are computed for compressed-tensors models; logic is localized to hub metadata parsing with a safe fallback path.
> 
> **Overview**
> Fixes **inflated parameter counts** for `compressed-tensors` models that use **multiple `config_groups`** at different bit widths. Packed `I32` tensors were all unpacked with the **first** group's `num_bits`, so mixed setups (e.g. 2-bit embeddings vs 4-bit layers) could roughly double the reported total.
> 
> The hub now derives each tensor's module name from its safetensors key, finds the `config_group` whose **`targets`** match via new **`matchesCompressedTensorsTarget`** (exact names and a safe subset of `re:` patterns mapped to **`globMatch`**, not `RegExp` on `config.json`), and applies that group's **`format`** / **`num_bits`** for the pack-quantized multiplier. **`QuantizationConfig`** typing adds per-group **`targets`** and **`format`**. When no target matches, behavior stays the same as before (global `pack-quantized` + first `num_bits`). Unit tests cover the matcher.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b352b2a63cbf114bd53c3fce45d9797cf3708bd8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/hub/src/lib/parse-safetensors-metadata.ts
+++ b/packages/hub/src/lib/parse-safetensors-metadata.ts
@@ -377,7 +377,7 @@ export interface QuantizationConfig {
 	load_in_8bit?: boolean;
 	// compressed-tensors specific
 	format?: string;
-	config_groups?: Record<string, { weights?: { num_bits?: number } }>;
+	config_groups?: Record<string, { format?: string; targets?: string[]; weights?: { num_bits?: number } }>;
 }
 
 export interface ModelConfig {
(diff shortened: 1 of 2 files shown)


mishig25:
## Security review summary

| Category | Result |
|---|---|
| Code execution / RegExp injection | ✅ Safe — never constructs `RegExp`/`eval` from attacker input; deliberately follows the existing RegExp-free `globMatch` pattern; unsupported regex syntax is screened with a static literal regex and treated as non-matching |
| Prototype pollution | ✅ Safe — attacker data is only *read* (`Object.values`, `g.targets?.some`); no object writes with attacker-controlled keys are introduced |
| Path traversal / injection | ✅ Safe — `tensorName`/`moduleName` are sliced and compared as plain strings, never used in paths, URLs, or queries |
| Output sink impact | ✅ Benign — the function only selects an integer multiplier clamped to ≥ 1 (`Math.max(1, Math.floor(32 / numBits))`); worst case an attacker mis-reports their own model's parameter count (data-quality, not a security boundary) |

coyotte508 (approved):

packages/hub/src/lib/parse-safetensors-metadata.ts
```diff
@@ -377,7 +377,7 @@ export interface QuantizationConfig {
 	load_in_8bit?: boolean;
 	// compressed-tensors specific
 	format?: string;
-	config_groups?: Record<string, { weights?: { num_bits?: number } }>;
+	config_groups?: Record<string, { format?: string; targets?: string[]; weights?: { num_bits?: number } }>;
```
julien-c, line 380: add this to moon-landing validation too 🙏

julien-c (commented):
ok

packages/hub/src/lib/parse-safetensors-metadata.ts
julien-c: [(internal)](https://github.com/huggingface-internal/moon-landing/pull/18492)

PR #2230, opened 2026-06-15
[Inference Providers] fal-ai: fix ASR for browser webm & wav (data-URL MIME types)

hanouticelina:
## What this fixes

fal's data-URL decoder maps the declared MIME type to a file extension and **rejects any type that doesn't resolve to one** — so the label matters, not just the bytes. As a result, fal-ai automatic-speech-recognition failed for the most common inputs:

- Browser recordings: `MediaRecorder` produces `audio/webm;codecs=opus`, which the client either rejected (not in the allowlist) or fal rejected with `Unsupported data URL`.
- `audio/wav` uploads: rejected by fal with `Unsupported data URL`, even though `audio/wav` was in our allowlist.

The allowlist was inaccurate on both ends — it omitted webm and listed `audio/wav`/`audio/mp4`, which fal cannot decode from a data URL.

## Change

Remap the blob's MIME type to a label fal's data-URL decoder accepts before building the `audio_url`, instead of forwarding the raw type:

| input blob type | sent to fal |
|---|---|
| `audio/webm` (+ `;codecs=…`) | `video/webm` |
| `audio/wav` / `audio/wave` | `audio/x-wav` |
| `audio/mpeg` / `audio/mp3` | `audio/mpeg` |

Also: codec parameters (e.g. `;codecs=opus`) are stripped, the type is matched case-insensitively, and types fal can't decode from a data URL (`audio/mp4`, `audio/ogg`, …) now fail fast with a clear client-side error instead of fal's opaque `Unsupported data URL`.

## Verification

Tested end-to-end against `fal-ai/whisper`: `audio/mpeg`, `audio/x-wav` and `video/webm` decode and return correct transcripts; `audio/wav`, `audio/webm`, `audio/mp4`, `audio/x-m4a` and `audio/ogg` return `Unsupported data URL`. The client now sends only the accepted labels.

## Notes

- Safari's `MediaRecorder` emits `audio/mp4`, which fal can't decode from a data URL — this now surfaces a clear client-side error rather than a confusing one.
- The Python client (`huggingface_hub`) likely has the same fal-ai ASR limitation and should get the equivalent fix for parity.

---

> [!NOTE]
> **Low Risk**
> Scoped to fal-ai ASR payload encoding; no auth or shared API surface changes beyond clearer errors for unsupported MIME types.
> 
> **Overview**
> Fixes **fal-ai automatic speech recognition** when audio is sent as a base64 data URL: fal rejects many blob MIME labels even when the bytes are valid, so the client now **remaps** types before building `audio_url`.
> 
> Replaces the flat allowlist with **`FAL_AI_AUDIO_MIME_MAP`** (e.g. `audio/webm` → `video/webm`, `audio/wav` → `audio/x-wav`). **`preparePayloadAsync`** strips codec parameters (`;codecs=opus`), normalizes the base type case-insensitively, and uses the mapped label in the data URL. Unsupported types (e.g. Safari `audio/mp4`) fail early with a clear **`InferenceClientInputError`** instead of fal’s opaque “Unsupported data URL”.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 128f9ef57a5230c8b3f1aeb28713214b07c2402e. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


hanouticelina:
this should fix the error when using ASR models with fal
Screenshot 2026-06-15 at 14 03 01

SBrandeis (approved):

PR #2239, opened 2026-06-18
fix(inference): resolve provider mapping by task when a model maps to several tasks

adampachowicz:
## What

`getInferenceProviderMapping` selected the first mapping matching the provider and only *then* validated the task, throwing when it didn't match:

```ts
const providerMapping = mappings.find((mapping) => mapping.provider === params.provider);
// ...later...
if (!typedInclude(equivalentTasks, providerMapping.task)) {
  throw new InferenceClientInputError(`Model ${modelId} is not supported for task ${task} ...`);
}
```

When a model is mapped to the **same provider for more than one task** (e.g. a single checkpoint exposed for both `text-to-image` and `image-to-image`), a request for any task other than the first one returned by the API is incorrectly rejected — even though a valid mapping exists.

## Fix

Filter the provider's mappings and find the one matching the requested task (keeping the existing `EQUIVALENT_SENTENCE_TRANSFORMERS_TASKS` equivalence for `hf-inference`). The "unsupported task" error is preserved and now lists **every** task the provider supports for the model.

## Tests

Adds `test/getInferenceProviderMapping.spec.ts` (offline — seeds `inferenceProviderMappingCache`, no network):

- resolves each task to its own mapping when a model maps to one provider for two tasks
- resolves regardless of the order the API returns the mappings in
- still throws a helpful error (now listing all supported tasks) for an unsupported task
- single-mapping case unchanged

## Why

This is needed for any provider registering one model under multiple tasks. We hit it while preparing a new provider integration (deAPI) where `black-forest-labs/FLUX.2-klein-4B` is mapped to both `text-to-image` and `image-to-image`, but the change is provider-agnostic.

cc @Wauplin @hanouticelina — happy to adjust the error wording or test placement if you'd prefer.

---

> [!NOTE]
> **Medium Risk**
> Changes inference request routing for multi-task provider mappings; behavior is narrowed and tested but affects which providerId is used for live API calls.
> 
> **Overview**
> Fixes **incorrect task rejection** when one model has **multiple Hub mappings for the same provider** (e.g. `text-to-image` and `image-to-image` on the same checkpoint).
> 
> `getInferenceProviderMapping` no longer takes the **first** provider match and then validates the task. It **filters** mappings for the provider and **selects the entry whose task matches** the request (still honoring `EQUIVALENT_SENTENCE_TRANSFORMERS_TASKS` for `hf-inference`). Unsupported tasks still throw `InferenceClientInputError`, but the message now lists **all** tasks that provider supports for that model.
> 
> Adds offline Vitest coverage in `getInferenceProviderMapping.spec.ts` (cache seeding, multi-task resolution, mapping order, error text, single-mapping regression).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 06dc9e81892bbd499fc16e250b831cc87412f074. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/inference/src/lib/getInferenceProviderMapping.ts
+++ b/packages/inference/src/lib/getInferenceProviderMapping.ts
@@ -139,15 +139,23 @@ export async function getInferenceProviderMapping(
 		return HARDCODED_MODEL_INFERENCE_MAPPING[params.provider][params.modelId];
 	}
 	const mappings = await fetchInferenceProviderMappingForModel(params.modelId, params.accessToken, options);
-	const providerMapping = mappings.find((mapping) => mapping.provider === params.provider);
-	if (providerMapping) {
+	const providerMappings = mappings.filter((mapping) => mapping.provider === params.provider);
+	if (providerMappings.length > 0) {
 		const equivalentTasks =
 			params.provider === "hf-inference" && typedInclude(EQUIVALENT_SENTENCE_TRANSFORMERS_TASKS, params.task)
 				? EQUIVALENT_SENTENCE_TRANSFORMERS_TASKS
 				: [params.task];
-		if (!typedInclude(equivalentTasks, providerMapping.task)) {
+		// A model can be mapped to the same provider for several tasks (e.g. a single checkpoint
+		// exposed for both text-to-image and image-to-image). Match on the requested task rather
+		// than blindly taking the first mapping for the provider, which would reject valid
+		// requests when the task we want is not the first entry returned by the API.
+		const providerMapping = providerMappings.find((mapping) => typedInclude(equivalentTasks, mapping.task));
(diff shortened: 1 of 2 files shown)


SBrandeis:
Hi, thank you for your contribution!
A model can only be mapped to a single task per provider server-side, so this change is not needed
I will close the Pull Request

adampachowicz:
Thanks for the quick look — makes sense, closing is fine. Quick follow-up for a provider integration we're preparing: when a single underlying checkpoint should be exposed for two tasks (e.g. `text-to-image` and `image-to-image`), is the intended approach to register two separate HF model repos, one per task? Just want to map it correctly on our side. Thanks!

PR #2250, opened 2026-06-23
fix: use causal LM auto class for text-generation snippets

OMCHOKSI108:
## Summary

This PR fixes Transformers snippet generation for text-generation causal LM models whose `transformersInfo.auto_model` is incorrectly set to `AutoModelForMultimodalLM`.

## Problem

Some text-generation models, such as models with architectures like `Qwen2ForCausalLM`, can generate a snippet using:

```
AutoModelForMultimodalLM
```

even though the model is a text-generation causal language model. This can produce confusing or incorrect generated usage snippets.

## Fix

Inside `transformers()`, this PR now uses a local shallow copy of `model.transformersInfo` and overrides `auto_model` to `AutoModelForCausalLM` only when all of the following are true:

- `model.pipeline_tag === "text-generation"`
- `model.config?.architectures` contains an architecture ending with `ForCausalLM`
- `transformersInfo.auto_model === "AutoModelForMultimodalLM"`

The original `model.transformersInfo` object is not mutated.

## Tests

Added a test case matching the VibeThinker-3B shape:

- `pipeline_tag: "text-generation"`
- `architectures: ["Qwen2ForCausalLM"]`
- `auto_model: "AutoModelForMultimodalLM"`

The test verifies that the generated snippet contains `AutoModelForCausalLM` and does not contain `AutoModelForMultimodalLM`.

## Validation

- TypeScript type-check passed
- Vitest suite passed
- Existing multimodal snippet logic was not changed

---

> [!NOTE]
> **Low Risk**
> Display-only snippet string changes with narrow, rule-gated overrides; no inference, auth, or persistence paths touched.
> 
> **Overview**
> Fixes **Transformers** and **Pruna** usage snippets for text-generation causal LMs whose metadata incorrectly lists `AutoModelForMultimodalLM` as `auto_model` (e.g. `Qwen2ForCausalLM` models like VibeThinker-3B).
> 
> Adds **`getTransformersInfoForSnippet`**, which returns a shallow copy of `transformersInfo` and overrides `auto_model` to **`AutoModelForCausalLM`** only when `pipeline_tag` is `text-generation`, an architecture ends with `ForCausalLM`, and the stored class is `AutoModelForMultimodalLM`—without mutating the original model object. **`transformers()`** and **`pruna_transformers()`** now build snippets from that corrected info.
> 
> Vitest cases assert generated snippets use the causal LM class (and that Pruna output still uses `PrunaModel.from_pretrained` without leaking the wrong auto class).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 64d1e49e5c951c6874036d38e4e422ef46b085e9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/tasks/src/model-libraries-snippets.spec.ts
+++ b/packages/tasks/src/model-libraries-snippets.spec.ts
@@ -1,6 +1,6 @@
 import { describe, expect, it } from "vitest";
 import type { ModelData } from "./model-data.js";
-import { llama_cpp_python, transformers } from "./model-libraries-snippets.js";
+import { llama_cpp_python, pruna, transformers } from "./model-libraries-snippets.js";
 
 describe("model-libraries-snippets", () => {
 	it("llama_cpp_python conversational", async () => {
@@ -75,5 +75,27 @@ print(output)`);
 
 		expect(snippet).toContain("AutoModelForCausalLM");
 		expect(snippet).not.toContain("AutoModelForMultimodalLM");
+	});
+
+	it("pruna transformers cleanup stays consistent with the corrected auto_model", async () => {
+		const model: ModelData = {
+			id: "WeiboAI/VibeThinker-3B",
+			pipeline_tag: "text-generation",
+			tags: ["transformers"],
+			inference: "",
+			config: {
+				architectures: ["Qwen2ForCausalLM"],
+			},
+			transformersInfo: {
+				auto_model: "AutoModelForMultimodalLM",
+				pipeline_tag: "text-generation",
+				processor: "AutoTokenizer",
+			},
+		};
+		const snippet = pruna(model).join("\n");
+
+		expect(snippet).not.toContain("AutoModelForMultimodalLM");
+		expect(snippet).not.toContain("AutoModelForCausalLM.from_pretrained");
+		expect(snippet).toContain("PrunaModel.from_pretrained");
 	});
 });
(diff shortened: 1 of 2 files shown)


OMCHOKSI108:
Thanks, fixed in the latest commit.

I moved the auto-model correction into a shared `getTransformersInfoForSnippet(model)` helper so both `transformers()` and `pruna_transformers()` now use the same corrected shallow copy of `transformersInfo`.

Also added Pruna test coverage for the VibeThinker-style text-generation causal LM case to ensure:

* `AutoModelForMultimodalLM` is not kept in the generated Pruna snippet
* stale `AutoModelForCausalLM.from_pretrained` calls are not left behind
* `PrunaModel.from_pretrained` is used consistently

Validation passed locally:

* `pnpm --filter @huggingface/tasks check`
* `pnpm --filter @huggingface/tasks test` — 20/20 passing
* `pnpm --filter @huggingface/tasks lint:check`

zucchini-nlp:
Hey @OMCHOKSI108 ,

Thanks for reporting this! We don't really want to fix it per model-id and after some investigation found that it was originating from presaved metadata from `transformers`.  It will be fixed by https://github.com/huggingface/transformers/pull/46844 :)

I'll close this PR, and the fixes will prob be upstream in a day or so. The metadata on the hub has to be synced after merging above PR

OMCHOKSI108:
Thanks for checking and for tracing the root cause upstream!

That makes sense — if the incorrect `auto_model` is coming from presaved Transformers metadata, fixing it there is cleaner than adding a defensive override in `huggingface.js`.

I’ll follow `huggingface/transformers#46844` and verify the generated Hub snippet again after the metadata sync.

Thanks again for the review.

PR #2267, opened 2026-07-01
Add Radeon 680M and update Radeon 740M power

wildernessfamily:
> [!NOTE]
> **Low Risk**
> Data-only catalog change with no logic or security impact; wrong spec values could mis-rank cost/power for that SKU only.
> 
> **Overview**
> Adds a **Radeon 680M** entry to `AMD_GPU_SKUS` in `hardware-amd.ts`, alongside other integrated APUs (610M, 740M, etc.).
> 
> The new spec uses shared-memory options via `AMD_GPU_INTEGRATED_SHARED_MEMORY_OPTIONS`, **gfx1035**, ~6.76 TFLOPS, 50W power, $120 MSRP, and 2022 release year—so scheduling/costing logic can recognize this SKU when hardware is matched by name.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit edc15f467dbfa1486a5d80303b69481e2b9f75e6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


BlivionIaG:
Hi, for the tdp, the tricky part is that it's also the same igpu as the 8500G sku so it can be 45w but also for the 7540U laptop variant its included in the  global 15w to 30w tdp enveloppe (someone can correct me if i am wrong).

coyotte508 (approved):

packages/tasks/src/hardware-amd.ts
```diff
@@ -260,12 +260,20 @@ export const AMD_GPU_SKUS: Record<string, AmdGpuHardwareSpec> = {
 		power: 15,
 		releaseYear: 2022,
 	},
+	"Radeon 680M": {
+		tflops: 3.379,
```
BlivionIaG:
```suggestion
		tflops: 6.76,
```
2 digits rounded to nearest (6.758)

packages/tasks/src/hardware-amd.ts
```diff
@@ -260,12 +260,20 @@ export const AMD_GPU_SKUS: Record<string, AmdGpuHardwareSpec> = {
...
+		power: 50,
+		releaseYear: 2022,
+	},
 	"Radeon 740M": {
 		tflops: 5.12,
 		memory: AMD_GPU_INTEGRATED_SHARED_MEMORY_OPTIONS,
 		gfxVersion: "gfx1103",
 		msrp: 179,
-		power: 15,
+		power: 30,
```
pcuenca: Perhaps it's better to leave it as documented [here](https://www.techpowerup.com/gpu-specs/radeon-740m.c4162), which says 15W for 740M and 50 for 680M.

pcuenca:
```suggestion
		power: 15,
```

pcuenca (approved):

pcuenca:
This is good for me to merge @wildernessfamily. Do you agree with reverting the tdp change of the 740M, or do you have additional data in this regard?

packages/tasks/src/hardware-amd.ts
BlivionIaG: Yeah maybe lowest known is something that could be applied in order to choose. I had this exact same igpu in the past under a desktop apu and that is documented with a higher tdp [here](https://www.techpowerup.com/gpu-specs/radeon-740m.c4162)

pcuenca: I still see 15W in that page lol. But I understand draw can be higher, especially when used on desktop, as described [here](https://www.notebookcheck.net/AMD-Radeon-740M-GPU-Benchmarks-and-Specs.716455.0.html).

Our spec actually [recommends to use the max](https://github.com/huggingface/huggingface.js/blob/main/packages/tasks/src/hardware.ts#L48-L49) when there are various form factors. What would you think would be least surprising / more accurate for an user searching for this card? Feel free to revert again to 30 if you think that's better!

BlivionIaG: Ah yes my bad I ended up pasting the wrong one https://www.techpowerup.com/gpu-specs/radeon-740m.c4297

pcuenca: Ok, going back to 30 after the discussion.

pcuenca: Or maybe not, because then we'd need to update the 760M and others as well, right? Let's defer that for a new PR, feel free to open it as a followup @wildernessfamily!

PR #2281, opened 2026-07-08
Register Mobilint library with MXQ download counting

parkjinman98:
## Summary

This PR registers `mobilint` as a Hub model library and configures download counting for Mobilint `.mxq` deployment artifacts while preserving standard `config.json` counting.

The integration is intentionally opt-in. It applies to repositories that declare `library_name: mobilint` in their model card metadata. Repositories that continue to declare another library, such as `transformers`, are intentionally out of scope and should continue using their existing behavior.

## Changes

- Add `mobilint` to `packages/tasks/src/model-libraries.ts`
- Set Mobilint download counting to `path:"config.json" OR path_extension:"mxq"`
- Add a minimal Python snippet using `mblt-model-zoo`
- Link the library integration to `mobilint/mblt-model-zoo`
- Set `filter: false` for the newly registered library

## Download counting rationale

The Hub's default query files such as `config.json` remain valid for many Mobilint repositories and should continue to count downloads. For models that opt into `library_name: mobilint` and publish Mobilint deployment artifacts, users may also download `.mxq` files to run models on Mobilint NPUs through Mobilint SDK qb / qbruntime.

Therefore, the Mobilint library integration counts both the standard `config.json` signal and `.mxq` deployment artifacts.

The rule intentionally does not count `.onnx`, arbitrary `.json`, `.safetensors`, `.bin`, or other source/config files, and it does not affect repositories that remain associated with another library.

## Runtime snippet rationale

The Python snippet uses Mobilint Model Zoo's high-level `MBLT_Engine` API:
- load a model by `model_cls`
- let `mblt-model-zoo` download and cache the matching artifact when `model_path=""`
- run the standard preprocess, inference, and postprocess flow
- dispose the model after use

This keeps the Hub snippet minimal and leaves detailed model selection and runtime options to the Mobilint Model Zoo documentation.

## Scope

This PR does not propose `.mxq` as a generic Hub-wide file format. It scopes `.mxq` download counting to the Mobilint library/runtime ecosystem only.

## Validation

- [x] `pnpm --filter @huggingface/tasks format:check`
- [x] `pnpm --filter @huggingface/tasks lint:check`
- [x] `pnpm --filter @huggingface/tasks check`
- [x] `pnpm --filter @huggingface/tasks test`

### DIFF
--- a/packages/tasks/src/model-libraries-snippets.ts
+++ b/packages/tasks/src/model-libraries-snippets.ts
@@ -2312,6 +2312,29 @@ export const model2vec = (model: ModelData): string[] => [
 model = StaticModel.from_pretrained("${model.id}")`,
 ];
 
+export const mobilint = (model: ModelData): string[] => {
+	const modelName = nameWithoutNamespace(model.id);
+	return [
+		`# pip install mblt-model-zoo
+from mblt_model_zoo.vision import MBLT_Engine
+
+model = MBLT_Engine(
+    model_cls="${modelName}",
+    model_type="DEFAULT",
+    model_path="",
+    core_mode="global8",
+)
+
+try:
+    image = model.preprocess("path/to/image.jpg")
+    output = model(image)
+    result = model.postprocess(output)
+finally:
+    model.dispose()
+`,
+	];
+};
+
 export const pruna = (model: ModelData): string[] => {
 	let snippets: string[];
 
--- a/packages/tasks/src/model-libraries.ts
+++ b/packages/tasks/src/model-libraries.ts
@@ -967,6 +967,15 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		snippets: snippets.model2vec,
 		filter: false,
 	},
+	mobilint: {
+		prettyLabel: "Mobilint",
+		repoName: "mblt-model-zoo",
+		repoUrl: "https://github.com/mobilint/mblt-model-zoo",
+		docsUrl: "https://docs.mobilint.com",
+		countDownloads: `path:"config.json" OR path_extension:"mxq"`,
+		snippets: snippets.mobilint,
+		filter: false,
+	},
 	moshi: {
 		prettyLabel: "Moshi",
 		repoName: "Moshi",


packages/tasks/src/model-libraries.ts
```diff
@@ -967,6 +967,15 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		snippets: snippets.model2vec,
 		filter: false,
 	},
+	mobilint: {
+		prettyLabel: "Mobilint",
+		repoName: "mblt-model-zoo",
+		repoUrl: "https://github.com/mobilint/mblt-model-zoo",
+		docsUrl: "https://docs.mobilint.com",
+		countDownloads: `path_extension:"mxq"`,
```
Wauplin: Hi @parkjinman98, thanks for the PR. From what I can see in https://huggingface.co/models?other=mobilint, not all repos follow this convention. Typically a lot of them have a `config.json` which is already tracked correctly (and which won't be tracked anymore if the rule is overwritten here).

I'd suggest:
- either update the download counter rule in this PR
- or, define a library tag that filters only the models with .mxq files

My favorite is the second option since it also fixes the problem that the snippet is for mxq files only

packages/tasks/src/model-libraries-snippets.ts
```diff
@@ -2312,6 +2312,50 @@ export const model2vec = (model: ModelData): string[] => [
...
+elif core_mode == "global8":
+    mc.set_global8_core_mode()
+    model = Model(mxq_path, mc)
+else:
+    raise ValueError(f"Unsupported core mode: {core_mode}")
+
+model.launch(acc)
+
+# Prepare model-specific input before running inference.
+# result = model.infer(preprocessed_input)
```
Wauplin: we usually recommend a single self-contained minimal example and then let the user explore the docs by themselves

the ideal case would be for "mobilint" to have a `from_pretrained`-like method to take care of the file download + file loading at the same time. Do you think that's possible on your side?

parkjinman98: Sure, that should be possible on our side.

We already provide a similar higher-level loading flow through `mblt-model-zoo`. For vision models, users can use `MBLT_Engine`, and if `model_path` is not provided or does not point to an existing local MXQ file, the corresponding MXQ artifact is downloaded from the Hugging Face Hub and cached automatically.

https://github.com/mobilint/mblt-model-zoo/tree/master

So, I agree that the Hub snippet can stay as a single minimal example, while the more complete download/load/preprocess/postprocess workflow can be handled by our model-zoo API and documentation.

packages/tasks/src/model-libraries.ts
parkjinman98: Hi @Wauplin,

Thank you for the suggestion. The second option is exactly aligned with the plan I had discussed with my team before opening this PR.

We have two categories of Mobilint repositories:

1. Repositories that are seamless overrides for Hugging Face Transformers models.
   These repositories contain `config.json` and will keep `library_name: transformers`. They may still have a broad `mobilint` tag, but they are intentionally not opting into the Mobilint library integration, so their existing Transformers / `config.json`-based behavior is preserved.

2. MXQ deployment repositories, such as AlexNet, that do not contain `config.json`.
   These repositories will opt into `library_name: mobilint`. In our supported download flow, https://github.com/mobilint/mblt-model-zoo downloads the selected `.mxq` artifact one at a time, so counting `.mxq` files reflects downloads of the actual deployable artifact and should not introduce duplicate counting.

To be conservative, I will also update the download counting rule to include both cases explicitly:

```ts
countDownloads: `path:"config.json" OR path_extension:"mxq"`,
```

This preserves `config.json` counting for any repository that explicitly opts into `library_name: mobilint`, while adding `.mxq` counting for MXQ-only Mobilint deployment repositories.

It also does not count `.onnx` files or arbitrary JSON result files, since the rule only matches the exact `config.json` path or files with the `.mxq` extension.

Wauplin (approved):
Thanks for iterating! Now looks good 👍 

I'll merge this PR now. Expect 2-3 days before getting it on the Hub.

PR #2294, opened 2026-07-14
don't restrict safetensors to model repos

YassineYousfi:
datasets like:
https://huggingface.co/datasets/commaai/comma1M
https://huggingface.co/datasets/stady4d-anon/StaDy4D
https://huggingface.co/datasets/LAXMAYDAY/pdm3-ht-20260528-flux2-vae-latents-public
https://huggingface.co/datasets/Carnot-EBM/token-activations

are using safetensors to save tensor data, not model weights. which currently throws the error "TypeError: Only model repos should contain safetensors files."

this removes that check and changes the docstrings a bit to reflect that.

> [!NOTE]
> **Low Risk**
> Small API relaxation in hub metadata parsing; quantization-aware counting behavior for model repos is unchanged.
> 
> **Overview**
> **`parseSafetensorsMetadata`** no longer requires a **model** repo; any Hugging Face repo type can be parsed for safetensors headers and optional parameter counts.
> 
> The previous **`TypeError`** for non-model repos is removed. **`config.json`** (quantization hints for parameter counting) is still loaded only when **`computeParametersCount`** is true **and** the repo is a model; other repo types skip that fetch. Docs and the “no weights found” error now say **repo** instead of **model**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 620eb6d2a239cf3b14326e04c000e72d95e59b43. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/hub/src/lib/parse-safetensors-metadata.ts
+++ b/packages/hub/src/lib/parse-safetensors-metadata.ts
@@ -282,12 +282,12 @@ const LIBRARY_WEIGHT_CANDIDATES: Record<string, Array<{ single: string; index: s
 };
 
 /**
- * Analyze model.safetensors.index.json or model.safetensors from a model hosted
- * on Hugging Face using smart range requests to extract its metadata.
+ * Analyze a safetensors file or index hosted in a Hugging Face repository
+ * using range requests to extract its metadata.
  */
 export async function parseSafetensorsMetadata(
 	params: {
-		/** Only models are supported */
+		/** Repository containing the safetensors file. */
 		repo: RepoDesignation;
 		/**
 		 * Relative file path to safetensors file inside `repo`. Defaults to `SAFETENSORS_FILE` or `SAFETENSORS_INDEX_FILE` (whichever one exists).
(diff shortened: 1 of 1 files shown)


julien-c (commented):
cc @mishig25 but i don't think we want to do that.

This'd be better served by a util in a different repo, @YassineYousfi

coyotte508 (approved):

julien-c (changes requested):
blocking merge in the meantime

YassineYousfi:
Thanks for reviewing @julien-c 
I'm not sure I understand what you mean by "served by a util in a different repo"
Do you mean to not upload safetensors in dataset repos? 
Or do you mean that this specific parsing function should be moved somewhere else?
Currently that shows this in the hub.
image

julien-c:
ok yes it is indeed a bug that it is trying to load the safetensors file there (it shouldn't do anything)

PR #2323, opened 2026-07-28
Stop counting mxfp4, bitsandbytes and actorder bookkeeping tensors as parameters

mishig25:
Follow-up to #2321. That PR established that block scales are quantization bookkeeping rather than model parameters, but applied the rule to only two of the four code paths that need it. The three remaining paths each overcount — two by more than the Kimi-K2.5 correction #2321 did make.

## Root causes

**1. `quant_method: "mxfp4"` (gpt-oss) counts its block scales.** transformers pairs [`{proj}_blocks` with `{proj}_scales`](https://github.com/huggingface/transformers/blob/dff4572dfa4bfa9f00cc8414e4b84877552fefe9/src/transformers/integrations/mxfp4.py#L111-L123) — joined with an **underscore**, not a dot — so `getTensorSuffix` returns the whole segment and `QUANTIZATION_AUXILIARY_SUFFIXES` can never match; `shouldSkipTensor` then bails because the method is neither gptq nor awq. That they're exponents and not weights is explicit in the dequant: [`scales.to(torch.int32) - 127`](https://github.com/huggingface/transformers/blob/dff4572dfa4bfa9f00cc8414e4b84877552fefe9/src/transformers/integrations/mxfp4.py#L271), one per 32 weights per the [`hidden_size // 32, 16` uint8 block shape](https://github.com/huggingface/transformers/blob/dff4572dfa4bfa9f00cc8414e4b84877552fefe9/src/transformers/integrations/mxfp4.py#L349-L360). Same UE8M0 scales #2321 already excludes for Kimi-K3 via `weight_scale`.

Gated on the `U8` container the scales actually use, mirroring the existing `_blocks` gate, so a learnable `*_scales` parameter elsewhere is left alone.

**2. bitsandbytes counts its double-quantization state, and doubles it.** The exact key set is upstream's [`valid_qs_keys` / `valid_qs_type_keys`](https://github.com/bitsandbytes-foundation/bitsandbytes/blob/b88ddac2d3e7484f6adfe4668053f4647be89809/bitsandbytes/functional.py#L423-L429): `absmax`, `quant_map`, `nested_absmax`, `nested_quant_map`, plus the `bitsandbytes__{nf4,fp4}` blob. Because `absmax` shares the weights' `U8` container it also picked up the 4-bit packing factor — counted **and** multiplied, exactly the double error `weight_shape` had.

**3. compressed-tensors `weight_g_idx` is not excluded.** [`pack_quantized.compression_param_names`](https://github.com/vllm-project/compressed-tensors/blob/46196ce607421d9992a47644612dc6b443f3c434/src/compressed_tensors/compressors/pack_quantized/base.py#L46-L54) emits `weight_packed`, `weight_scale`, `weight_shape`, `weight_zero_point`, and `weight_g_idx` when `actorder: "group"`. #2321 handles the first four. `weight_g_idx` is `I32`, so like `weight_shape` it was counted and scaled by the packing factor — one index per input column, ×8.

Also added `input_global_scale`, emitted by the [nvfp4 compressor](https://github.com/vllm-project/compressed-tensors/blob/46196ce607421d9992a47644612dc6b443f3c434/src/compressed_tensors/compressors/nvfp4/base.py#L37-L49) alongside the already-excluded `weight_global_scale`. A scalar, so numerically irrelevant; included for completeness.

## Verification

Every corrected total independently matches a published figure — exact counts, not thresholds:

| model | scheme | before | after | reference |
|---|---|---|---|---|
| `openai/gpt-oss-120b` | mxfp4 | 120,412,337,472 | **116,829,156,672** | OpenAI: 116.8B / "117B" |
| `openai/gpt-oss-20b` | mxfp4 | 21,511,953,984 | **20,914,757,184** | OpenAI: 20.9B / "21B" |
| `unsloth/Meta-Llama-3.1-8B-Instruct-bnb-4bit` | bnb 4-bit | 8,248,929,342 | **8,030,261,248** | Llama-3.1-8B |
| `RedHatAI/Llama-3.3-70B-Instruct-quantized.w4a16` | ct `pack-quantized`, actorder | 70,603,514,976 | **70,553,706,496** | `meta-llama/Llama-3.3-70B-Instruct`, exactly |

The 70B row is the tightest check available: a w4a16 quantization should report the count of the model it came from, and it now matches to the digit. The `weight_g_idx` delta decomposes exactly too — 80 layers × (6 × 8192 + 28672) = 6,225,920 indices × 8 = 49,807,360. The `before` figures for gpt-oss and bnb are what the Hub API serves today.

## Not included

Further findings from reviewing #2321, left out to keep this reviewable:

- [ ] Scope the `fp8` + `expert_dtype` packing to experts — it currently applies to every integer container, safe today only because DeepSeek-V4-Pro's one non-expert integer tensor is `I64`
- [ ] `packingFactor` floors `containerBits / numBits`, but the compressor packs densely (`ceil(cols * num_bits / 32)`) — 3-bit gives 10 instead of 10.67
- [ ] Drop the `["fp6", 6]` entry, or fix it — `floor(8 / 6) === 1`, so it can never fire in an 8-bit container
- [ ] Fix #2321's blast-radius table: `Meta-Llama-3.1-8B-Instruct-quantized.w4a16` is `quant_method: "gptq"`, not compressed-tensors
- [ ] Rewrite the two "sub-byte data types" tests (`spec.ts:246`, `:344`) — they shadow the real counter with a local reimplementation, asserting `E8M0 === 24` and `UE8 === 5000` while production skips both dtypes

### DIFF
--- a/packages/hub/src/lib/parse-safetensors-metadata.spec.ts
+++ b/packages/hub/src/lib/parse-safetensors-metadata.spec.ts
@@ -306,11 +306,74 @@ describe("parseSafetensorsMetadata", () => {
 
 		const totalParams = parse.parameterTotal || sum(Object.values(parse.parameterCount));
 
-		assert.strictEqual(totalParams, 21_511_953_984); // 21.5B
-
-		assert.ok(parse.parameterCount.BF16 && parse.parameterCount.U8);
+		// 20.9B, matching OpenAI's published figure. Previously 21_511_953_984, which counted the
+		// 597_196_800 `..._scales` UE8M0 block exponents as parameters (+2.86%).
+		assert.strictEqual(totalParams, 20_914_757_184);
+		assert.deepStrictEqual(parse.parameterCount, { BF16: 1_804_459_584, U8: 19_110_297_600 });
 
 		assert.strictEqual(Object.keys(parse.headers).length, 3);
+	});
+
+	it("excludes mxfp4 block scales stored as `..._scales` (openai/gpt-oss-120b)", async () => {
+		// gpt-oss joins the scales suffix with an underscore rather than a dot, so it slips past the
+		// dotted auxiliary-suffix list. Counting the 3_583_180_800 U8 scales reported 120_412_337_472
+		// (120.4B) for a model OpenAI documents as 116.8B / "117B".
+		const parse = await parseSafetensorsMetadata({
+			repo: "openai/gpt-oss-120b",
+			revision: "b5c939de8f754692c1647ca79fbf85e8c1e70f8a",
+			computeParametersCount: true,
+		});
+
+		assert(parse.sharded);
+		assert.strictEqual(Object.keys(parse.headers).length, 15);
+		assert.deepStrictEqual(parse.parameterCount, {
+			BF16: 2_167_371_072,
+			U8: 114_661_785_600, // 57_330_892_800 packed bytes x 2, scales excluded
+		});
+		assert.strictEqual(sum(Object.values(parse.parameterCount)), 116_829_156_672);
+	});
+
+	it("excludes bitsandbytes double-quantization state (unsloth/Meta-Llama-3.1-8B-Instruct-bnb-4bit)", async () => {
+		// `absmax` shares the weights' U8 container, so it was counted *and* doubled by the 4-bit
+		// packing factor — the same double error `weight_shape` had. Reported 8_248_929_342 for a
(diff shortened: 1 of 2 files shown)


coyotte508 (approved):
You should probably take over https://github.com/huggingface-internal/moon-landing/pull/19229 then

(bump version + add new props, ...)

mishig25:
yep

mishig25:
I'm cross checking against transfortmer codebase for all those changes

Issue #1279, opened 2025-03-12
Fine-tune download speed for xet + stream backpressure


coyotte508:
Currently we have a mechanism to have 1000 chunks in backpressure. (would love that to be a byte amount but it's not doable easily due to incompatibilities between implementations of `ReadableStream`, would need per-platform adaptations)

But if the stream is consumed instantly (eg writing to file), we have a bottleneck because we don't make calls in parallel to the CAS service.

Eg someone with a download speed of 1GB/s, each chunk CAS range is at most 64MB compressed, that's 16 HTTP request/second, assuming a 30ms ping, that's a 33% download speed decrease just from doing the request one after the other.

And it can even be worse, with smaller ranges, and higher bandwiths.

So we need to add some parallelism probably, and ideally be able to react to the download speed / latency

Like maybe open multiple read streams in // and join them and add on a new read stream as an old one is consumed (see also https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream/pipeTo)

coyotte508:
https://github.com/huggingface/huggingface.js/pull/2350

PR #2337, opened 2026-08-01
🐢 [hub] Parallelize xet downloads with a downloadConcurrency option

take-cheeze:
I'd like to have this feature in the library since it was quite useful on onnx model processing in wasm.
https://onnxsim.github.io/onnxsim/ is where I deploy the onnx simplifier and https://github.com/onnxsim/onnxsim/pull/559 was the PR I've worked.
Currently I'm using `8` as preferred concurrency number (on my 100Mbps network, about 2MB/s was the serial speed and theoretical limit should be around 100MB/s so I've selected 8 to 10) but if there is other better value, I'd like to switch to it.

Rel: https://github.com/huggingface/huggingface.js/issues/1279

Add an optional `downloadConcurrency` to `downloadFile` (threaded into `XetBlob`) so a xet-backed file's reconstruction ranges can be fetched in parallel. Defaults to 1, preserving the current fully-serial behavior.

XetBlob's reconstruction loop awaits each term's CAS range request in order, so only one request is ever in flight, capping xet downloads at single-connection throughput. This adds a bounded look-ahead prefetch window: up to `downloadConcurrency` range requests are kept in flight, deduplicated by CAS range so it never issues more unique requests than the serial path and still honors the RangeList cache fast-path.

Decompression and in-order yielding are unchanged; only the launch of each per-term fetch moves earlier. Prefetched-but-unused responses (slice / maxBytes early-exit) are cancelled, and the 403 token-refresh path drops stale prefetched responses before retrying.

Claude-Session: https://claude.ai/code/session_01PfdUuzCxrTY8kNS7PKbxD2

---

> [!NOTE]
> **Medium Risk**
> Touches core xet reconstruction and streaming with new concurrency and cancellation paths; default remains serial, but higher concurrency increases load on CAS/CDN and relies on correct prefetch invalidation on 403/errors.
> 
> **Overview**
> Adds optional **`downloadConcurrency`** on `downloadFile` (forwarded to `XetBlob`) so xet reconstruction can issue multiple CAS byte-range fetches in parallel instead of one-at-a-time. Default **`1`** keeps today’s serial behavior.
> 
> **`XetBlob`** gains a bounded look-ahead prefetch: up to `downloadConcurrency` range requests can be in flight, deduped by CAS range key so prefetch never exceeds unique work the serial path would do. Decompression and in-order output are unchanged; only when each term’s HTTP fetch starts moves earlier.
> 
> Failure and lifecycle handling: on **403** token refresh, stale prefetched responses are dropped before retry; a **`finally`** block and stream **`cancel`** cancel the active reader and unused prefetched bodies (including when a term fails mid-download). Tests cover parallel vs serial in-flight counts and prefetch cancellation on error.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7726229ebf2fcf45a5c52c732915e302d4822de5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/hub/src/lib/download-file.ts
+++ b/packages/hub/src/lib/download-file.ts
@@ -43,6 +43,14 @@ export async function downloadFile(
 		 * Can save an http request if provided
 		 */
 		downloadInfo?: FileDownloadInfoOutput;
+		/**
+		 * Number of CAS range requests to keep in flight when downloading via the
+		 * xet protocol. Higher values parallelize the transfer; ~8–10 is usually
+		 * enough to saturate a fast link. Defaults to `1` (serial). Ignored for
+		 * non-xet downloads.
+		 * @default 1
+		 */
+		downloadConcurrency?: number;
 	} & Partial<CredentialsParams>,
 ): Promise<Blob | null> {
 	const accessToken = checkCredentials(params);
@@ -71,6 +79,7 @@ export async function downloadFile(
 			accessToken,
 			size: info.size,
 			readToken: typeof params.xet === "object" ? params.xet.readToken : undefined,
+			downloadConcurrency: params.downloadConcurrency,
 		});
 	}
 
--- a/packages/hub/src/utils/XetBlob.spec.ts
+++ b/packages/hub/src/utils/XetBlob.spec.ts
@@ -811,6 +811,130 @@ describe("XetBlob", () => {
 			});
 		});
 
+		describe("downloadConcurrency", () => {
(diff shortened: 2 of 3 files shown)


coyotte508:
will look at it together with https://github.com/huggingface/huggingface.js/pull/2265

take-cheeze:
https://github.com/huggingface/huggingface.js/pull/2350 seems to be better! Thank you!

PR #2339, opened 2026-08-03
Add QVAC to local apps

thomasblc:
## Summary
- Add [QVAC](https://docs.qvac.tether.io) as a local app for GGUF text-generation models
- QVAC is Tether's open-source local AI SDK (Apache 2.0, JS/TS). Its CLI ships an OpenAI-compatible server, so a GGUF hosted on the Hub runs locally with no cloud dependency
- Unlike Pi / Hermes Agent / OpenClaw, QVAC **is** the inference engine rather than a client of one, so the snippet has no separate llama.cpp server step: the CLI serves the weights itself
- `serve.models` resolves SDK model constants under the `model` shorthand, so Hub weights use the explicit `{ type, src }` form, where `src` is handed to the SDK as `modelSrc` verbatim

The snippet is two steps: install `@qvac/cli` plus a `qvac.config.json` registering the model, then `qvac serve openai`. It supports the `{{GGUF_FILE}}` placeholder and uses the resolved filename when one is selected.

## Verified before submitting
Ran the generated snippet for real against a Hub GGUF (`unsloth/Qwen3-0.6B-GGUF`, served straight from a `huggingface.co/.../resolve/main/...` URL):
- `GET /v1/models` lists the alias
- `POST /v1/chat/completions` returned a normal completion (`finish_reason: stop`, 73 completion tokens)

## Notes on two deliberate choices
- **`displayOnModelPage: isLlamaCppGgufModel`** is intentionally broad. Whether a given GGUF loads depends on the engine version and architecture support, which we cannot express as a predicate, so we would rather show the button widely and keep the engine catching up than hide QVAC from models it can already run.
- **`mainTask: "text-generation"`** understates what the same server does. Beyond chat, `qvac serve openai` also implements `/v1/embeddings`, `/v1/audio/transcriptions`, `/v1/audio/translations`, `/v1/audio/speech` (TTS), `/v1/images/generations`, `/v1/images/edits` and `/v1/videos`, all fully on-device. If you would like QVAC surfaced on text-to-image, ASR, TTS or text-to-video model pages too, we are glad to extend this entry or add more, just tell us the shape you prefer.

## Test plan
- `pnpm --filter @huggingface/tasks exec vitest run src/local-apps.spec.ts` (20 passed, includes 2 new QVAC cases: `{{GGUF_FILE}}` placeholder and resolved-filename)
- `pnpm --filter @huggingface/tasks format:check`
- `pnpm --filter @huggingface/tasks check`

Note the `qvac` app key will need its logo asset on your side, as with previous additions. Happy to adjust the label, docs URL or predicate.

### DIFF
--- a/packages/tasks/src/local-apps.spec.ts
+++ b/packages/tasks/src/local-apps.spec.ts
@@ -363,4 +363,37 @@ curl -X POST "http://localhost:8000/v1/chat/completions" \\
 			{ label: "Releases", url: "https://github.com/bartowski/Llama-3.2-3B-Instruct-GGUF/releases" },
 		]);
 	});
+	it("qvac", async () => {
+		const { snippet: snippetFunc } = LOCAL_APPS.qvac;
+		const model: ModelData = {
+			id: "bartowski/Llama-3.2-3B-Instruct-GGUF",
+			tags: ["conversational"],
+			gguf: { total: 1, context_length: 4096 },
+			inference: "",
+		};
+		const snippet = snippetFunc(model);
+
+		expect(snippet[0].setup).toContain("npm install -g @qvac/cli");
+		expect(snippet[0].content).toContain('"type": "llamacpp-completion"');
+		expect(snippet[0].content).toContain(
+			'"src": "https://huggingface.co/bartowski/Llama-3.2-3B-Instruct-GGUF/resolve/main/{{GGUF_FILE}}"',
+		);
+		expect(snippet[1].content).toContain("qvac serve openai --model Llama-3.2-3B-Instruct-GGUF");
+	});
+
+	it("qvac - resolved gguf file", async () => {
+		const { snippet: snippetFunc } = LOCAL_APPS.qvac;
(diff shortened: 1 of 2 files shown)


thomasblc:
Friendly ping on this one, happy to make any change you would like.

Quick recap for whoever picks it up: it adds QVAC (Tether's open-source local AI SDK, Apache 2.0) to the local apps list. Two files touched, `local-apps.ts` plus its spec, +75 lines, no other surface. CI is green.

One thing worth flagging, since it makes the snippet look different from the neighbours: QVAC is the inference engine, not a client of one, so there is no separate llama.cpp server step. The CLI serves the Hub weights itself, which is why the snippet is two steps rather than three.

I verified it end to end before opening the PR: `unsloth/Qwen3-0.6B-GGUF` served straight from its `huggingface.co/.../resolve/main` URL answered on `/v1/chat/completions` (`finish_reason: stop`, 73 completion tokens).

Docs, if useful for review: https://docs.qvac.tether.io/

PR #49, opened 2023-03-05
add publish to GPR support

o-az:
This adds support for publishing to GitHub Package Registry (https://github.com/huggingface/huggingface.js/issues/47).

I can merge the two `'*-publish.yml` workflows into one once `changesets` support is added 💫

PR #2304, opened 2026-07-21
[Inference Providers] deepinfra: add text-to-speech support

ovuruska:
## What

Adds `text-to-speech` support to the **deepinfra** inference provider.

## Details

- New `DeepInfraTextToSpeechTask` (implements `TextToSpeechTaskHelper`) in `packages/inference/src/providers/deepinfra.ts`.
- Targets DeepInfra's OpenAI-compatible endpoint `POST /v1/openai/audio/speech` and returns the raw audio `Blob`.
- The mapped provider model is applied **last** in the payload so caller `parameters` cannot override it (consistent with the review feedback on the deepinfra ASR integration).
- `voice` is model-specific and optional, so it is passed through untouched and the API surfaces a clear error if a model requires one.
- Registered under `deepinfra` → `text-to-speech` in `getProviderHelper.ts`.
- Added a `textToSpeech` case to the existing `DeepInfra` test block.

Mirrors the existing `TogetherTextToSpeechTask` (also OpenAI-compatible `/v1/audio/speech` → Blob).

## Validation

- `pnpm run format:check` ✅
- `pnpm run lint` ✅
- `pnpm exec tsc --noEmit` ✅ (no new errors)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Low Risk**
> Small, additive provider task following the existing Together TTS pattern; no auth or core routing changes beyond registering a new task helper.
> 
> **Overview**
> Adds **text-to-speech** for the **deepinfra** inference provider so `InferenceClient.textToSpeech` can route to DeepInfra when models are mapped to that task.
> 
> A new `DeepInfraTextToSpeechTask` calls DeepInfra’s OpenAI-compatible `POST v1/openai/audio/speech`, maps `inputs` → `input`, merges optional `parameters` (e.g. `voice`), and sets **`model` last** so callers cannot override the mapped provider model. Responses are returned as a raw audio **`Blob`**. The task is registered on the deepinfra provider map, and the DeepInfra integration tests include a `textToSpeech` case (Kokoro model + `af_bella` voice).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ed6ccf4018444b8ad819326405fc020ad602353d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #2241, opened 2026-06-18
[Inference Providers] deepinfra: add automatic-speech-recognition support

ovuruska:
> [!NOTE]
> **Low Risk**
> Scoped provider addition with input validation and response parsing; no changes to auth or shared routing beyond registering a new task helper.
> 
> **Overview**
> Adds **automatic-speech-recognition** for the **deepinfra** inference provider so `InferenceClient.automaticSpeechRecognition({ provider: "deepinfra", ... })` can transcribe audio via DeepInfra’s OpenAI-compatible API.
> 
> Registers `DeepInfraAutomaticSpeechRecognitionTask` in the provider map and implements a helper that POSTs multipart form data to `v1/openai/audio/transcriptions`, normalizing `Blob`/`ArrayBuffer` input, mapping MIME types to file extensions for the upload, and mapping API `text` plus optional `segments` into the standard ASR output (`text` and `chunks` with timestamps).
> 
> Integration tests cover DeepInfra ASR with a hardcoded Nemotron ASR model mapping and a sample WAV file.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 00247a307b7c26d977298131960dde9b9ef8d182. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #2343, opened 2026-08-04
Add Infersia as an inference provider

infersia:
Adds [Infersia](https://infersia.com) as an inference provider, per step 2 of [register-as-a-provider](https://huggingface.co/docs/inference-providers/en/register-as-a-provider).

Infersia serves open-weight models (DeepSeek, Qwen, StepFun) over an OpenAI-compatible API, with the serving quantisation published per endpoint and DeepSeek V4 Flash offered at its full 1,048,576-token context.

## Implementation

A plain `BaseConversationalTask` with no overrides. I checked whether a `preparePayload` override was needed the way `cerebras.ts` strips `store` — it isn't: the endpoint accepts `store`, `logit_bias`, `user`, `parallel_tool_calls` and `metadata` and returns 200 for each.

| File | |
|---|---|
| `providers/infersia.ts` | The task helper |
| `types.ts` | `INFERENCE_PROVIDERS` + `PROVIDERS_HUB_ORGS` |
| `lib/getProviderHelper.ts` | Import + `PROVIDERS` entry |
| `providers/consts.ts` | `HARDCODED_MODEL_INFERENCE_MAPPING` entry |
| `README.md` | Both provider lists |
| `test/InferenceClient.spec.ts` | Conversational + streaming tests |

Provider slug is `infersia` and the Hub org is `infersia-com`, which differ — following the existing `together` → `togethercomputer` and `cohere` → `CohereLabs` precedent. Happy to align them if you'd rather.

One note for whoever writes the next provider: `consts.ts` is a fourth registration point that the guide doesn't mention. `HARDCODED_MODEL_INFERENCE_MAPPING` is a `Record<InferenceProvider, …>`, so omitting it is a `tsc` failure rather than a runtime one — easy to miss if you follow the guide's three steps and don't run `pnpm check`.

## Testing

Run against the real endpoint by temporarily un-skipping the suite:

- `pnpm check` (tsc), `pnpm lint:check` (eslint), `pnpm format:check` (oxfmt) — all clean
- `pnpm test` — 37 passed, no regressions
- Both new tests executed live, **six consecutive green runs**

The tests set `temperature: 0`, which is deliberate rather than decorative. Copying the Scaleway block verbatim gave me a flaky test on two counts: its stream case sends a bare `system` message with no user turn, which makes any model ramble, and at default temperature DeepSeek answered `"6"` instead of `"2"` roughly one run in five. Sending a user turn and pinning temperature fixed both. Mentioning it because the same flake is latent in a few of the existing provider blocks.

## Status

Steps 1, 2 and 4 of the registration guide are done our side — the endpoint is OpenAI-compatible, this PR is step 2, and we already serve an `Inference-Id` header on every response (including streaming) plus a billing endpoint returning integer nano-USD.

Our Hub org `infersia-com` is on a Team plan, so we're ready for step 3 (Model Mapping API) once someone can enable us server-side. I'll follow up on [discussion #49](https://huggingface.co/spaces/huggingface/HuggingDiscussions/discussions/49) rather than clutter this PR with it.

---

**Disclosure:** I work on Infersia, so this is a vendor-submitted integration.

---

> [!NOTE]
> **Low Risk**
> Additive provider integration following existing patterns; no changes to core client auth or routing beyond a new optional `infersia` target.
> 
> **Overview**
> Adds **Infersia** (`provider: "infersia"`) as a third-party inference provider so `InferenceClient` can route requests to `https://api.infersia.com` using the same registration pattern as other partners (`INFERENCE_PROVIDERS`, `PROVIDERS_HUB_ORGS` → `infersia-com`, `HARDCODED_MODEL_INFERENCE_MAPPING`, and `getProviderHelper`).
> 
> New `providers/infersia.ts` wires three tasks: **conversational** via `BaseConversationalTask` on OpenAI-style `/v1` routes; **automatic-speech-recognition** via multipart `v1/audio/transcriptions` (Blob/ArrayBuffer input, MIME-based filename extension, optional segment → `chunks` mapping); and **text-to-speech** via `v1/audio/speech` returning a `Blob`. README lists Infersia and its partner models API; `InferenceClient.spec.ts` adds live-style tests for chat, streaming chat, ASR, and TTS with hardcoded model mappings.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c4ecd10d1b8f0f192ae39afbda7e9bf4e523eb81. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


infersia:
Pushed `c4ecd10d` — adds `automatic-speech-recognition` and `text-to-speech` helpers, so this PR now covers all three tasks we serve rather than conversational alone.

We put two audio models into production today: `openai/whisper-large-v3-turbo` behind `/v1/audio/transcriptions` and `hexgrad/Kokoro-82M` behind `/v1/audio/speech`. Without these helpers a mapping for either could be registered but never routed.

Shaped after the DeepInfra precedent (#2241 for ASR, #2304 for TTS) — same three files, same class structure. Two intentional differences:

- Our OpenAI routes sit directly under `/v1`, not under `/v1/openai/`, so the routes are `v1/audio/transcriptions` and `v1/audio/speech`.
- The tests use one static `HARDCODED_MODEL_INFERENCE_MAPPING` entry per model rather than reassigning the record inside each test. Our four models have distinct `hfModelId`s, so a single object works and avoids mutating shared state inside a `describe.concurrent` block.

`model` is spread last in both payloads, so a caller-supplied `parameters.model` cannot override the mapped provider id — I checked that explicitly, since it came up in review on the DeepInfra ASR PR.

Verified offline against the real `makeRequestOptions` path: ASR emits `FormData` with `file` and `model` and no explicit `Content-Type`, so `fetch` sets the multipart boundary; TTS emits JSON. The ASR response mapping handles both shapes our API returns — `verbose_json` carries `segments`, which map to `chunks`, and the default `json` returns `{ text }` alone.

Checks pass rebased onto current `main`: eslint and `tsc` clean, `pnpm -r format` no diff, and the suites green (inference 25 passed/122 skipped, tasks-gen 114/114). Worth noting three `tasks-gen` snippet fixtures fail against our fork's `main` because it trails upstream by ten commits, including e073dc78 which updated exactly those fixtures — they pass once rebased.

Disclosure: I work on Infersia.

PR #2345, opened 2026-08-04
Add multi-range support and update ReconstructionInfo schema (xet v2)

coyotte508:
Based on https://github.com/huggingface/huggingface.js/pull/2265/changes

Removing fallbacks, also keep the request streaming to limit memory footprint

---

> [!NOTE]
> **Medium Risk**
> Touches the core Hub Xet download path and changes the CAS reconstruction contract, but behavior is guarded by explicit errors and extensive new tests.
> 
> **Overview**
> Updates **Xet blob downloads** to the v2 reconstruction API and the new `xorbs` schema: each signed URL lists one or more `{ chunks, bytes }` ranges instead of the old `fetch_info` single `range` + `url_range` pair. Reconstruction calls use **`/v2/reconstructions/`** (rewriting v1 link URLs when needed).
> 
> When a fetch entry has **multiple signed byte ranges**, `XetBlob` issues one HTTP request with a combined `Range` header and **requires** a `multipart/byteranges` response. A new **`StreamingMultipartParser`** streams and splits parts without buffering the full body; each part is decoded through shared **`parseChunkHeader` / `decompressChunk` / `decodePartChunks`** helpers with strict checks (part count, chunk counts, `unpacked_length`). If the CDN returns a non-multipart body for a multi-range request, the client **throws** rather than guessing—servers that lack multi-range support are expected to emit separate single-range entries instead.
> 
> Supporting utilities: variadic **`combineUint8Arrays`**, Boyer–Moore **`indexOfSubsequence`**, plus broad **`XetBlob`** and parser tests for happy path and failure modes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit aefc52de1f43c2790cafef32ecf3352359a78b28. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/hub/src/utils/XetBlob.spec.ts
+++ b/packages/hub/src/utils/XetBlob.spec.ts
@@ -1,9 +1,15 @@
-import { describe, expect, it } from "vitest";
+import { beforeEach, describe, expect, it } from "vitest";
 import type { ReconstructionInfo } from "./XetBlob";
-import { bg4_regroup_bytes, bg4_split_bytes, XetBlob } from "./XetBlob";
+import { bg4_regroup_bytes, bg4_split_bytes, clearMultiRangeSupportCache, XetBlob } from "./XetBlob";
 import { sum } from "./sum";
 
 describe("XetBlob", () => {
+	beforeEach(() => {
+		// Multi-range support detection is cached per host at module scope;
+		// reset it so tests don't leak state into each other.
+		clearMultiRangeSupportCache();
+	});
+
 	it("should handle empty files (size 0) without making network requests", async () => {
 		let fetchCount = 0;
 
@@ -176,14 +182,13 @@ describe("XetBlob", () => {
 		// 				"range": { "start": 0, "end": 5 },
 		// 			},
 		// 		],
-		// 	"fetch_info":
+		// 	"xorbs":
 		// 		{
 		// 			"be748f77930d5929cabd510a15f2c30f2f460b639804ef79dea46affa04fd8b2":
 		// 				[
 		// 					{
-		// 						"range": { "start": 0, "end": 5 },
 		// 						"url": "...",
-		// 						"url_range": { "start": 0, "end": 2839 },
+		// 						"ranges": [{ "chunks": { "start": 0, "end": 5 }, "bytes": { "start": 0, "end": 2839 } }],
 		// 					},
 		// 				],
 		// 		},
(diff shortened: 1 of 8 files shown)


coyotte508:
Benchmarked @assafvayner :

- local home FR
- french VPS
- internal prod (no multirange support)

v2 is slightly slower than v1 and uses more RAM, but makes less requests. v2 streamed was the right call, can get more improvement after https://github.com/huggingface/huggingface.js/pull/2337 probably

> All cleaned up (pod, coyo, local worktrees + bundles). Here's the comprehensive matrix:
> 
> ## v1 vs v2-streamed vs v2-buffered — 3 networks
> 
> **Correctness: identical SHAs across all 18 runs.** ✅
> 
> ### Qwen2.5-0.5B `model.safetensors` (942MB out, 814MB fetched)
> 
> | location | version | time | speed | xorb reqs | peak RSS |
> |---|---|--:|--:|--:|--:|
> | **local** (~home) | v1 | 20.0s | 47.1 MB/s | 24 | 163MB |
> | | v2-streamed | 20.5s | 46.0 MB/s | **17** | 238MB |
> | | v2-buffered | 23.4s | 40.2 MB/s | **17** | 273MB |
> | **coyo** (VPS) | v1 | 16.4s | 57.5 MB/s | 24 | 296MB |
> | | v2-streamed | 18.5s | 50.9 MB/s | **17** | 286MB |
> | | v2-buffered | 23.2s | 40.6 MB/s | **17** | 302MB |
> | **pod** (internal) | v1 | 10.8s | 87.3 MB/s | 24 → **s3** | 310MB |
> | | v2-streamed | 9.9s | **95.4 MB/s** | 24 → **s3** | 266MB |
> | | v2-buffered | 10.1s | 92.9 MB/s | 24 → **s3** | 330MB |
> 
> (SmolLM2 257MB follows the same pattern: v2 = 7 reqs vs v1's 8 externally, equal internally; times 3.0–10.7s by location.)
> 
> ## The headline findings
> 
> **1. The pod result validates the whole server-side contract, live in prod.** On the internal network the CAS is `cas-server-direct` and xorb URLs point at **raw S3** — and v2's request count equals v1's (24), because the server detected a non-multi-range backend and emitted **one URL per range at signing time** (`supports_multirange() == false`), exactly as the xetcas code promised. Our client transparently took the single-range streaming path — no multipart, no fallback logic needed, SHAs match. This is empirical confirmation that dropping the client-side "CDN ignored multi-range" fallback was the right call.
> 
> **2. Externally (CloudFront), v2's −29% requests is consistent** across local and coyo, with no extra bytes transferred.
> 
> **3. Wall time: v2 ≈ v1 everywhere, and fastest-of-three on the internal network** (95.4 MB/s). Externally v1 occasionally edges ahead within noise — with serial large requests, request-count savings just don't convert to time yet. They will with concurrency.
> 
> **4. Streamed vs buffered:** streamed is faster or equal in 5 of 6 pairings (buffered notably slower on coyo's Qwen: 23.2s vs 18.5s) and peaks lower on RSS in 5 of 6 — modest deltas because production multi-range entries are small, consistent with the earlier local finding: streaming is worst-case insurance + concurrency groundwork, at zero cost.
> 
> **5. Bandwidth ranking behaved as expected:** internal pod (87–95 MB/s) > coyo (40–58) > local (24–47), and sequential runs kept the comparisons contention-free.

PR #2346, opened 2026-08-04
Switch from secret to OIDC auth

coyotte508:
> [!NOTE]
> **Medium Risk**
> CI authentication now depends on Hugging Face trusted-publisher/OIDC configuration; misconfiguration or exchange failures cause tests to run without credentials rather than failing the job outright.
> 
> **Overview**
> CI no longer injects **`secrets.HF_TOKEN`** into the **node**, **browser**, and **e2e** jobs in `.github/workflows/test.yml`. Each job now requests **`id-token: write`** and **`contents: read`**, installs the **hf CLI**, and exchanges the GitHub OIDC identity for a short-lived Hub token via **`HF_OIDC_RESOURCE`** (trusted publisher for user `coyotte508`). The token is written to **`GITHUB_ENV`** only after a strict **`hf_...`** regex check so malformed CLI output cannot poison the env file; exchange failures log a warning and tests continue **unauthenticated**.
> 
> E2E steps that previously passed the secret explicitly now rely on that job-level **`HF_TOKEN`**: the TypeScript smoke test uses shell expansion (`token="$HF_TOKEN" pnpm start`), and the Deno step drops the per-step **`HF_TOKEN`** env mapping while still allowing **`HF_TOKEN`** from the environment.
> 
> A one-line JSDoc comment was added above **`HUB_URL`** in `packages/hub/src/consts.ts`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8885fe60ef988a7db07d8f3f8d221ac8837d0aa3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,6 +9,9 @@ jobs:
   node:
     runs-on: ubuntu-latest
     timeout-minutes: 10
+    permissions:
+      id-token: write # needed to mint the OIDC token exchanged for a short-lived HF token (trusted publishers)
+      contents: read
 
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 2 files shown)


coyotte508:
E2E fails, maybe sign we need to extend OIDC with eg oauth scopes, like `inference-api`

Still, improvement

PR #2349, opened 2026-08-05
[Inference Providers] fix fal-ai image-text-to-video URL rewrite breaking routed calls

hanouticelina:
## Problem

`FalAIImageTextToVideoTask.preparePayloadAsync` (and its `image-text-to-image` counterpart) drops the last path segment of the request URL when no image is provided:

```ts
urlTransform: (url) => {
    const urlObj = new URL(url);
    urlObj.pathname = urlObj.pathname.split("/").slice(0, -1).join("/");
    return urlObj.toString();
},
```

The intent is to reach the text-only variant of a fal endpoint (`fal-ai/flux-2/edit` → `fal-ai/flux-2`), which works when calling fal directly.

It does not work when routing through huggingface.co: there, the URL path **is** the provider model id the router resolves the mapping from, so rewriting it makes the model unresolvable. The request fails with `Model not supported by provider fal-ai` before ever reaching fal — and no fal mapping registers the stripped parent path (`fal-ai/flux-2`, `minimax/h3`, … are all absent).

The first model to hit this is `MiniMaxAI/MiniMax-H3` (`minimax/h3/image-to-video`), the first fal `image-text-to-video` mapping. A prompt-only call was sent to `/fal-ai/minimax/h3`:

```
POST https://router.huggingface.co/fal-ai/minimax/h3?_subdomain=queue
→ 400 {"error":"Model not supported by provider fal-ai"}
```

Surfaced in the model page widget, which offers "Enter a prompt, an image, or both", as `Failed to perform inference: Model not supported by provider fal-ai`.

## Fix

Keep the rewrite for direct fal calls (behaviour unchanged) and skip it when the request goes through the router. Both duplicated closures now share one documented helper, so the FLUX.2-specific rationale lives in one place instead of being silently inherited.

## Verification

Live, against `router.huggingface.co` and fal:

| case | before | after |
|---|---|---|
| prompt only | `POST /fal-ai/minimax/h3` → 400 `Model not supported by provider fal-ai` | `POST /fal-ai/minimax/h3/image-to-video` → 422 `image_url: Field required` |
| image + prompt | correct URL already | unchanged, returns `video/mp4, 2256308 bytes` |
| prompt only, direct fal key | `queue.fal.run/minimax/h3` | unchanged |

The 422 is the correct outcome: fal's OpenAPI has `minimax/h3/image-to-video` requiring both `prompt` and `image_url` (prompt-only lives at `minimax/h3/text-to-video`). The fix stops the SDK from corrupting the URL and lets the provider's real error through instead of a misleading routing failure.

`test/fal-ai-url.spec.ts` pins both sides of the guard. Run against the pre-fix code, exactly the two routed assertions fail and the three direct-key/with-image ones pass — so direct-fal behaviour is provably unchanged.

---

> [!NOTE]
> **Low Risk**
> Targeted URL-building fix in fal-ai inference with tests; no auth or data-handling changes.
> 
> **Overview**
> **Fixes prompt-only fal-ai `image-text-to-image` and `image-text-to-video` calls when using the Hugging Face router.** Those tasks used to always strip the last URL path segment (e.g. `minimax/h3/image-to-video` → `minimax/h3`) so direct fal calls could hit a text-only endpoint. On `router.huggingface.co` the path is the provider model id used for mapping, so that rewrite caused `Model not supported by provider fal-ai`.
> 
> The change introduces `dropEndpointSegmentOnDirectCalls`, which **only** drops the last segment when the request is **not** routed through the router; router URLs are left unchanged. Direct fal behavior and image+prompt URLs are unchanged. **`test/fal-ai-url.spec.ts`** locks in router vs direct-key URL behavior for both tasks.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d61b9372c998e11576cf79a1edae2f9847335e1c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/inference/src/providers/fal-ai.ts
+++ b/packages/inference/src/providers/fal-ai.ts
@@ -37,7 +37,7 @@ import {
 	type TextToVideoTaskHelper,
 	type ImageToVideoTaskHelper,
 } from "./providerHelper.js";
-import { HF_HUB_URL } from "../config.js";
+import { HF_HUB_URL, HF_ROUTER_URL } from "../config.js";
 import type { AutomaticSpeechRecognitionArgs } from "../tasks/audio/automaticSpeechRecognition.js";
 import type { AudioToAudioArgs, AudioToAudioOutput } from "../tasks/audio/audioToAudio.js";
 import {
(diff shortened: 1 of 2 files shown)


hanouticelina:
✅ tested the fix, it's working as expected and backward compatible. merging!

PR #2353, opened 2026-08-06
fix(mcp-client): preserve model reasoning between tool rounds

### LINKED ISSUE #2334: McpClient drops model reasoning between tool-calling rounds
`McpClient` (`packages/mcp-client/src/McpClient.ts`) correctly replays tool calls and results back into `this.messages` across rounds of its agentic loop (the assistant `tool_calls` message plus `role: "tool"` results, around lines 164-211), but it never captures or re-attaches the model's reasoning/thinking content on those replayed messages. Searching `packages/mcp-client`, `packages/tiny-agents`, and `packages/inference` for `reasoning`, `reasoning_content`, `reasoning_text`, or `thinking` returns no matches.

For preserved-thinking models this drops the chain of thought between tool rounds of the same task. Moonshot documents this as a hard requirement for the Kimi K2/K3 family (assistant `reasoning_content` from prior turns must be sent back verbatim during multi-step tool calling, exposed as the `thinking.keep` API parameter): https://platform.kimi.ai/docs/guide/use-kimi-k2-thinking-model. Their own reference harness (`MoonshotAI/kimi-cli`) implements exactly this: re-attaching captured thinking as `reasoning_content` on each assistant message, and omitting `content` entirely when nothing visible remains (some backends 400 on an empty text part next to `tool_calls`).

Other clients have hit real bugs from the same gap: BerriAI/litellm#26156, BerriAI/litellm#21672, open-webui/open-webui#23175, strands-agents/sdk-python#1150.

Suggested fix: when building the assistant message for the next round in `McpClient.processSingleTurnWithTools` (or equivalent), capture any streamed reasoning/thinking delta from the completion and re-attach it as `reasoning_content` on the pushed assistant message, mirroring what the model itself streamed. Omit `content` when it would otherwise be empty next to `tool_calls`.

Scope note: only `McpClient`'s in-loop history was checked; `tiny-agents`' cross-session persistence (if any) wasn't audited.

lchetanlc:
Fixes #2334

**Description**
When using the `McpClient` with reasoning-enabled models (like Kimi K2/K3), the model's reasoning/thinking content was previously dropped between tool-calling rounds. This broke the requirement that some models have for preserving thinking history across turns. Furthermore, some backends return a 400 error if an empty `content` string is sent alongside `tool_calls`.

This PR:
- Captures `reasoning_content` from the streamed chunks in `McpClient`.
- Appends `reasoning_content` to the `assistantMessage` that gets pushed into the message history for subsequent rounds.
- Conditionally omits the `content` property entirely if it is empty and there are `tool_calls` present to prevent API 400 errors.

**Testing**
All unit tests in `mcp-client` pass successfully.

---

> [!NOTE]
> **Low Risk**
> Localized change to assistant message assembly in the MCP client loop; no auth or data-path changes, with behavior aligned to reasoning-model API expectations.
> 
> **Overview**
> **`McpClient.processSingleTurnWithTools`** now keeps **reasoning model** history intact across MCP tool rounds by aggregating streamed **`reasoning_content`** and copying it onto the assistant message pushed into **`messages`** for the next completion.
> 
> When the model returns **`tool_calls`** with no visible text, **`content`** is **omitted** instead of sending an empty string, avoiding **400** responses from some inference backends.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ac15be068e3ddc48c25e110fa213cc2455d5aa48. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/mcp-client/src/McpClient.ts
+++ b/packages/mcp-client/src/McpClient.ts
@@ -138,10 +138,10 @@ export class McpClient {
 			},
 		);
 
-		const message = {
+		const message: ChatCompletionInputMessage & { reasoning_content?: string } = {
 			role: "unknown",
 			content: "",
-		} satisfies ChatCompletionInputMessage;
+		};
 		const finalToolCalls: Record<number, ChatCompletionStreamOutputDeltaToolCall> = {};
 		let numOfChunks = 0;
 
@@ -160,6 +160,9 @@ export class McpClient {
 			}
 			if (delta.content) {
 				message.content += delta.content;
+			}
+			if (typeof delta.reasoning_content === "string") {
+				message.reasoning_content = (message.reasoning_content ?? "") + delta.reasoning_content;
 			}
 			for (const toolCall of delta.tool_calls ?? []) {
 				// aggregating chunks into an encoded arguments JSON object
@@ -187,8 +190,11 @@ export class McpClient {
 
 		const assistantMessage: ChatCompletionInputMessage = {
 			role: "assistant",
-			content: message.content,
 		};
+
+		if (message.reasoning_content !== undefined) {
+			assistantMessage.reasoning_content = message.reasoning_content;
+		}
 
 		const finalToolCallValues = Object.values(finalToolCalls);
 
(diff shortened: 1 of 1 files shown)


coyotte508:
Should we just archive https://www.npmjs.com/package/@huggingface/mcp-client and https://www.npmjs.com/package/@huggingface/tiny-agents @julien-c ? 60/50 weekly downloads

PR #2354, opened 2026-08-07
Add cortiq to the model libraries

infosave2007:
`cortiq` reads **CMF**, a single-file model format: quantized weights, the tokenizer and the chat template in one memory-mapped file that verifies its own per-tensor hashes. The runtime is a small Rust binary with no ML framework under it — no torch, no ONNX, no CUDA install, no C++ toolchain — running on CPU everywhere and on GPU through wgpu (Vulkan / DX12 / Metal).

- Source: https://github.com/infosave2007/cmf (Apache-2.0)
- Format specification: [`docs/CMF_V2_SPEC.md`](https://github.com/infosave2007/cmf/blob/master/docs/CMF_V2_SPEC.md)
- Runtime: https://crates.io/crates/cortiq-cli, plus prebuilt binaries for six targets on the releases page

### Models on the Hub

12 models carry `library_name: cortiq` today, across three tasks:

| task | models |
|---|---|
| `text-generation` | Qwen3.6-27B, Qwen3.6-35B-A3B (+ a 2-bit build), KAT-Coder-V2.5, Kimi-Linear-48B-A3B, DeepSeek-V4-Flash, Nanbeige4.2-3B, Bonsai 1.7B / 8B / 27B |
| `text-to-image` | Lumina-Image-2.0 |
| `text-to-video` | MiniMax-H3 Turbo (video **with synchronized audio**) |

The snippet branches on `pipeline_tag` accordingly: `cortiq run` / `cortiq serve`, `cortiq imagine`, `cortiq animate`.

### Two notes on the entry

**No `filter` requested.** The field is documented for libraries with more than 100 models on the Hub; this one has 12.

**`countDownloads` carries an `OR`.** One repository ships a 455 GB model as concatenable slices under `parts-*/` rather than as `.cmf` files, so `path_extension:"cmf"` alone would miss it.

Happy to adjust naming, the snippet, or the download query to whatever fits your conventions.

---

> [!NOTE]
> **Low Risk**
> Additive library metadata and example snippets only; no runtime, auth, or data-handling changes.
> 
> **Overview**
> Registers **cortiq** as a Hub model library so CMF models get library metadata, download counting on `.cmf` files, and usage snippets.
> 
> Snippets install `cortiq-cli`, download `*.cmf` weights, then branch on `pipeline_tag`: `imagine` (text-to-image), `animate` (text-to-video), `music` (text-to-audio), or `run`/`serve` for text generation.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 19f536fd02a6ab5829ee2896d8c932a7569fd942. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/tasks/src/model-libraries-snippets.ts
+++ b/packages/tasks/src/model-libraries-snippets.ts
@@ -244,7 +244,7 @@ print("R² Score:", r2)`;
 };
 
 export const cortiq = (model: ModelData): string[] => {
-	const setup = `# one Rust binary: no torch, no ONNX, no CUDA install, no Python
+	const setup = `# one Rust binary, no additional dependencies
 cargo install cortiq-cli   # or a prebuilt binary from github.com/infosave2007/cmf/releases
 hf download ${model.id} --include "*.cmf" --local-dir .
 ls *.cmf                   # some repos ship more than one quantization`;
--- a/packages/tasks/src/model-libraries-snippets.ts
+++ b/packages/tasks/src/model-libraries-snippets.ts
@@ -254,6 +254,13 @@ ls *.cmf                   # some repos ship more than one quantization`;
 	if (model.pipeline_tag === "text-to-video") {
 		return [setup, `cortiq animate FILE.cmf --prompt "a corgi in a chef hat flipping a pancake" --out clip.avi`];
 	}
+	if (model.pipeline_tag === "text-to-audio") {
+		return [
+			setup,
+			`cortiq music FILE.cmf --prompt "dream pop, warm analog synths, brushed drums" \\
+  --lyrics "[verse] the tide came in and took the map" --seconds 20 --out song.wav`,
+		];
+	}
 	return [
 		setup,
 		`cortiq run FILE.cmf --prompt "What is the capital of France?"`,
--- a/packages/tasks/src/model-libraries-snippets.ts
+++ b/packages/tasks/src/model-libraries-snippets.ts
@@ -243,6 +243,25 @@ print("R² Score:", r2)`;
 	return [installSnippet, classificationSnippet, regressionsSnippet];
 };
 
+export const cortiq = (model: ModelData): string[] => {
+	const setup = `# one Rust binary: no torch, no ONNX, no CUDA install, no Python
+cargo install cortiq-cli   # or a prebuilt binary from github.com/infosave2007/cmf/releases
+hf download ${model.id} --include "*.cmf" --local-dir .
+ls *.cmf                   # some repos ship more than one quantization`;
+	if (model.pipeline_tag === "text-to-image") {
+		return [`${setup}\n\ncortiq imagine FILE.cmf --prompt "a red fox in a snowy forest" --out fox.ppm`];
+	}
+	if (model.pipeline_tag === "text-to-video") {
+		return [
+			`${setup}\n\ncortiq animate FILE.cmf --prompt "a corgi in a chef hat flipping a pancake" --out clip.avi`,
+		];
+	}
+	return [
+		`${setup}\n\ncortiq run FILE.cmf --prompt "What is the capital of France?"`,
+		`${setup}\n\ncortiq serve FILE.cmf --port 8080   # OpenAI-compatible server`,
+	];
+};
+
 export const cxr_foundation = (): string[] => [
 	`# pip install git+https://github.com/Google-Health/cxr-foundation.git#subdirectory=python
 
--- a/packages/tasks/src/model-libraries-snippets.ts
+++ b/packages/tasks/src/model-libraries-snippets.ts
@@ -249,16 +249,15 @@ cargo install cortiq-cli   # or a prebuilt binary from github.com/infosave2007/c
 hf download ${model.id} --include "*.cmf" --local-dir .
 ls *.cmf                   # some repos ship more than one quantization`;
 	if (model.pipeline_tag === "text-to-image") {
-		return [`${setup}\n\ncortiq imagine FILE.cmf --prompt "a red fox in a snowy forest" --out fox.ppm`];
+		return [setup, `cortiq imagine FILE.cmf --prompt "a red fox in a snowy forest" --out fox.ppm`];
 	}
 	if (model.pipeline_tag === "text-to-video") {
-		return [
-			`${setup}\n\ncortiq animate FILE.cmf --prompt "a corgi in a chef hat flipping a pancake" --out clip.avi`,
-		];
+		return [setup, `cortiq animate FILE.cmf --prompt "a corgi in a chef hat flipping a pancake" --out clip.avi`];
 	}
 	return [
-		`${setup}\n\ncortiq run FILE.cmf --prompt "What is the capital of France?"`,
-		`${setup}\n\ncortiq serve FILE.cmf --port 8080   # OpenAI-compatible server`,
+		setup,
+		`cortiq run FILE.cmf --prompt "What is the capital of France?"`,
+		`cortiq serve FILE.cmf --port 8080   # OpenAI-compatible server`,
 	];
 };
 
--- a/packages/tasks/src/model-libraries.ts
+++ b/packages/tasks/src/model-libraries.ts
@@ -334,6 +334,14 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		repoUrl: "https://github.com/Unbabel/COMET/",
 		countDownloads: `path:"hparams.yaml"`,
 	},
+	cortiq: {
+		prettyLabel: "cortiq",
+		repoName: "cmf",
+		repoUrl: "https://github.com/infosave2007/cmf",
+		docsUrl: "https://github.com/infosave2007/cmf/blob/master/docs/CMF_V2_SPEC.md",
+		snippets: snippets.cortiq,
+		countDownloads: `path_extension:"cmf"`,
+	},
 	cosmos: {
 		prettyLabel: "Cosmos",
 		repoName: "Cosmos",
--- a/packages/tasks/src/model-libraries.ts
+++ b/packages/tasks/src/model-libraries.ts
@@ -336,7 +336,7 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 	},
 	cortiq: {
 		prettyLabel: "cortiq",
-		repoName: "cmf",
+		repoName: "cortiq",
 		repoUrl: "https://github.com/infosave2007/cmf",
 		docsUrl: "https://github.com/infosave2007/cmf/blob/master/docs/CMF_V2_SPEC.md",
 		snippets: snippets.cortiq,


infosave2007:
Good catch from Bugbot, and it was right: `path_prefix` is the complete prefix segment — the doc comment in `model-libraries-downloads.ts` says `"prefix/"`, and the live example in this file is `path_prefix:"checkpoints/"` with the trailing slash. My `path_prefix:"parts-"` would never have matched `parts-q2tp-v3/`, so the clause was dead weight in a shared registry.

Simplified to `path_extension:"cmf"` alone. That is the honest rule for the format, and it should not carry a special case for one repository's folder names anyway — the repo in question ships a 455 GB model as concatenable slices with no extension, which is my layout problem to fix on my side, not something this entry should paper over.

Pushed as `fc8d976`; the entry is now six lines with no exceptions in it.

infosave2007:
Friendly ping — anything else needed from my side here? The workflow checks seem to be awaiting first-time-contributor approval.

Some context on why this matters right now: there are already **18 CMF models published under [infosave](https://huggingface.co/infosave)** (Qwen3.6 27B/35B-A3B, DeepSeek-V4-Flash, Kimi-Linear-48B, Gemma, Lumina image, and more), and nearly all of them show **0 downloads** despite real traffic — this week alone we pulled 45+ GB from these repos while testing. The only repo with a non-zero counter is the one that also ships files in already-counted formats. So authors publishing CMF builds are flying blind on usage today.

We also have Android and iOS apps built on the `cortiq` runtime shipping soon — Hub download stats are the main signal both for us and for model authors deciding whether to publish CMF builds.

Happy to adjust anything in the entry if the current shape doesn't fit the registry conventions.

infosave2007:
@Wauplin — sorry for the direct ping; you merged the recent library entries, so you are probably the right person to look at this one.

It seems to be stuck on something mechanical rather than on review: the `Lint` and `Test` runs are both sitting at `action_required` (first-time contributor), so CI has never actually run on this branch and the PR shows as blocked. A single "Approve and run workflows" would unstick it.

Two adjustments I am happy to push if you want them:

- **`filter: false`** — the integration guide says to set it explicitly and I left it out, since the field is documented for libraries with more than 100 models and this one has 12.
- **the snippet** currently does `hf download --include "*.cmf"`, which pulls every quantization in a repo — one of mine ships three — and so counts as several downloads for a single model. Switching it to one named file matches the download-stats guidance about not double counting.

Otherwise the entry is unchanged since the Bugbot fix: six lines, `path_extension:"cmf"`, no exceptions, and 12 models on the Hub carrying `library_name: cortiq`.

infosave2007:
@Kakulukian — sorry for the direct ping; you merged #2358 (MiniMax H3) last week, so you may be the right person while @Wauplin is away.

This one looks stuck on something mechanical rather than on review: the repo's **Lint** and **Test** workflows have never run on it — first-time-contributor approval is pending, so there is nothing to go green. One click on "Approve and run workflows" would settle that.

Happy to make it easier to say yes: if the snippet in `model-libraries-snippets.ts` is what needs a closer look, **I will gladly split it out** and leave this PR as a single-file change to `model-libraries.ts`, same shape as #2358. The library entry is the part that matters — `.cmf` files are not counted at all today, so 20+ published models show `0` downloads under real traffic. Just say the word and I will push the split.

packages/tasks/src/model-libraries.ts
```diff
@@ -334,6 +334,14 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		repoUrl: "https://github.com/Unbabel/COMET/",
 		countDownloads: `path:"hparams.yaml"`,
 	},
+	cortiq: {
+		prettyLabel: "cortiq",
+		repoName: "cmf",
```
pcuenca: We generally recommend these to match. There are some exceptions but we are trying to unify going forward.

packages/tasks/src/model-libraries.ts
```diff
@@ -334,6 +334,14 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
...
+		repoUrl: "https://github.com/infosave2007/cmf",
+		docsUrl: "https://github.com/infosave2007/cmf/blob/master/docs/CMF_V2_SPEC.md",
+		snippets: snippets.cortiq,
+		countDownloads: `path_extension:"cmf"`,
```
pcuenca, line 343: These could result in duplicate counting if users download the complete repo instead of the file they use. I think you instruct them to select the desired file, but you should keep it in mind. Many orgs choose to use a repo-per-checkpoint policy so they get granular download counts too.

packages/tasks/src/model-libraries-snippets.ts
```diff
@@ -243,6 +243,25 @@ print("R² Score:", r2)`;
 	return [installSnippet, classificationSnippet, regressionsSnippet];
 };
 
+export const cortiq = (model: ModelData): string[] => {
+	const setup = `# one Rust binary: no torch, no ONNX, no CUDA install, no Python
+cargo install cortiq-cli   # or a prebuilt binary from github.com/infosave2007/cmf/releases
+hf download ${model.id} --include "*.cmf" --local-dir .
```
pcuenca, line 249: This will result in multiple counts for the same repo, as explained in another comment.

The ideal end-state, in my opinion, would be that `cortiq` could be given a Hub repo id and quantized file specification as command-line arguments, and it would download and cache the file. But we can go with this and update later.

packages/tasks/src/model-libraries-snippets.ts
```diff
@@ -243,6 +243,25 @@ print("R² Score:", r2)`;
...
+	}
+	if (model.pipeline_tag === "text-to-video") {
+		return [
+			`${setup}\n\ncortiq animate FILE.cmf --prompt "a corgi in a chef hat flipping a pancake" --out clip.avi`,
+		];
+	}
+	return [
+		`${setup}\n\ncortiq run FILE.cmf --prompt "What is the capital of France?"`,
+		`${setup}\n\ncortiq serve FILE.cmf --port 8080   # OpenAI-compatible server`,
+	];
```
pcuenca, line 268: In all cases, I'd recommend including the setup phase as the first item of the array.

packages/tasks/src/model-libraries.ts
infosave2007: Done — `repoName` is now `cortiq`. The GitHub repo is called `cmf` and keeps that in `repoUrl`; since `repoName` is what the Hub shows rather than anything a URL is built from, matching the label is the better read anyway.

packages/tasks/src/model-libraries-snippets.ts
infosave2007: Done. The setup block is its own first entry now, and the commands follow it — previously those four lines were pasted in front of every snippet, so a reader saw the install three times.

packages/tasks/src/model-libraries-snippets.ts
infosave2007: Taking the point, and I want to be straight about the trade rather than wave it through.

The `--include "*.cmf"` form exists because a CMF repo usually ships the same weights at several quantizations — a 27B sits at 27.4 GB in 8-bit and 14.3 GB in 4-bit, and which one a person wants depends on their card. The snippet cannot know that, so it fetches the set and the `ls *.cmf` line invites the choice. The cost is exactly what you say: someone who runs it verbatim counts every file.

Your end-state is the right one and it is not far off. `cortiq` already has an HF downloader inside it — `cortiq convert --model owner/name` streams a checkpoint shard by shard from the Hub today — so teaching `cortiq run` to take `owner/repo` plus a file spec and cache it is wiring, not new machinery. I would rather land that and then simplify this snippet to one line than write a snippet now for a flag that does not exist yet.

On the repo-per-checkpoint policy: noted, and it is a fair reading of what granular counts are worth. For this family I think one repo per model with the quantizations beside each other serves people better — the choice between them is the whole point, and splitting it across repos hides the comparison. If the counting matters more than that to the Hub, say so and I will split them.

packages/tasks/src/model-libraries.ts
infosave2007: Understood on the double counting — same root as the snippet comment, and the answer is the same: once `cortiq run` can take a repo id and one file, the instruction becomes "fetch this file" rather than "fetch the cmf files", and the count follows what people actually use.

Leaving `path_extension:"cmf"` as it is for now, since it is the only clause that matches a CMF repo at all. Worth noting the Bugbot comment above it is stale — it flags a `path_prefix:"parts-"` clause from an early revision of this PR that is no longer in the diff.

infosave2007:
@pcuenca thanks for the review — both concrete points are addressed in `846c800a`, and I replied inline to all four. Summarising here since thread replies are easy to miss:

- **`repoName` now matches `prettyLabel`** (`cortiq`, not `cmf`). The GitHub repo keeps its own name in `repoUrl`.
- **The setup block is its own first array entry** instead of being pasted in front of every command — a reader was seeing the install three times.
- **Duplicate download counts**: taking the point. `--include "*.cmf"` exists because one repo carries the same weights at several quantizations (a 27B is 27.4 GB at 8-bit and 14.3 GB at 4-bit) and the snippet cannot know which the reader wants. Your end state is the right one and is close: `cortiq` already streams a checkpoint from the Hub in `convert --model owner/name`, so teaching `run` to take a repo id plus a file is wiring rather than new machinery. I would rather ship that and then cut this snippet to one line than write a snippet for a flag that does not exist.
- **Repo-per-checkpoint**: noted. For this family I think the quantizations belong side by side — choosing between them is the point — but if the Hub weighs granular counts higher, I will split them.

One thing outside my control: `Lint` and `Test` have never run on this PR, on any commit. They sit at `action_required` — first-time-contributor approval. Nothing to fix on my side, but it does mean the checks cannot go green on their own.

packages/tasks/src/model-libraries-snippets.ts
```diff
@@ -243,6 +243,24 @@ print("R² Score:", r2)`;
 	return [installSnippet, classificationSnippet, regressionsSnippet];
 };
 
+export const cortiq = (model: ModelData): string[] => {
+	const setup = `# one Rust binary: no torch, no ONNX, no CUDA install, no Python
```
pcuenca:
```suggestion
	const setup = `# one Rust binary, no additional dependencies
```

pcuenca (approved):
Looks good to me, with the limitations discussed in the review. Do let us know when the snippet can be improved and made more granular.

For future interactions, please consider answering as a human when receiving comments from humans. It's fine to submit agent-assisted PRs, but as a reviewer it feels wasteful to interact with bots.

Deferring to @Wauplin for final approval or further comments when he's back. Please, do not ping any more team members individually.

infosave2007:
Thanks for taking it the rest of the way, @pcuenca — and Bugbot caught a real one, now fixed in `19f536fd`.

**text-to-audio was falling through.** `infosave/MiniMax-Music-3-cmf` is on the Hub with `library_name: cortiq` and `pipeline_tag: text-to-audio`, and would have shown people `cortiq run` / `cortiq serve` for a model that makes songs. It now gets `cortiq music` with the caption and the lyrics, which are the two arguments nobody would guess. That is the full set of non-text pipelines in this family today: `text-to-image` (Lumina), `text-to-video` (LTX-2.5, MiniMax-H3), `text-to-audio` (MiniMax-Music-3).

**On the red Test job** — it is not this PR. One test of 430:

```
FAIL packages/hub  src/utils/XetBlob.spec.ts
  > XetBlob > parallelDownloads > settles back to serial on a bandwidth-capped link
AssertionError: expected 67 to be greater than 85.8
```

A wall-clock throttling assertion in `packages/hub`, on a branch that touches two files in `packages/tasks` and nothing else. Lint is green and the other 429 pass. Worth a re-run rather than a fix from me — I have no way to make that assertion less timing-dependent from here, and I would rather not touch a package this PR has no business in.

Wauplin (approved):
Thanks for the iterations and thanks @pcuenca for the many reviews.

@infosave2007 I'll now merge this PR. Expect 2-3 days before getting it live on the Hub. Please don't aggressively ping maintainers on PRs, this is usually detrimental to a good ecosystem and doesn't make the review faster. Opening a PR is enough, maintainers will look at them when they have the time to.

PR #2367, opened 2026-08-13
🐛 blob: use a Range probe instead of HEAD in WebBlob.create

### LINKED ISSUE #2366: WebBlob.create returns a size-0 blob in the browser (packages/blob still probes with HEAD)
In a browser, `WebBlob.create()` from `@huggingface/blob` can resolve to a blob with `size === 0`, so `.text()`, `.stream()` and `.slice()` all come back empty. `packages/blob`'s own browser suite reproduces it on current `main` — 2 of 5 tests fail:

```
$ cd packages/blob && pnpm test:browser   # headless chrome

FAIL  src/utils/WebBlob.spec.ts > WebBlob > should create a WebBlob with a slice on the entire resource
- Expected
+ Received
-   "end": 139242,
+   "end": 0,
```

Same failure in CI, on the [browser job of #2365](https://github.com/huggingface/huggingface.js/actions/runs/31666633403/job/94440804028) (that PR only added a `throw` in `slice()`; the failure is unrelated to it).

## Cause

`WebBlob.create` sizes the resource from a `HEAD` request:

https://github.com/huggingface/huggingface.js/blob/7b3e12c5000cf681bf2f2e2a10fc39471a686750/packages/blob/src/utils/WebBlob.ts#L20-L32

When CloudFront gzips a response on the fly — typical for the small text/JSON files behind `/raw/` and `/api/resolve-cache/` — the cached response carries neither `Content-Length` nor `Accept-Ranges`, and a later `HEAD` served from that cache inherits the missing headers. `Number(null)` is `0`, so `size` is `0`.

This is the same bug that was fixed for `packages/hub` in #2118 (May 2026), by probing with `Range: bytes=0-0` and reading the total size from `Content-Range`, which survives regardless of the cached encoding state. `packages/blob` still has the pre-#2118 code; the fix was never backported.

`packages/dduf` reaches the same path through `createBlob()`.

## Why CI didn't catch it

`test.yml` runs `pnpm --filter ...[<merge-base>] test:browser`, so only packages changed by a PR (and their dependents) are tested. `packages/blob` hadn't been touched since January, so its browser specs simply never ran — until #2365 touched the package last week and the browser job went red for this reason.

Happy to send the backport.

codeAnqiang-ma:
Fixes #2366

## What

`packages/hub` fixed this in #2118 by probing with `Range: bytes=0-0` instead of `HEAD`; `packages/blob` was left on the old code. This copies that fix over.

The `HEAD` probe takes the size from `Content-Length`, which CloudFront drops when it gzips a response on the fly, so in browsers `WebBlob.create` ended up with `size = 0` and handed back an empty blob. A range probe gets the total size from `Content-Range`, which survives whatever encoding the cache happens to hold.

`create()` is now the same as hub's apart from two things: no `accessToken` (this package's options don't have one), and a plain `Error` on a failed response since `packages/blob` has no `createApiError`.

## Why CI didn't catch it

`test.yml` tests only the packages changed since the merge base, and `packages/blob` hadn't been touched since January, so its browser specs never ran. #2365 touched the package last week and the browser job went red for exactly this, not for anything in that PR.

## Tests

The `beforeAll` change also comes from #2118: it took the reference size from a `HEAD` request too, so the spec was measuring the same unreliable header it was asserting against.

Two new tests inject a `fetch` so they don't depend on the CDN's cache state:

- 206 without `Content-Length` → size comes from `Content-Range`. Fails before this change (`create` returns a 1-byte buffered blob instead of a `WebBlob`).
- server ignores the range and answers 200 → body is buffered.

Results:

- `packages/blob`: `pnpm test` 10/10, `pnpm test:browser` 7/7 — the latter was `2 failed | 3 passed` before
- `packages/dduf` (the only dependent, reaches this through `createBlob`): `pnpm test` and `pnpm test:browser` both pass
- `lint:check`, `check` (tsc) and `format:check` clean

## Disclosure

This fix was prepared with AI assistance; every change and test result was reproduced and reviewed locally before submission.

---

> [!NOTE]
> **Low Risk**
> Targeted fetch-probe change in blob utilities with regression tests; no auth or data-model changes.
> 
> **Overview**
> **`WebBlob.create`** no longer uses a **`HEAD`** request to read **`Content-Length`** and **`Accept-Ranges`**. It issues a **`GET`** with **`Range: bytes=0-0`**, derives total size from **`Content-Range`** on **206**, cancels the probe body for large files, and either returns a lazy **`WebBlob`** or buffers via a full fetch when below **`cacheBelow`**. If the server answers **200** and ignores the range, it returns **`probe.blob()`** instead of starting a second download.
> 
> Browser specs now size the fixture from the downloaded body (not **`HEAD`**), and add injected-**`fetch`** cases for missing **`Content-Length`** on **206** and for range-ignored **200** responses.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3d293532699d28f93067fc25b16b9c34e83e5bfb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/blob/src/utils/WebBlob.spec.ts
+++ b/packages/blob/src/utils/WebBlob.spec.ts
@@ -8,10 +8,14 @@ describe("WebBlob", () => {
 	let contentType: string;
 
 	beforeAll(async () => {
-		const response = await fetch(resourceUrl, { method: "HEAD" });
-		size = Number(response.headers.get("content-length"));
+		// Compute the reference size from the response body itself; in browsers
+		// `Content-Length` is not reliably exposed when the response is gzipped
+		// on the fly by CloudFront.
+		const response = await fetch(resourceUrl);
+		const blob = await response.blob();
+		size = blob.size;
+		fullText = await blob.text();
 		contentType = response.headers.get("content-type") || "";
-		fullText = await (await fetch(resourceUrl)).text();
 	});
 
 	it("should create a WebBlob with a slice on the entire resource", async () => {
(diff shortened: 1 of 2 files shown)


coyotte508:
note @huggingface/blob isn't really used, might as well archive it

codeAnqiang-ma:
Yeah, 66 downloads last month, hard to argue with that. One thing: `packages/dduf` depends on it (`workspace:^`), so that'd need somewhere to go first. Either way works for me — merge this as a stopgap, or close it if you're archiving.

coyotte508:
yea please post comments yourself not with your AI

PR #2387, opened 2026-08-18
🐛 jinja: `not` on empty arrays/objects should follow Python truthiness

ArunveerSidhu:
## Bug

`evaluateUnaryExpression` negates the raw JavaScript value:

```ts
case "not":
    return new BooleanValue(!argument.value);
```

For `ArrayValue`/`ObjectValue`, `.value` is a JS array/`Map`, which is always truthy in JavaScript — so `not []` and `not {}` evaluate to `false`, while Python Jinja2 evaluates them to `True`:

| template | `@huggingface/jinja` (main) | Python `jinja2` |
|---|---|---|
| `{{ not [] }}` | `false` | `True` |
| `{{ not {} }}` | `false` | `True` |
| `{% if not tools %}EMPTY{% else %}HAS{% endif %}` with `tools: []` | `HAS` | `EMPTY` |

In practice this means chat templates guarding with `{% if not tools %}` / `{% if not messages %}` silently take the wrong branch when the list is empty.

## Fix

Use the existing `__bool__()` helper — the same Python-truthiness path already used by `and`/`or` (short-circuiting), ternaries, `{% if %}` and `{% for ... if %}` — so `not x` is consistent with `if x`. One-line change in `runtime.ts`.

## Tests

- Added `LOGICAL_NOT_COLLECTIONS` to `templates.test.js` (lexing + interpretation + formatting), with expected outputs matching Python `jinja2` (verified against jinja2 3.x).
- Full suite passes, including the e2e suite that renders real Hub chat templates against reference outputs (`539 + 55` tests).

---

> [!NOTE]
> **Low Risk**
> Single-line runtime change aligned with existing `and`/`or`/`if` truthiness; behavior change only for `not` on empty arrays/objects.
> 
> **Overview**
> Fixes **`not`** so it uses **`__bool__()`** (Python-style truthiness) instead of negating the raw JavaScript `.value`, which made empty arrays and objects incorrectly truthy.
> 
> **`not []`**, **`not {}`**, and **`not tools`** when `tools` is `[]` now match Python Jinja2 (e.g. `true` / `false` / `true` for the new `LOGICAL_NOT_COLLECTIONS` fixture). Chat templates that use **`{% if not tools %}`** or similar guards on empty lists will take the correct branch.
> 
> Adds **`LOGICAL_NOT_COLLECTIONS`** coverage in `templates.test.js` (template, tokens, context, expected output).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 40672ddcce119d068ed5871e963cf7c970481c55. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


xenova:
did you intend to close the PR?

ArunveerSidhu:
> did you intend to close the PR?

yes i will reopen it soon, i think it still needs some work

PR #2406, opened 2026-08-24
[jinja] Fix namespaces and add support for ** syntax

xenova:
> [!NOTE]
> **Medium Risk**
> Changes core template evaluation (namespaces, macro/call binding, and call semantics), which can alter existing templates that relied on old namespace or macro behavior; coverage is broad but behavior is subtle.
> 
> **Overview**
> Brings the Jinja engine closer to Jinja2 for **namespaces**, **call/macro arguments**, **`**` unpacking**, and **`**` exponentiation**, and tightens **template formatting** around operator precedence.
> 
> **Namespaces** now use a dedicated mutable `NamespaceValue` (not plain objects): `namespace()` accepts optional positional mapping/iterable input plus keyword args, copies on construction, and only namespaces support `{% set ns.attr = ... %}`. Collection filters (`map`, `selectattr`, etc.) treat namespaces like attribute containers.
> 
> **Calls and macros** gain `**expr` keyword unpacking (must be last), stricter parse-time rules (positional before keyword, single `*`, no unpacking in parameter lists), and runtime binding aligned with Jinja: defaults, optional `kwargs`/`varargs` when referenced in the body (with AST-based detection), shared `bindMacroArguments` for macros and `{% call %}` blocks, and Python-like evaluation order for positional vs keyword args.
> 
> **Expressions**: lexer/parser/runtime add left-associative `**` with numeric edge-case errors; the formatter uses a full precedence table and smarter parentheses (including `**` in output).
> 
> Extensive template, format, parse, and regression tests cover the new behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5f628d7b41d7c003fc96c7ea42dd323d20099dd0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/jinja/src/runtime.ts
```diff
@@ -1749,32 +1855,51 @@ export class Interpreter {
...
+					if (value === undefined && defaultValue !== undefined) {
+						value = this.evaluate(defaultValue, macroScope);
+					}
+					macroScope.setVariable(name, value ?? new UndefinedValue());
+				}
+
+				// Python Jinja2 raises for arguments the macro cannot accept, unless the
+				// body catches them via the special variables.
+				if (catchKwargs) {
+					macroScope.setVariable("kwargs", new ObjectValue(kwargs));
```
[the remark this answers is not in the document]
xenova: incorrect. jinja2 doesn't raise multiple-values errors here.
```py
import jinja2
print(jinja2.__version__)  # 3.1.6

env = jinja2.Environment()

# Parameter given both positionally and as keyword, body catches kwargs:
print(env.from_string(
    "{% macro f(x) %}x={{ x }} kwargs={{ kwargs }}{% endmacro %}{{ f(1, x=2) }}"
).render())
```

produces
```
3.1.6
x=1 kwargs={'x': 2}
```

xenova (commented):
did a couple hours of iterations with this PR, so I think it's good to go now. Started it a month or so ago, but only got to finishing it now 😅

packages/jinja/src/format.ts
```diff
@@ -26,15 +26,20 @@ import type {
...
 
 const NEWLINE = "\n";
 const OPEN_STATEMENT = "{%- ";
 const CLOSE_STATEMENT = " -%}";
+const CONDITIONAL_EXPRESSION_PRECEDENCE = -1;
+const TEST_EXPRESSION_PRECEDENCE = 6;
 
 function getBinaryOperatorPrecedence(expr: BinaryExpression): number {
 	switch (expr.operator.type) {
+		case "ExponentiationBinaryOperator":
```
nico-martin, line 53: `UnaryExpression` needs to account for the parent precedence here. For example, `{{ (not true) ** 2 }}` currently formats to `{{- not true ** 2 -}}`; the original renders `0`, but the formatted template renders `false`. Please preserve the parentheses when a `not` expression is an exponentiation operand and add a format round-trip regression test.

packages/jinja/src/runtime.ts
```diff
@@ -864,6 +1104,23 @@ export class Interpreter {
...
+
+			const result = a ** b;
+			if (!Number.isFinite(result)) {
+				// NaN represents an unsupported complex result; Infinity represents overflow.
+				throw new Error("Exponentiation result is not a finite real number");
+			}
+
+			// Negative exponents produce floats.
+			const isFloat = left instanceof FloatValue || right instanceof FloatValue || b < 0;
+			return isFloat ? new FloatValue(result) : new IntegerValue(result);
```
nico-martin, line 1123: Integer exponentiation can silently return a rounded value because the result is stored as a JavaScript `number` and still wrapped as an `IntegerValue`. Jinja 3.1.6 renders `{{ 3 ** 40 }}` as `12157665459056928801`, while this implementation renders `12157665459056929000`. Please use exact integer arithmetic or reject unsafe integer results instead of returning an incorrect value, and add a boundary regression test.

packages/jinja/src/runtime.ts
```diff
@@ -967,13 +1230,19 @@ export class Interpreter {
 				for (const item of val.value) {
 					positionalArguments.push(item);
 				}
+			} else if (argument.type === "KeywordSpreadExpression") {
```
nico-martin, line 1243: This evaluates arguments in source order, but Jinja evaluates the dynamic positional argument before keyword arguments for the newly supported keyword-before-`*` syntax. With `f(b=mark("b"), *mark("k"))`, Jinja calls `mark("k")` and then `mark("b")`, while this implementation does the reverse. Please match Jinja's evaluation order and add a side-effect/order regression test.

nico-martin (changes requested):
Looks like a valuable compatibility improvement, especially for real chat templates that rely on namespaces and keyword unpacking.
Imo there are three issues in newly added behavior that should be fixed before merging.

xenova:
thanks @nico-martin for the review. I made fixes 1 and 3.

Fix 2 is mostly unrelated to this PR as it's more general about arithmetic (js is capped, py is effectively uncapped), and is probably not feasible to fix without introducing `bigint` support. It's also super rare to have any of that in a chat template.

nico-martin (approved):
lgtm :)!
I agree that the integer behavior does not need to block this chat-template-focused PR.

PR #2463, opened 2026-09-08
Fix library snippets that render broken or misleading code

gary149:
## Why

Rendering the library snippets for the most downloaded models of each library and comparing the output with the library's current release and with each model card turned up code that cannot run as pasted. This PR takes only the fixes where the current output is definitely wrong for the affected models and the replacement cannot be worse for any model: syntax errors, undefined names, unquoted repo ids, removed commands, wrong loader classes, and branches that were unreachable because of how tags are emitted.

Larger changes (per-task example calls for transformers, text-to-video and single-file handling for diffusers, mlx ASR branches, lerobot) are left for follow-ups since they need per-model fixtures.

## What

**transformers**
- `question-answering`, `visual-question-answering` and `image-to-image` were removed from `pipeline()` in v5 but were missing from the removed-pipelines warning list; also fixes the malformed `pip install "transformers<5.0.0'` hint.
- The pipeline chat example was gated on the `conversational` tag alone. A model tagged conversational without a chat template (e.g. one shipping a custom encoder) crashed on `pipe(messages)`. It now requires `hasChatTemplate`, like the AutoModel branch already did.
- `max_new_tokens=40` becomes 256, the pipeline default; 40 tokens is consumed entirely by the thinking preamble of current chat models.
- `# pip install -U transformers accelerate` next to `device_map="auto"`, which requires accelerate and is not a hard dependency.

**sentence-transformers**
- Rerankers saved in the CrossEncoder format but tagged `text-classification` (e.g. `BAAI/bge-reranker-v2-m3`) fell into the bi-encoder branch, which silently discards the classification head. They are detected from `config.architectures` (`*ForSequenceClassification`), gated on the `text-classification` tag so bi-encoders are unaffected.
- New `SparseEncoder` branch for repos tagged `sparse-encoder` (the tag the library sets on push).

**diffusers**
- The inpainting and FluxFill branches tested for bare class-name tags, but the Hub emits `diffusers:<ClassName>`, so they were unreachable. A small helper now also checks the prefixed tag and `config.diffusers._class_name`.
- LoRA image-to-video called `export_to_video(output, ...)` with `output` never assigned.
- The inpainting template no longer forces `variant="fp16"`: several inpainting repos (including ones now reaching this branch) only ship non-variant weights.
- `cardData.base_model` can be an array; it was joined with commas into an invalid repo id. The first entry is used.

**transformers.js**
- `sentence-similarity` and `text-ranking` are not `pipeline()` tasks in transformers.js and threw immediately; they map to `feature-extraction` and `text-classification`.

**peft**
- `QUESTION_ANS` and `FEATURE_EXTRACTION` adapters rendered "Task type is invalid."; they now load (FEATURE_EXTRACTION through `AutoModel`, as PEFT's own auto class does).
- Whisper-style `SEQ_2_SEQ_LM` adapters used `AutoModelForSeq2SeqLM`, which raises for those architectures; they use `AutoModelForSpeechSeq2Seq`. The override is keyed on the base model's Hub repo id only (Whisper, SpeechT5, Speech2Text as `s2t-*`, SeamlessM4T), since Qwen2-Audio and GLM-ASR are registered under the text class.
- When `adapter_config.json` has no `base_model_name_or_path`, the model card's `base_model` is used instead of "Base model is not found."

**keras / tf-keras**
- `keras.saving.load_model("hf://...")` requires TensorFlow even on other backends, and the snippet had no install line; it now installs it and selects the tensorflow backend.
- `from_pretrained_keras` was removed in huggingface_hub 1.0; the tf-keras snippet pins `huggingface_hub<1.0`.

**Long tail, deterministic fixes**
- Unquoted repo ids: sam2, mlx-image, index-tts, describe-anything, sam-3d-body, diffusionkit.
- Syntax and imports: cartesia_pytorch (stray parenthesis), vui (trailing comma in import), lvface / univa / outetts (indentation), phantom_wan (`WANI2V` vs the exported `WanI2V`, undefined prompts, the image is opened with PIL as `generate` expects), depth-anything-v2 (missing bracket and `hf_hub_download` import), kimi-audio and espnet (`soundfile`), chat_tts (`torch` import, and `Chat.load()` since `load_models` does not exist), lvface (drops a `timeout` kwarg the constructor does not accept).
- ml-agents `--local-dir` literal, timm `hf-hub:` prefix, kernels `version=` (required by `get_kernel`), gliner2 variable name, pruna_pro substitution no longer rewrites the repo id, sklearn uses the filename declared in `config.json`, speechbrain `SpeakerRecognition`, ultralytics uses `YOLO(path)` for mainline and keeps `YOLOv10.from_pretrained` for the fork, litert-lm passes the required `.litertlm` filename, colipri passes the model object as its card does, `huggingface-cli download` replaced by `hf download`, adapters placeholder made unmistakable.
- One inference call added to load-only snippets: gliner, span-marker, setfit, model2vec, flair.

## Tests

- New spec cases in `model-libraries-snippets.spec.ts` for each behavioral change (chat gate, v5 warning, reranker and sparse routing, prefixed diffusers tags, base_model arrays, LoRA video, transformers.js aliases, PEFT task types and speech adapters, ultralytics, pruna, sklearn, speechbrain, quoting, static templates).
- `pnpm test`, `pnpm check`, `pnpm lint` and `pnpm format` pass in `packages/tasks`.
- Snippets were rendered against live Hub metadata for the top models of each touched library, and every changed Python template parses.
- Every import, class, kwarg and CLI flag used by the changed snippets was checked against the upstream library source at the pinned versions (transformers 5.16.1, sentence-transformers 6.0.1, diffusers 0.40.0, peft 0.20.0, keras 3.15.1, huggingface_hub 1.30.0, @huggingface/transformers 4.2.0, kernels 0.16.1, speechbrain 1.1.1, and the current releases of the long-tail libraries).

### DIFF
--- a/packages/tasks/src/model-libraries-snippets.ts
+++ b/packages/tasks/src/model-libraries-snippets.ts
@@ -22,7 +22,7 @@ export const adapters = (model: ModelData): string[] => [
 	`from adapters import AutoAdapterModel
 
 model = AutoAdapterModel.from_pretrained("${escapeStringForJson(
-		model.config?.adapter_transformers?.model_name ?? "fill-in-model-name",
+		model.config?.adapter_transformers?.model_name ?? "<base-model-id>",
 	)}")
 model.load_adapter("${model.id}", set_active=True)`,
 ];
@@ -83,7 +83,10 @@ result, message = detector.detect_watermark(watermarked_audio, sr)`;
 };
 
 function get_base_diffusers_model(model: ModelData): string {
-	return escapeStringForJson(model.cardData?.base_model?.toString() ?? "fill-in-base-model");
+	const baseModel = model.cardData?.base_model;
+	// several base models can be listed (e.g. a LoRA trained for a base and its turbo variant); the first one is used
+	const first = Array.isArray(baseModel) ? baseModel[0] : baseModel;
+	return escapeStringForJson(first?.toString() ?? "fill-in-base-model");
 }
 
 function get_prompt_from_diffusers_model(model: ModelData): string | undefined {
@@ -174,7 +177,7 @@ detector = cvg.NeuralCornerDetector(checkpoint)
 embedder = cvg.NeuralEmbedder(checkpoint)`,
 ];
 
-export const colipri = (model: ModelData): string[] => {
+export const colipri = (): string[] => {
 	const installSnippet = `pip install colipri`;
 
 	const exampleSnippet = `from colipri import get_model
@@ -184,7 +187,7 @@ from colipri import ZeroShotImageClassificationPipeline
 
 model = get_model().cuda()
 processor = get_processor()
-pipeline = ZeroShotImageClassificationPipeline("${model.id}", processor)
+pipeline = ZeroShotImageClassificationPipeline(model, processor)
 
 image = load_sample_ct()
 
@@ -308,7 +311,7 @@ export const depth_anything_v2 = (model: ModelData): string[] => {
 	} else if (model.id === "depth-anything/Depth-Anything-V2-Large") {
 		encoder = "vitl";
 		features = "256";
-		out_channels = "[256, 512, 1024, 1024";
+		out_channels = "[256, 512, 1024, 1024]";
 	}
 
 	return [
@@ -318,6 +321,7 @@ export const depth_anything_v2 = (model: ModelData): string[] => {
 # Load the model and infer depth from an image
 import cv2
 import torch
+from huggingface_hub import hf_hub_download
 
 from depth_anything_v2.dpt import DepthAnythingV2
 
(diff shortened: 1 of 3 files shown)


packages/tasks/src/model-libraries-snippets.ts
```diff
@@ -2149,6 +2188,28 @@ print(scores)`,
...
+]
+
+query_embeddings = model.encode_query(queries)
+document_embeddings = model.encode_document(documents)
+
+similarities = model.similarity(query_embeddings, document_embeddings)
+print(similarities)`,
+		];
+	}
+
```
tomaarsen, line 2212: As of https://huggingface.co/blog/multi-vector-encoder from a few weeks back, there's also `MultiVectorEncoder` with a near identical usage format as that `SparseEncoder`. It's identified by the `multi-vector` tag alongside the `sentence-transformers` library name.

tomaarsen (commented):
Looks good at quick glance. It can be extended a bit further as well, would be quite valuable. Thanks for this!

PR #2465, opened 2026-09-08
fix(jinja): lower-case the rest of the string in capitalize and title

pucedoteth:
## What

`capitalize` and `title` only upper-cased characters; they never lower-cased the rest. Python does both, so rendering diverged from Jinja on any input that already contained upper-case letters.

```
                     python jinja2 3.1.6      @huggingface/jinja (before)
'hello WORLD' | capitalize   'Hello world'    'Hello WORLD'
'hello WORLD' | title        'Hello World'    'Hello WORLD'
'ABC def'     | capitalize   'Abc def'        'ABC def'
'HELLO'       | title        'Hello'          'HELLO'
```

While fixing `title` I also had to change how it finds word starts. It used `/\b\w/`, but Jinja splits on `([-\s({\[<]+)` — `\b` additionally breaks on `'` and `.`:

```
"don't stop" | title    python "Don't Stop"    before "Don'T Stop"
'x.y z'      | title    python 'X.y Z'         before 'X.Y Z'
```

Both filters resolve through the string builtins, so `s.capitalize()` and `s.title()` were wrong in the same way and are fixed by the same change.

## Why it went unnoticed

`FILTER_OPERATOR_2` is the only existing coverage, and it feeds both filters `'test test'` — already lower-case, so it renders identically whether or not the remainder is lowered. The expected value `|3|ABCD|abcd|Test test|Test Test|a b|4|` is correct under both the old and new code.

## Verification

Differentially tested against **python jinja2 3.1.6** over 22 inputs × 2 filters (mixed case, leading/trailing whitespace, `-`/`(`/`{`/`[`/`<` separators, apostrophes, periods, digits, non-ASCII, empty string):

```
before   31 / 44 mismatches
after     0 / 44
```

Red/green with only the two source files reverted and the new test kept:

```
without the fix   9 failed | 629 passed
with the fix      638 passed
```

Full package suite: **686 passed**. `tsc --noEmit`, `oxfmt --check` and `eslint` all clean.

One note on what I could not run: `test/e2e.test.js` fails to load `@huggingface/hub` in my environment because I installed only this package rather than the whole workspace. It fails identically on a clean checkout of `main`, so it is environmental and unrelated to this change — CI will cover it.

## AI disclosure

Written with Claude Opus 5 via Claude Code. The divergence was found by reading `runtime.ts`/`utils.ts` against Python's `do_title`/`str.capitalize`, then confirmed by running both implementations side by side; every expected value in the new test is the literal output of python jinja2 3.1.6, not hand-written.

---

> [!NOTE]
> **Low Risk**
> Localized Jinja string filter semantics with broad regression tests; no auth, data, or infrastructure impact.
> 
> **Overview**
> **`capitalize` and `title`** (filters and `s.capitalize()` / `s.title()` builtins) now lower-case the rest of each affected segment after upper-casing the first character, matching Python Jinja2 instead of leaving existing capitals unchanged.
> 
> **`title`** no longer uses word-boundary regex; it splits on Jinja’s separator set (`-`, whitespace including Python-specific separators, `(`, `{`, `[`, `<`), so strings like `don't stop` and `x.y z` title-case like upstream rather than breaking on `'` or `.`.
> 
> Shared helpers in `utils.ts` (`titleCase`, new `capitalizeFirst`, code-point–aware casing) replace the old inline logic. A **String case filters** test block asserts behavior against Python jinja2 3.1.6, including Unicode edge cases.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cd7916fed0bbd6c02fa0d376e49c53b25f9a7b75. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/jinja/src/utils.ts
+++ b/packages/jinja/src/utils.ts
@@ -59,14 +59,40 @@ export function slice<T>(array: T[], start?: number, stop?: number, step = 1): T
  * @param value The string to title case.
  * @returns The title cased string.
  */
+// Jinja splits words on Python's `([-\s({\[<]+)`. The class is spelled out because JS and
+// Python disagree on `\s`: JS misses U+001C-U+001F and U+0085, and adds U+FEFF.
+// eslint-disable-next-line no-control-regex -- Python's `\s` matches these separators, so Jinja splits on them.
+const WORD_SEPARATOR = /([-\t\n\v\f\r\x1c-\x1f \x85\xa0\u1680\u2000-\u200a\u2028\u2029\u202f\u205f\u3000({[<]+)/;
+
+/** Upper-case the first code point and lower-case the rest. Indexing by code point keeps an
+ *  astral letter such as Deseret "\u{10428}" intact instead of splitting its surrogate pair. */
+function upperFirstLowerRest(value: string): string {
+	const first = value.codePointAt(0);
+	if (first === undefined) {
+		return value;
+	}
+	const head = String.fromCodePoint(first);
+	return head.toUpperCase() + value.slice(head.length).toLowerCase();
+}
+
+/** Jinja's `title` filter: upper-case each word's first character, lower-case the rest. */
 export function titleCase(value: string): string {
-	// Mirrors Python's `str.title()` as Jinja implements it: split on runs of "-", whitespace,
-	// "(", "{", "[" or "<" (keeping the separators), then upper-case the first character of each
-	// piece and lower-case the rest. `\b` is not equivalent — it also breaks on "'" and ".".
-	return value
-		.split(/([-\s({[<]+)/)
-		.map((piece) => piece.charAt(0).toUpperCase() + piece.slice(1).toLowerCase())
-		.join("");
+	return value.split(WORD_SEPARATOR).map(upperFirstLowerRest).join("");
+}
+
+/** Python's `str.capitalize()`: title-case the first character, lower-case the rest. Title-case
+ *  is not upper-case where a character expands — "\u00df" title-cases to "Ss" but upper-cases to
+ *  "SS", which is why {@link titleCase} cannot be reused here. JS has no title-case mapping, so
+ *  the ~45 code points whose title-case is a distinct character (the "\u01f3" digraphs, and the
+ *  Greek iota-subscript block) still come out upper-cased. */
+export function capitalizeFirst(value: string): string {
+	const first = value.codePointAt(0);
+	if (first === undefined) {
+		return value;
+	}
+	const head = String.fromCodePoint(first);
+	const upper = head.toUpperCase();
+	return (upper.length > 1 ? upper[0] + upper.slice(1).toLowerCase() : upper) + value.slice(head.length).toLowerCase();
 }
 
 export function strftime_now(format: string): string {
--- a/packages/jinja/src/utils.ts
+++ b/packages/jinja/src/utils.ts
@@ -59,8 +59,13 @@ export function slice<T>(array: T[], start?: number, stop?: number, step = 1): T
  * @param value The string to title case.
  * @returns The title cased string.
  */
-// Jinja splits words on Python's `([-\s({\[<]+)`. The class is spelled out because JS and
-// Python disagree on `\s`: JS misses U+001C-U+001F and U+0085, and adds U+FEFF.
+// Jinja splits words on `([-\s({\[<]+)`:
+// https://github.com/pallets/jinja/blob/3.1.6/src/jinja2/filters.py#L328
+// `\s` is spelled out because the two languages define it differently. Python's `\s` matches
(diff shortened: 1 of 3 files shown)


xenova:
Thanks for the PR! Surprised we didn't catch this earlier.

pucedoteth:
Thanks @xenova! And Copilot was right on both points — both are fixed in `f7ee7a4`, along with a third I found while checking them.

**1. Code points.** `charAt(0)` took a UTF-16 code unit, so Deseret `𐐨` was left uncased. Now indexed with `codePointAt`.

**2. Docs naming.** Also correct, and it was my error: Jinja's `title` is *not* `str.title()`. They disagree exactly where this PR changed the split:

| input | `str.title()` | jinja `\|title` |
|---|---|---|
| `don't stop` | `Don'T Stop` | `Don't Stop` |
| `x.y z` | `X.Y Z` | `X.y Z` |

The docstrings now name Jinja's filter.

**3. Whitespace (found while verifying 1 and 2).** The separator class genuinely differs in both directions, so `\s` cannot be used. Enumerating the whole BMP against Python's `([-\s({\[<]+)`:

```
python splits, JS does not:  U+001C U+001D U+001E U+001F U+0085
JS splits, python does not:  U+FEFF
```

It is now spelled out explicitly.

**One more thing worth flagging:** `capitalize` *title-cases* its first character while `title` *upper-cases* — `ß` title-cases to `Ss` but upper-cases to `SS` — so they can no longer share a helper. JS has no title-case mapping, so ~45 code points whose title-case is a distinct character (the `ǳ` digraphs and the Greek iota-subscript block) still upper-case under `capitalize`. That is 1 residual mismatch out of 72 in my differential; I left it documented in the docstring rather than embedding a Unicode table in a template engine, but happy to add one if you'd prefer exactness there.

**Verification** — differential against python jinja2 3.1.6 over 36 inputs × 2 filters, now including astral letters, `ß`/`ﬁ`/`ı`, and each of the separators above:

```
origin/main   45 / 72 mismatches
this branch    1 / 72   (the documented digraph case)
```

Red/green with both source files reverted to `origin/main` and the tests kept: **14 failed → 645 passed**. `eslint`, `tsc` and `oxfmt` clean.

packages/jinja/src/utils.ts
```diff
@@ -59,8 +59,40 @@ export function slice<T>(array: T[], start?: number, stop?: number, step = 1): T
  * @param value The string to title case.
  * @returns The title cased string.
  */
+// Jinja splits words on Python's `([-\s({\[<]+)`. The class is spelled out because JS and
+// Python disagree on `\s`: JS misses U+001C-U+001F and U+0085, and adds U+FEFF.
+// eslint-disable-next-line no-control-regex -- Python's `\s` matches these separators, so Jinja splits on them.
+const WORD_SEPARATOR = /([-\t\n\v\f\r\x1c-\x1f \x85\xa0\u1680\u2000-\u200a\u2028\u2029\u202f\u205f\u3000({[<]+)/;
```
xenova: do you have a reference for this? It would be nice to link to something.

pucedoteth: Added in `cd7916f` — three links, one per claim:

- **the separator set itself** — Jinja's [`_word_beginning_split_re`](https://github.com/pallets/jinja/blob/3.1.6/src/jinja2/filters.py#L328), `([-\s({\[<]+)`, used by `do_title` on the next line.
- **why Python's `\s` covers U+001C-U+001F and U+0085** — [`str.isspace()`](https://docs.python.org/3/library/stdtypes.html#str.isspace): "Other" or "Separator" category, or bidirectional class WS, B or S. Those five are `Cc` but bidi `B`/`S`, so they qualify. I checked that Python's regex `\s` and `str.isspace()` agree on every code point in Unicode — same 29, no exceptions — so the documented `str.isspace()` rule is a sound reference for the regex.
- **why JS differs** — [ECMAScript `\s`](https://tc39.es/ecma262/#sec-white-space) is WhiteSpace + LineTerminator, which omits those four control characters and explicitly includes `<ZWNBSP>` (U+FEFF).

That last one is the whole divergence in one line: U+FEFF is `Cf` with bidi `BN`, so it is whitespace to ECMAScript and not to Python.

```
U+001C..U+001F  Cc  bidi B/S   python \s ✓   JS \s ✗
U+0085          Cc  bidi B     python \s ✓   JS \s ✗
U+FEFF          Cf  bidi BN    python \s ✗   JS \s ✓
```

pucedoteth:
@xenova all the review points are addressed: code-point indexing via `codePointAt` for the Deseret case, docstrings naming Jinja's `title` rather than `str.title()`, and the separator-class difference I turned up while checking those. Current head is `cd7916f` and Bugbot is green. Happy to split the whitespace change into its own PR if that would make it easier to review.

PR #2466, opened 2026-09-09
Add Core AI to model libraries

john-rocky:
Registers `coreai`, Apple's on-device runtime for iOS 27 / macOS 27, so that repos hosting `.aimodel` bundles get the library label, a Swift "Use this model" snippet, and download counting.

198 repos from 24 authors already carry `library_name: coreai` (https://huggingface.co/models?other=coreai). Without an entry they show as an unregistered tag, and most count no downloads, because a bundle folder usually carries no `config.json`.

Following the registration docs:

- Entry between `comet` and `cortiq`, `filter: false`. `repoUrl` points at Apple's runtime package ([apple/coreai-models](https://github.com/apple/coreai-models)), `docsUrl` at Apple's Core AI documentation.
- `countDownloads: path_extension:"mlirb"`: every `.aimodel` bundle contains one `main.mlirb`, so a bundle download counts once, the same shape as the `litert` and `litert-lm` rules.
- Snippets: text-generation shows Apple's `CoreAILanguageModels` loader first, then the community CoreAIKit package, which loads straight from the Hub repo. Other tasks get the generic `AIModel(contentsOf:)` loader and a pointer to the card for inputs and outputs.

Rendered for `mlboydaisuke/MiniCPM5-2B-CoreAI` (text-generation):

```swift
// Apple Core AI (iOS 27 / macOS 27): this repo holds an .aimodel bundle that runs fully on-device.
// 1. Download the bundle folder:  hf download mlboydaisuke/MiniCPM5-2B-CoreAI --local-dir ./MiniCPM5-2B-CoreAI
// 2. Load it with Apple's Swift package: https://github.com/apple/coreai-models
import FoundationModels
import CoreAILanguageModels

let model = try await CoreAILanguageModel(resourcesAt: bundleFolderURL)   // folder with the .aimodel + tokenizer/
let session = LanguageModelSession(model: model)
print(try await session.respond(to: "Explain on-device AI in one sentence."))
```

```swift
// Or with the community CoreAIKit package, which downloads from this repo on first use:
// https://github.com/john-rocky/coreai-kit
import CoreAIKit

let chat = try await ChatSession(model: ModelID("mlboydaisuke/MiniCPM5-2B-CoreAI"))   // add path: "<subfolder>" if the bundle sits in one
print(try await chat.respond(to: "Explain on-device AI in one sentence."))
```

For transparency: I maintain the community [Core AI model zoo](https://github.com/john-rocky/coreai-model-zoo) and CoreAIKit, where 78 of those repos come from; the other authors' repos use the same `.aimodel` layout. I am not affiliated with Apple. The key name, the snippet order, and the counting rule are all easy to change to whatever fits the Hub's conventions.

---

> [!NOTE]
> **Low Risk**
> Adds registry metadata and display snippets only; no runtime, auth, or data-handling logic changes.
> 
> **Overview**
> Registers **Apple Core AI** (`library_name: coreai`) in the Hub model-libraries registry so `.aimodel` repos get a proper **Core AI** label, docs/repo links, and download stats.
> 
> Adds a **`coreai` snippet generator** with Swift “use this model” examples: `hf download` plus Apple’s `CoreAILanguageModels` path for **text-generation**, an optional **CoreAIKit** Hub-based alternative, and a generic **`AIModel(contentsOf:)`** snippet for other tasks. Download counting uses **`path_extension:"mlirb"`** (one count per bundle via `main.mlirb`), with **`filter: false`** like other niche runtimes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d48cd868a6551149541004c7bb49c61fb2d78773. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/tasks/src/model-libraries.ts
+++ b/packages/tasks/src/model-libraries.ts
@@ -342,6 +342,15 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
 		repoUrl: "https://github.com/Unbabel/COMET/",
 		countDownloads: `path:"hparams.yaml"`,
 	},
+	coreai: {
+		prettyLabel: "Core AI",
+		repoName: "coreai-models",
+		repoUrl: "https://github.com/apple/coreai-models",
+		docsUrl: "https://developer.apple.com/documentation/coreai",
+		snippets: snippets.coreai,
+		filter: false,
+		countDownloads: `path_extension:"mlirb"`,
+	},
 	cortiq: {
 		prettyLabel: "cortiq",
 		repoName: "cortiq",
--- a/packages/tasks/src/model-libraries-snippets.ts
+++ b/packages/tasks/src/model-libraries-snippets.ts
@@ -247,6 +247,37 @@ print("R² Score:", r2)`;
 	return [installSnippet, classificationSnippet, regressionsSnippet];
 };
 
+export const coreai = (model: ModelData): string[] => {
+	const folder = model.id.split("/").pop() ?? "model";
+	const header = `// Apple Core AI (iOS 27 / macOS 27): this repo holds an .aimodel bundle that runs fully on-device.
+// 1. Download the bundle folder:  hf download ${model.id} --local-dir ./${folder}
+// 2. Load it with Apple's Swift package: https://github.com/apple/coreai-models`;
+	if (model.pipeline_tag === "text-generation") {
+		return [
+			`${header}
+import FoundationModels
+import CoreAILanguageModels
+
+let model = try await CoreAILanguageModel(resourcesAt: bundleFolderURL)   // folder with the .aimodel + tokenizer/
+let session = LanguageModelSession(model: model)
+print(try await session.respond(to: "Explain on-device AI in one sentence."))`,
+			`// Or with the community CoreAIKit package, which downloads from this repo on first use:
+// https://github.com/john-rocky/coreai-kit
+import CoreAIKit
+
+let chat = try await ChatSession(model: ModelID("${model.id}"))   // add path: "<subfolder>" if the bundle sits in one
+print(try await chat.respond(to: "Explain on-device AI in one sentence."))`,
+		];
+	}
+	return [
+		`${header}
+import CoreAI
+
+let model = try await AIModel(contentsOf: aimodelURL)   // the .aimodel inside the downloaded folder
+// Inputs, outputs and any companion files are described on this model's card.`,
+	];
+};
+
 export const cortiq = (model: ModelData): string[] => {
 	const setup = `# one Rust binary, no additional dependencies
 cargo install cortiq-cli   # or a prebuilt binary from github.com/infosave2007/cmf/releases


packages/tasks/src/model-libraries.ts
```diff
@@ -342,6 +342,15 @@ export const MODEL_LIBRARIES_UI_ELEMENTS = {
...
 		countDownloads: `path:"hparams.yaml"`,
 	},
+	coreai: {
+		prettyLabel: "Core AI",
+		repoName: "coreai-models",
+		repoUrl: "https://github.com/apple/coreai-models",
+		docsUrl: "https://developer.apple.com/documentation/coreai",
+		snippets: snippets.coreai,
+		filter: false,
+		countDownloads: `path_extension:"mlirb"`,
```
Wauplin, line 352: as far as I can see, all coreai models (https://huggingface.co/models?other=coreai) have a config.json, which is the default file we are counting downloads on. What do you think of counting downloads on this one instead? (otherwise file with mlirb ones)

Wauplin (commented):
Thanks for the PR @john-rocky  :)

I have one comment about download count. Otherwise I'll defer to @pcuenca for the snippet part as I feel you'll have more context about it.

pcuenca (commented):
Hi @john-rocky, thanks a lot for your PR, for your community repo and for converting and uploading many Core AI models! 🙌 

I'm not convinced about the download snippets. We want them to be super easy, concise and reproducible. If I'm understanding your proposal correctly, it has a few limitations. First, the main use case is for `text-generation` models, where examples for both Apple's package and yours are shown. If the model is not tagged as `text-generation`, then it falls back to loading via `AIModel`, which might not be super useful. Even in the text generation case, model selection is not straightforward. Consider the example you used in your PR description, model https://huggingface.co/mlboydaisuke/MiniCPM5-2B-CoreAI. I think that loading the repo using `let chat = try await ChatSession(model: ModelID("mlboydaisuke/MiniCPM5-2B-CoreAI"))` would fail, it lacks an specific path to use that depends on whether the application is running on macOS or iOS, and whether a suitable version exists for the runtime platform. For this particular model, [there are three subdirectories](https://huggingface.co/mlboydaisuke/MiniCPM5-2B-CoreAI/tree/main): `int8`, `ios-ane-h18p`, and `ios-static`. According to your [catalog definition](https://github.com/john-rocky/coreai-kit/blob/f3c33b12b982fb17111d1baeedb0088d5c9e2c43/catalog.json#L194), `int8` is eligible for macOS and iOS, but I think your library doesn't check the catalog for an arbitrary HF URL, so it won't download from that directory. The `ios-ane-h18p` appears to contain a pre-compiled AOT version that would only be compatible with the iPhone 17 Pro, whereas the `ios-static` should be AOT-compiled before use on a particular arbitrary device, which might be super-slow on an actual iPhone.

These complexities would increase friction and create confusion, in my opinion.

I agree that we should have download rules for these files. But perhaps we should do it internally in the Hub, just as we deal with [Core ML files](https://huggingface.co/models?library=coreml&sort=trending). This would allow automatic recognition, and coexistence with other libraries, such as transformers or anything else. See [this repo](https://huggingface.co/google-bert/bert-base-uncased), for instance: it contains transformers weights and also a Core ML implementation which is automatically recognized.

I will open a PR in our internal codebase for discussion.

From my point of view, adoption should start with download counts, either configured automatically (my internal PR), or via a `coreai` library definition as an alternative. Later on, we can continue here with a library definition for your package, that would add sensible code snippets. Ideally, your library should automatically auto-resolve what version to download from the Hub, so the promise of `ChatSession(model: ModelID("mlboydaisuke/MiniCPM5-2B-CoreAI"))` is fulfilled no matter where the code is running. Perhaps you could upload a platform-configuration file to the Hub (instead of adding to the catalog), so your library checks what versions exist and then selects the appropriate one or fails gracefully.

Please, give us a few days to consider the Core AI native "format" and we can continue the discussion.

john-rocky:
Thanks for the careful read, @pcuenca, and sorry for the time it took: I should have run the snippet against the repo the PR cites before sending it. Thank you as well for opening the internal discussion on the native format. Agreed on the order: download rules first, code snippets later through a library definition for the package, once it auto-resolves the variant from the Hub. I'll hold this PR as it is while you look at the format. When you have a preference, it can shrink to the label plus the download rule (a 32-line deletion, ready on my fork: https://github.com/john-rocky/huggingface.js/commit/a25b464a8f2f1305dfb4cefd6176231ab5ed91c1) or close in favour of the internal detection, whichever is easier on your side.

**The library today.** I re-read the resolver at 0.4.2, and it matches your reading. `ModelID("mlboydaisuke/MiniCPM5-2B-CoreAI")` with no `path` looks for a `macos/` or `ios/` folder, the layout of the `-official` starter repos. It never consults the catalog. On that repo it stops at the tree listing, and the error says the model may not be published for this platform, which is wrong there. Only the catalog path, `ChatSession(catalog: "minicpm5-2b")`, resolves a variant, and at 0.4.2 that is `int8` on both platforms. The released library never touches `ios-ane-h18p` or `ios-static`. The compiled bundle loads on one architecture id, `h18p`, which follows the device identifier major `iPhone18` (validated on the iPhone 17 Pro). The static bundle is the same export before compilation, kept so anyone can compile it for another chip with `xcrun coreai-build compile`. The snippet in this PR skipped all of that. That was my gap.

**What I'll do.** The fix is the platform-configuration file you describe. A `coreai-kit.json` at the repo root carries what the catalog holds for a model today: the kind, and the variants with their paths, keyed `macos` / `ios`, plus `ios-ane-<arch>` for a compiled bundle. The library reads it before downloading. A repo without the file falls back to the tree: one bundle resolves, several fail gracefully with the folder names and a request for `path:`. Then `ChatSession(model: ModelID("<repo>"))` works on any repo or fails with a reason. I'll ship that in the library, add the file to the 104 repos I maintain, and come back here for the library definition once it is out.

**For the format detection**, the signature across the 200 tagged repos, from their file lists:

| bundle | directory | graph file | files | repos |
|---|---|---|---:|---:|
| source | `*.aimodel/` | `main.mlirb` | 583 | 192 |
| compiled | `*.aimodelc/` | `main-<arch>.mlirb` | 76 | 34 |

| layout | repos |
|---|---:|
| one bundle at the root | 60 |
| one bundle in a subfolder | 37 |
| several bundles (platform variants, or decoder + vision pairs) | 100 |
| `main.mlirb` in a plain folder, no `.aimodel` directory | 3 |

A per-file rule counts one download per bundle fetched, the same shape as `path_extension:"tflite"`. The `config.json` numbers are in Wauplin's thread. If a suggestion is welcome: recognise a repo on any `*.aimodel/` or `*.aimodelc/` directory, and count on `main*.mlirb`, one file per bundle. If a marker file or a different layout would make the detection cleaner, tell me once and I'll apply it across my repos. And once the format is recognised, if a docs page would help, I can draft one in the shape of the GGUF page.

Thanks again for taking the format seriously. Take whatever time it needs.

packages/tasks/src/model-libraries.ts
john-rocky: Thanks for the pointer. I checked the 200 tagged repos by file list. A root `config.json` exists in 88. In 75 of those it is a marker file (`model_type: coreai-aimodel`, the zoo's layout) that no loader reads, and in 13 it is the source model's transformers config kept beside the bundle. 47 repos have one only inside a subfolder, and 65 have none. Of the 88, 82 keep their bundles in subfolders, so a download of one variant subtree, which is what the library does, never requests the root file. That was the reason for `mlirb`: one count per bundle fetched. Either rule works for me, and if the format detection @pcuenca mentioned lands, the choice moves to your side anyway.

PR #2495, opened 2026-09-14
[hub] Force HTTP/1.1 for downloads on Node >= 26

assafvayner:
Node 26 bundles undici 8, whose `fetch` negotiates HTTP/2 by default and runs every HTTP/2 session with fixed 256 KiB stream / 512 KiB connection flow-control windows. Each download stream is capped near `window / (2 × RTT)`, about 3.5 MB/s at 75 ms RTT to `us.aws.cdn.hf.co`, so xet downloads run 3 to 10× slower on Node 26 than on Node 24 (13.5 vs 38.7 MB/s in a 6 × 256 MiB benchmark; 3.6 vs 33 MB/s on a single 128 MiB `downloadFile` slice). This follows the same approach as the Hub server (moon-landing `server/lib/Fetch.ts`, `new Agent({ allowH2: false })`) and xet-core (`reqwest::Client::builder().http1_only()`).

- On Node whose built-in undici is >= 8, `XetBlob` defaults to a fetch that disables HTTP/2; a caller-supplied `fetch` bypasses it. Regular (non-xet) downloads keep the runtime's default fetch. Bun and browsers are unaffected.
- No new dependency and `engines` unchanged: the request gets an `init.dispatcher` that forwards to Node's own global dispatcher (`Symbol.for("undici.globalDispatcher.2")`, `.1` as fallback) with the per-request `allowH2: false` that undici's `Agent` supports, so a user-installed global dispatcher (proxy, custom TLS) keeps working. If the global dispatcher cannot be found it falls back to the plain fetch.
- Verified on hub's own download path: Node 26 27.7 MB/s (was 3.6), Node 24 and Bun 1.4.2 unchanged. Raising the H2 windows to 16 MiB also restores full speed, which confirms flow control as the cause.
- The two hooks used (the global dispatcher symbol and per-request `allowH2`) are undocumented undici internals; both are present in undici 8.2.0 (Node 26.1) and 8.10.2 (Node 26.x, main), and the code is guarded to only run when the built-in undici is >= 8.

The real fix belongs upstream in undici (larger or adaptive default windows; Chromium uses 6 MiB / 15 MiB); this keeps hub fast in the meantime.

---

> [!NOTE]
> **Medium Risk**
> Changes the default download transport on Node 26+ using undocumented undici dispatcher symbols and `allowH2`, with fallback to plain fetch if discovery fails.
> 
> **Overview**
> **Fixes severely throttled xet downloads on Node 26+** by avoiding undici 8’s default HTTP/2 flow-control limits. When the caller does not supply `fetch`, `XetBlob` now defaults to an `http1Fetch` wrapper instead of bound `globalThis.fetch`.
> 
> On Node with built-in **undici ≥ 8**, that wrapper resolves Node’s global undici dispatcher (via `Symbol.for("undici.globalDispatcher.*")`), wraps it with per-request **`allowH2: false`**, and passes it as `init.dispatcher` so downloads use HTTP/1.1 while proxies and custom global dispatchers still work. Bun, browsers, and older Node keep ordinary `fetch`. Custom `params.fetch` is unchanged.
> 
> Adds **`hasH2Fetch`**, **`withHttp1`**, and unit tests for runtime detection and dispatcher forwarding.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 57af5bbf1186439ec32a67039a7523d9130093ed. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


assafvayner:
@coyotte508 feel free to re-write and/or re-create anything.

packages/hub/src/utils/XetBlob.ts
```diff
@@ -312,7 +313,7 @@ export class XetBlob extends Blob {
 	constructor(params: XetBlobCreateOptions) {
 		super([]);
 
-		this.fetch = params.fetch ?? fetch.bind(globalThis);
+		this.fetch = params.fetch ?? defaultFetch;
```
coyotte508: I think this should be only here? That we override http2 to http1.

Not in `WebBlob`, ... as it's something xet specific? Or are regular downloads impacted too?

assafvayner: as far as I understood, using a node version that uses undici >= 8 will auto-resolve http 2 rather than http 1. So this would affect regular downloads too (unless you want to keep them http 2?)

I can revert that behavior, just let me know what's the desired protocol for regular downloads.

coyotte508: > So this would affect regular downloads too (unless you want to keep them http 2?)

IMO if the main problem with node 26 is http2 for xet protocol downloads, let's just change the default there (and it's possible in a future iteration they fix it in node) not for all downloads

assafvayner: Fixed this and also folded into XetBlob files. take a look again when you get the chance

packages/hub/src/utils/XetBlob.ts
```diff
@@ -9,6 +9,82 @@ import { sum } from "./sum";
...
+	const dispatcher = await getHttp1Dispatcher();
+	return dispatcher ? fetch(input, { ...init, dispatcher } as RequestInit) : fetch(input, init);
+};
+
+/**
+ * The fetch used for xet downloads when the caller does not pass one. Only Node >= 26 actually forces
+ * HTTP/1.1; on every other runtime there is nothing to force and this is the global fetch, looked up at
+ * call time so test doubles installed on `globalThis` still apply.
+ */
+const http1Fetch: typeof fetch = hasH2Fetch() ? undiciHttp1Fetch : (input, init) => fetch(input, init);
```
coyotte508, line 86:
```suggestion
const http1Fetch: typeof fetch = hasH2Fetch() ? undiciHttp1Fetch : fetch;
```

?

coyotte508 (approved):

packages/hub/src/utils/XetBlob.ts
```diff
@@ -9,6 +9,82 @@ import { sum } from "./sum";
...
+	const dispatcher = await getHttp1Dispatcher();
+	return dispatcher ? fetch(input, { ...init, dispatcher } as RequestInit) : fetch(input, init);
+};
+
+/**
+ * The fetch used for xet downloads when the caller does not pass one. Only Node >= 26 actually forces
+ * HTTP/1.1; on every other runtime there is nothing to force and this is the global fetch, looked up at
+ * call time so test doubles installed on `globalThis` still apply.
+ */
+const http1Fetch: typeof fetch = hasH2Fetch() ? undiciHttp1Fetch : fetch;
```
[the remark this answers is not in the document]
assafvayner: @coyotte508 I think this is why there was a whole `(p1, p2) => fetch(p1, p2)` before.

Is this a legitimate issue?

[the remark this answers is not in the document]
coyotte508: ah ok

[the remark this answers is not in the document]
coyotte508: can revert

PR #2496, opened 2026-09-15
Add text-ranking inference support

tomaarsen:
Hello!

## Pull Request overview
* Add text-ranking inference support
* ~Fix single quotes in cURL inputs~

## Details
Some time ago, we split `text-ranking` off from `text-classification`, specifically for the many (search) models that do ranking of some sort. There's now ~800 Sentence Transformers text ranking models (https://huggingface.co/models?library=sentence-transformers&other=text-ranking&sort=downloads), many of which with 100k+ monthly downloads, and I think it's a great time to be able to improve API and widget support for them.

Note firstly that `text-ranking` is already supported for some `text-classification` models via the `/pipeline/text-ranking` suffix:
```
curl -s -X POST https://router.huggingface.co/hf-inference/models/BAAI/bge-reranker-v2-m3/pipeline/text-ranking -H "Authorization: Bearer $T" -H 'Content-Type:
  application/json' -d '{"inputs":{"query":"what is panda?","texts":["hi","The giant panda is a bear species endemic to China."]}}'
[{"score":0.83948773,"index":1},{"score":0.00027905524,"index":0}]
```

The idea of this PR is that we can 1) deploy `text-ranking` models more easily and 2) query them with just e.g. https://router.huggingface.co/hf-inference/models/cross-encoder/ms-marco-MiniLM-L6-v2. This is also the base for a follow-up PR that adds Widget support on the Hub.

I've adopted the format from https://github.com/huggingface/huggingface-inference-toolkit/blob/main/src/huggingface_inference_toolkit/sentence_transformers_utils.py#L50:
* Inputs: `query` (str), `texts` (list[str]), and `return_documents` (bool).
* Outputs: list of dicts with "index", "score" and optionally "text".

This means that we stay compatible with current users that rely on the `/pipeline/text-ranking` suffix. Note that this does differ from TEI its `predict` (https://huggingface.github.io/text-embeddings-inference/#/Text%20Embeddings%20Inference/predict), which has always diverged.

~The changes are mostly focused on the text-ranking, except the changes from 95dfc1faca5a10be7f8185c35a122179df5001d1. That second commit fixes json serialization for curl snippets. In short, it didn't play nicely with the newlines (it kept the whole input on one line) and it didn't escape the quote from `Earth's`, so a copy-paste of the default snippet fails. I can also kill this commit if you want to keep it minimal, but I think it's a good inclusion.~
Edit: I decided to remove this as my solution was still a bit faulty, so the pre-existing cURL bug stays existing.

Some screenshots about what it'll look like once the Hub is also looped in:

Schermafbeelding 2026-09-15 124047

<details><summary>Click for a few more screenshots</summary>

Schermafbeelding 2026-09-15 124354

Schermafbeelding 2026-09-15 124404

Schermafbeelding 2026-09-15 130840

</details>

cc @Wauplin @hanouticelina as I'd also like to eventually add this to `huggingface_hub`, but first I want to merge this, then a Hub PR, and only then hf_hub. I'd love your feedback on the format here.
cc @oOraph as we discussed the input/output spec for this

- Tom Aarsen

---

> [!NOTE]
> **Low Risk**
> Additive NLP task wiring with no changes to auth or existing task behavior; risk is mainly new API surface and HF Inference response parsing for the new endpoint.
> 
> **Overview**
> Adds **text-ranking** as a first-class inference task so reranker models can be called with `query` + `texts` and get back ranked `{ index, score }` results (aligned with the existing `/pipeline/text-ranking` shape).
> 
> In **`@huggingface/tasks`**, this introduces JSON schemas, generated TypeScript types, default widget examples, and snippet input helpers for the new task. In **`@huggingface/inference`**, it wires **`textRanking`** through provider helpers (including HF Inference response validation), exports the task, documents **`hf.textRanking`** in the README, and registers snippet generation plus fixture tests for fetch/JS/Python/curl.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2128a618ba9a0831a54cb5cca4469153a1681503. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Wauplin (commented):
PR looks almost good to me as-is 👍 

Just a comment on the format: 

> Inputs: query (str), texts (list[str]), and return_documents (bool).
> Outputs: list of dicts with "index", "score" and optionally "text".
> This means that we stay compatible with current users that rely on the /pipeline/text-ranking suffix. Note that this does differ from TEI its predict (https://huggingface.github.io/text-embeddings-inference/#/Text%20Embeddings%20Inference/predict), which has always diverged.

=> isn't the equivalent route in TEI [`/rerank`](https://huggingface.github.io/text-embeddings-inference/#/Text%20Embeddings%20Inference/rerank) ? From what I can read it diverge with more options + a `return_text` instead of `return_documents`. A solution is that for now we define only of subset of the input/output schema with only

- `query: str` and `texts: str[]` as input
- `index: int` and `scope: float` as output

This way we have a subset of both TEI and sentence transformers and don't loose anything. It's always easier to add more parameters later if a consensus is found rather than making a choice upfront. WDYT of this solution? It would still allow the widget to work + users to run inference via the API.


Happy to get others' opinion as well

tomaarsen:
> => isn't the equivalent route in TEI [`/rerank`](https://huggingface.github.io/text-embeddings-inference/#/Text%20Embeddings%20Inference/rerank) ?

Ah yes, I think you're right. I looked for `predict` as Sentence Transformers' reranker class (`CrossEncoder`) uses `predict` as the main inference interface.

Then I'd be happy to remove `return_documents` to simplify. I can double-check the Widget PR, but it likely already uses the index + the inputs rather than the return scores. If others agree, then I can work on this spec simplification now.

- Tom Aarsen

tomaarsen:
I've implemented the `return_documents` removal to keep this even simpler and more flexible.

- Tom Aarsen

Wauplin (approved):
Looks good!

Let's wait for others' opinion / review before merging.

hanouticelina (approved):
Looks good to me, thank you for adding the snippets!

